PR #8, opened 2019-08-20
Add runner_phpmd

yykamei:


lib/node_harness/runners/phpmd/processor.rb
```diff
@@ -0,0 +1,163 @@
...
+          yield targets, rule, options
+        end
+
+        def rule(config)
+          rules = config[:rule] || config.dig(:options, :rule)
+          if rules
+            rules
+          else
+            # NOTE: It assumes that default config xml is located in $HOME
+            (Pathname(Dir.home) / 'sider_config.xml').realpath.to_s
```
yykamei: Previously, this method supposes that our default config file is placed in the same directory with this module. Now, the module exists in node_harness `lib` directory, so I changed the location. Also, I changed the config file name :wine_glass:

PR #22, opened 2019-08-20
Add runner_rails_best_practices

yykamei:


yykamei:
Hmm... After adding `require "rails_best_practices"`, runner_goodcheck fails.
I'll address it later.

lib/node_harness/runners/rails_best_practices/processor.rb
```diff
@@ -0,0 +1,176 @@
...
+              'rails_best_practices',
+              '--without-color',
+              '--silent',
+              '--output-file=/dev/stdout',
+              '--format=yaml',
+              *options
+            )
+            # NOTE: rails_best_practices returns top-level YAML array with tags.
+            #       We want to just parse a YAML, so we remove YAML tags before passing to `YAML.load`.
+            YAML.load(stdout.gsub('- !ruby/object:RailsBestPractices::Core::Error', '-'), symbolize_names: true).each do |yaml|
```
yykamei: I don't want to load `rails_best_practices` as a gem. So, I added `gsub` to remove YAML tags.

PR #36, opened 2019-08-21
Make CircleCI to push docker images for a Git tag

yykamei:
This is just a test. So, after merging this PR, CircleCI does not push images when a Git tag is pushed.


.circleci/config.yml
```diff
@@ -155,7 +160,7 @@ jobs:
 
 workflows:
   version: 2
-  build:
+  build_and_test:
```
yykamei: I renamed the workflow to correspond to its behavior.

yykamei: This workflow also does "deploy" :thinking: ... I will revert the name.

PR #45, opened 2019-08-23
Enable skipped tests

yykamei:
Also, I renamed the module `Rubocop` with `RuboCop.`

### DIFF
--- a/lib/node_harness/cli.rb
+++ b/lib/node_harness/cli.rb
@@ -80,7 +80,7 @@ module NodeHarness
     def processor_class
       # TODO: Add more runners
       @processor_class ||= {
-        rubocop: NodeHarness::Runners::Rubocop::Processor,
+        rubocop: NodeHarness::Runners::RuboCop::Processor,
         rails_best_practices: NodeHarness::Runners::RailsBestPractices::Processor,
         reek: NodeHarness::Runners::Reek::Processor,
         goodcheck: NodeHarness::Runners::Goodcheck::Processor,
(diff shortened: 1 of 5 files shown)


lib/node_harness/runners/rubocop/processor.rb
```diff
@@ -58,11 +58,11 @@ class Processor < NodeHarness::Processor
           GemInstaller::Spec.new(name: "rubocop-rspec", version: []),
           GemInstaller::Spec.new(name: "rubocop-cask", version: []),
           GemInstaller::Spec.new(name: "rubocop-thread_safety", version: []),
-        ]
+        ].freeze
 
         CONSTRAINTS = {
           "rubocop" => [">= 0.35.0"]
-        }
+        }.freeze
```
ybiquitous: Need the rule? https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MutableConstant

yykamei: The reason I changed the line is that my IDE loudly issues it.
Actually, I'm neutral about it, but we can test the behavior of Sider if we enable the rule.

OK, I'll enable it in this PR. What do you think?

ybiquitous: Agree 👍

lib/node_harness/cli.rb
```diff
@@ -106,8 +106,7 @@ def processor_class
         golint: NodeHarness::Runners::Golint::Processor,
         gometalinter: NodeHarness::Runners::Gometalinter::Processor,
         swiftlint: NodeHarness::Runners::Swiftlint::Processor,
-      }
-      @processor_class.fetch(analyzer.to_sym)
+      }[analyzer.to_sym]
```
yykamei: I updated the line because the presence of `@processor_class` should be checked [here](https://github.com/sider/runners/pull/45/files#diff-272dff460a05d180c35d6eef74d49784R77).

This fix is related to #52.

PR #55, opened 2019-08-26
[EXPERIMENT][DO NOT MERGE] Enable type-checking

ybiquitous:
Fix #46

PR #61, opened 2019-08-30
Add helper method to extract an analyzer version

ybiquitous:
Usage:

```ruby
module NodeHarness
  class Foo::Processor < Processor
    def analyzer_version
      @analyzer_version ||= extract_version! "foo"
    end
  end
end
```

Also, this changes only Flake8 processor using the new methods.
For other processors, I will do it on other PRs.

### DIFF
--- a/lib/node_harness/processor.rb
+++ b/lib/node_harness/processor.rb
@@ -84,11 +84,11 @@ module NodeHarness
       MSG
     end
 
-    def extract_version!(command, version_option = "--version", *extra_options)
+    def extract_version!(command, version_option = "--version", *extra_options, pattern: /v?(\d+\.\d+\.\d+)/)
       command_line = [command, version_option, *extra_options]
       outputs = capture3!(*command_line)
       outputs.each do |output|
-        match = /v?(\d+\.\d+\.\d+)/.match(output)
+        match = pattern.match(output)
         return match[1] if match
       end
       raise "Not found version from '#{command_line.join(' ')}'"
--- a/lib/node_harness/processor.rb
+++ b/lib/node_harness/processor.rb
@@ -74,6 +74,26 @@ module NodeHarness
       analyzer or raise "No analyzer set"
     end
 
+    def analyzer_version
+      raise NotImplementedError, <<~MSG
+        A typical implementation:
+
+        def #{__method__}
+          @#{__method__} ||= extract_version! "some_command"
+        end
+      MSG
+    end
+
+    def extract_version!(command, version_option = "--version", *extra_options)
+      command_line = [command, version_option, *extra_options]
+      outputs = capture3!(*command_line)
+      outputs.each do |output|
+        match = /v?(\d+\.\d+\.\d+)/.match(output)
+        return match[1] if match
+      end
+      raise "Not found version from '#{command_line.join(' ')}'"
+    end
+
     def self.ci_config_section_name
       self.name
     end
--- a/sig/node_harness.rbi
+++ b/sig/node_harness.rbi
@@ -211,6 +211,8 @@ class NodeHarness::Processor
   def with_analyzer: <'x> (Analyzer?) { () -> 'x } -> 'x
   def analyzer: -> Analyzer?
   def analyzer!: -> Analyzer
+  def analyzer_version: -> String
+  def extract_version!: (String, String?) -> String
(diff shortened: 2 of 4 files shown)


lib/node_harness/processor.rb
```diff
@@ -74,6 +74,26 @@ def analyzer!
...
+          @#{__method__} ||= extract_version! "some_command"
+        end
+      MSG
+    end
+
+    def extract_version!(command, version_option = "--version", *extra_options)
+      command_line = [command, version_option, *extra_options]
+      outputs = capture3!(*command_line)
+      outputs.each do |output|
+        match = /v?(\d+\.\d+\.\d+)/.match(output)
```
yykamei: I'm skeptical about every analyzer is based on SemVer. I think Inherited classes should construct the version string on their own, but it's convenient that the method could be used in our all analyzers.

ybiquitous: It seems difficult to me that this helper method supports all analyzers, so I think analyzers without semver should not use this `extract_version!` helper, for example:

```ruby
def analyzer_version
  @analyzer_version ||= begin
    stdout, = capture3! "foo", "-v"
    stdout.match(/(\d\-\d)/)[1] # e.g. 1-234, 9-123
  end
end
```

yykamei: OK :+1: 

The code itself is good, but I have a suggestion. How about renaming the method to `extract_semantic_version!`?

ybiquitous: I got a good idea! How about changing the pattern as follow?

```ruby
processor.extract_version!("foo", "bar", "-v", pattern: /ver (\d+.\d+)\b/)
# input: "ver 1.24"
# output: "1.24"
```

ybiquitous: Wokrs fine. 😄 

```diff
diff --git a/lib/node_harness/processor.rb b/lib/node_harness/processor.rb
index 26fb3e6..ba20ee8 100644
--- a/lib/node_harness/processor.rb
+++ b/lib/node_harness/processor.rb
@@ -84,11 +84,11 @@ module NodeHarness
       MSG
     end
 
-    def extract_version!(command, version_option = "--version", *extra_options)
+    def extract_version!(command, version_option = "--version", *extra_options, pattern: /v?(\d+\.\d+\.\d+)/)
       command_line = [command, version_option, *extra_options]
       outputs = capture3!(*command_line)
       outputs.each do |output|
-        match = /v?(\d+\.\d+\.\d+)/.match(output)
+        match = pattern.match(output)
         return match[1] if match
       end
       raise "Not found version from '#{command_line.join(' ')}'"
diff --git a/test/processor_test.rb b/test/processor_test.rb
index c460f73..543e04b 100644
--- a/test/processor_test.rb
+++ b/test/processor_test.rb
@@ -389,6 +389,10 @@ class ProcessorTest < Minitest::Test
       mock(processor).capture3!("foo", "bar", "version") { ["", "7.8.20 version\n"] }
       assert_equal "7.8.20", processor.extract_version!("foo", "bar", "version")
 
+      # change pattern
+      mock(processor).capture3!("foo", "bar", "-v") { ["ver 1.2", ""] }
+      assert_equal "1.2", processor.extract_version!("foo", "bar", "-v", pattern: /ver (\d+.\d+)\b/)
+
       # not found
       mock(processor).capture3!("foo", "-v") { ["2.1", ""] }
       error = assert_raises { processor.extract_version!("foo", "-v") }
```

yykamei: :+1:

sig/node_harness.rbi
```diff
@@ -211,6 +211,8 @@ class NodeHarness::Processor
   def with_analyzer: <'x> (Analyzer?) { () -> 'x } -> 'x
   def analyzer: -> Analyzer?
   def analyzer!: -> Analyzer
+  def analyzer_version: -> String
+  def extract_version!: (String, String?) -> String
```
yykamei: Will you fix it in #55?

ybiquitous: #55 is too big, so I will close it. 😓 

I think we need type-checking **gradually**...

ybiquitous:
@yykamei Can I merge this PR? OK?

ybiquitous:
Thanks!

Issue #59, opened 2019-08-30
Rename NodeHarness namespace to something else?


ybiquitous:
Because we may be confusing... 🤔 

https://github.com/sider/runners/blob/770c6e06010fc8a19f218178252b8ccf6923d34d/lib/node_harness.rb#L86

yykamei:
I agree, but I have no ideas :sweat_smile:

ybiquitous:
`Sider::Runners`?

ybiquitous:
Or just `Runners`?

yykamei:
`Harness` is good to me, but the current implementation already uses `Harness` as a class.

`Runners::Runners::RuboCop` or `Sider::Runners::Runners::RuboCop` is weird...

ybiquitous:
Sure, let's keep the issue as is until we will get some good ideas...

ybiquitous:
One idea: `Runners::Processor::RuboCop`, `Runners::Processor::Eslint`, ...

yykamei:
`Runners::Processor::RuboCop` seems good :+1: I didn't get the idea.

PR #79, opened 2019-09-12
[Refactor] Move processors directory

ybiquitous:
E.g. `NodeHarness::Runners::Checkstyle::Processor` -> `NodeHarness::Processor::Checkstyle`

See also #59


ybiquitous:
@yykamei Could you please review this lightly if you have time?

yykamei:
I'm curious why you updated runners without nesting.

I know the following causes a TypeError:

```ruby
class A
end

module A
  class R1
  end
end
```

From my understanding, `Processor` in this project is declared as a class, so all nested classes are declared with this syntax. Right?

```ruby
class A
end

class A::R1
end
```

ybiquitous:
Yes, I understand this is just a stylistic issue because the all test passed.
Also, I don't like personally deeply nesting, but one level nesting seems to be reasonable like this:

```ruby
module NodeHarness
  class Processor::Cppcheck < Processor
  end
end
```

What do you think about this?

yykamei:
Yeah, I agree. I just wanted to know. It's reasonable.

ybiquitous:
Thanks. I will update them. 💪

ybiquitous:
I will update all files via this script:

```ruby
file = ARGV[0] or abort "Require file!"

lines = File.readlines(file, chomp: true).map do |line|
  "  #{line}".gsub("NodeHarness::", "") unless line.empty?
end
lines.prepend "module NodeHarness"
lines.append "end"

File.write file, (lines.join("\n") + "\n")
```

PR #84, opened 2019-09-12
Improve error message when `root_dir` is not found

ybiquitous:


lib/node_harness/processor.rb
```diff
@@ -122,22 +124,31 @@ def ci_section(default = {})
...
+    def assert_root_dir_exist!
+      key = "root_dir"
+      root_dir = ci_section[key]
+      return unless root_dir
+
+      (working_dir / root_dir).tap do |dir|
+        unless dir.directory?
+          message = "`#{root_dir}` directory is not found! Please check `#{key}` in your `#{ci_config_path_name}`."
+          trace_writer.error message
+          raise RootDirNotFound, message
```
ybiquitous: @yykamei Instead of `raise`, should we use `NodeHarness::Results::Failure`?

yykamei: `NodeHarness::Results::Failure` is appropriate because runners cannot continue processing if they can't change to `root_dir` and this failure is caused by users' configuration.

ybiquitous: Thanks. Then, should we check `root_dir` in `NodeHarness::Harness#run` instead of `Processor`?

https://github.com/sider/runners/blob/5bb8fff3c047376f5f2da6070b204f57185325ef/lib/node_harness/harness.rb#L37

yykamei: I agree.

By the way, `#push_root_dir` is a little difficult for me because it just pushes a directory string into its array stack instead of exactly changing directory.

ybiquitous: Me too. But I don't know a good solution... 🤷‍♂

PR #92, opened 2019-09-17
Fix stylelint smoke tests

ybiquitous:
Lock version to prevent unexpected results when stylelint will ship new versions.


ybiquitous:
Fixed via #93

PR #93, opened 2019-09-17
Fix `stylelint-config-recommended` version

ybiquitous:
`STYLELINT_CONFIG_STANDARD_VERSION` is not defined, so the latest version (`3.0.0`) of `stylelint-config-recommended` is unexpectedly installed instead of `stylelint-config-recommended@2.2.0`.


ybiquitous:
Close #92

PR #96, opened 2019-09-17
Rename namespace: `NodeHarness` -> `Runners`

### LINKED ISSUE #59: Rename NodeHarness namespace to something else?

ybiquitous:
Fix #59


ybiquitous:
Thanks for the review!

PR #99, opened 2019-09-17
Install default npm tools via `package.json`

ybiquitous:
Overview
--------

This installs the default versions of npm tools via `package.json` instead of `npm install --global`.
Each tool is installed into `${RUNNER_USER_HOME}/<tool_name>/node_modules`.

Also, this `${RUNNER_USER_HOME}/<tool_name>/node_modules` is set into the `NODE_PATH` environment variable.
(see <https://nodejs.org/api/cli.html#cli_node_path_path>)
By doing this, we can find packages installed into the `node_modules/` directory.

Why?
----

The installation via `package.json` is safer than directly writing the CLI command `npm install --global` in `Dockerfile`.
The former never raise an issue like #93 and makes the tool's installation easier for us.


images/stylelint/sider_recommended_config.yaml
```diff
@@ -1 +1 @@
-extends: "/usr/local/lib/node_modules/stylelint-config-recommended"
+extends: "stylelint-config-recommended"
```
ybiquitous: `NODE_PATH` help this simplicity.

images/tslint/Dockerfile
```diff
@@ -17,15 +17,20 @@ ENV BUNDLE_GEMFILE ${RUNNERS_DIR}/Gemfile
...
+ENV PATH ${RUNNER_USER_HOME}/tslint/node_modules/.bin:${PATH}
+# See https://nodejs.org/api/cli.html#cli_node_path_path
+ENV NODE_PATH ${RUNNER_USER_HOME}/tslint/node_modules:${NODE_PATH}
 
-USER root
-RUN npm install -g tslint@${LINT_TOOL_VERSION} typescript@${TYPESCRIPT_VERSION}
-USER $RUNNER_USER
 
 # By default, TSLint looks for a configuration file named tslint.json in the directory of the file being linted and, if not found, searches ancestor directories.
-# `default_tslint_${LINT_TOOL_VERSION}.json` is generated by `tslint --init`
```
ybiquitous: Outdated comment.

Rakefile
```diff
@@ -39,6 +41,12 @@ namespace :dockerfile do
         #{template.result.chomp}
       EOD
       File.write(path / 'Dockerfile', result)
+    ensure
+      if backup_analyzer
+        ENV['ANALYZER'] = backup_analyzer
+      else
+        ENV.delete 'ANALYZER'
+      end
```
ybiquitous: I feel a bit dirty. Please tell me if there is a better way!

ybiquitous:
Thanks!

PR #110, opened 2019-09-18
Install default gems via `Gemfile`

ybiquitous:
This uses `bundle install --system` via `Gemfile` instead of `gem install`.
By this, Dependabot auto pull request will help us.

Related to #99


yykamei:
By the way, each processor of Ruby analyzer is defining `Constraints` like this:

https://github.com/sider/runners/blob/42d055236db30ce557b4ec4a3cf43a8684014ed4/lib/runners/processor/reek.rb#L11-L13

So, we have to update it when the dependencies are bumped. Right?

ybiquitous:
Yes, I think so. If we would forget updating `Constraints`, I expect that smoke tests would fail.

PR #111, opened 2019-09-18
Use Pipenv to address the dependency of Flake8

yykamei:
Like #99, I want to install Flake8 via Pipenv.

### DIFF
--- a/images/Dockerfile.python.erb
+++ b/images/Dockerfile.python.erb
@@ -5,7 +5,12 @@ RUN cd ${RUNNER_USER_HOME} && \
     mkdir <%= analyzer %> && \
     mv <%= analyzer %>_Pipfile <%= analyzer %>/Pipfile && \
     cd <%= analyzer %> && \
-    pipenv install --python ${PYTHON2_VERSION} --system --skip-lock && \
-    pipenv install --python ${PYTHON3_VERSION} --system --skip-lock && \
-    pyenv rehash && \
-    pyenv whence <%= analyzer %>
+    for version in "${PYTHON2_VERSION}" "${PYTHON3_VERSION}"; do \
+      pyenv global ${version} && \
+      pipenv install --system --skip-lock && \
+      pyenv rehash && \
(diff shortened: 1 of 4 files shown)


yykamei:
We should install Pipenv in devon_rex. I'm addressing it with sider/devon_rex#48.

images/flake8/Pipfile
```diff
@@ -0,0 +1,7 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[packages]
+flake8 = "==3.7.7"
```
ybiquitous: [question] Is `3.7.4` installed when seeing a CircleCI build? ⬇️ 

https://circleci.com/gh/sider/runners/7961

image

yykamei: This is the Python runtime version generated by `pyenv whence`. However, the expected output is `2.7.16\n3.7.4\n` :thinking: 
I'll fix Dockerfile.python.erb

ybiquitous:
Dependabot supports now `Pipenv`!

https://dependabot.com/python/

images/Dockerfile.python.erb
```diff
@@ -0,0 +1,16 @@
...
+COPY images/<%= analyzer %>/Pipfile ${RUNNER_USER_HOME}/<%= analyzer %>_Pipfile
+RUN cd ${RUNNER_USER_HOME} && \
+    mkdir <%= analyzer %> && \
+    mv <%= analyzer %>_Pipfile <%= analyzer %>/Pipfile && \
+    cd <%= analyzer %> && \
+    for version in "${PYTHON2_VERSION}" "${PYTHON3_VERSION}"; do \
+      pyenv global ${version} && \
+      pipenv install --system --skip-lock && \
+      pyenv rehash && \
+      echo "In '$(python --version 2>&1)', the version of the analyzer '<%= analyzer %>' is ..." && \
```
yykamei: Python 2 writes its version to STDERR, so I used redirection here.

images/flake8/Pipfile
ybiquitous: Looks good! 👍 

image

PR #122, opened 2019-09-19
Remove useless `BUNDLE_GEMFILE` from `Dockerfile`

ybiquitous:
The `BUNDLE_GEMFILE` environment variable in `Dockerfile` has been used to prevent `Bundler::GemFileNotFound` raised by `Bundler::LockfileParser`.

However, this change adds a workaround to prevent the error, so the environment variable is unnecessary.
I think we should avoid too much usage of environment variables because they could produce unexpected behaviors.


lib/runners/ruby/gem_installer.rb
```diff
@@ -48,11 +48,11 @@ def install!
...
 
-          shell.push_env_hash env_hash do
-            versions = Bundler::LockfileParser.new(lockfile_path.read).specs.each.with_object({}) do |spec, versions|
-              versions[spec.name] = spec.version.version
-            end
+          versions = LockfileParser.parse(lockfile_path.read).specs.each_with_object({}) do |spec, versions|
+            versions[spec.name] = spec.version.version
+          end
 
+          shell.push_env_hash env_hash do
```
ybiquitous: `push_env_hash` affects ONLY on the usage of `shell.capture3`. This behavior seems confusing to me, so I put `LockfileParser` out of the block.

PR #123, opened 2019-09-19
Use `Processor#analyzer_version` in its sub-classes and normalize version strings of Go tools

ybiquitous:


ybiquitous:
This PR will change Go tools' versions. Is this breaking? What do you think? @yykamei 

image

https://circleci.com/gh/sider/runners/8117

ybiquitous:
Thanks for your review. These version string is internal data, so not breakign.

ybiquitous:
This PR is not refactoring when fixing Go tools' versions, so I modified the PR title.

PR #162, opened 2019-09-20
Add IO::AwsS3 to upload traces to AWS S3

yykamei:
We have a plan to save runners' traces to AWS S3. In order to do that, I added `IO::AwsS3` class.
It behaves like `IO` objects, but it has `#finalize!` method, which uploads all traces to S3.
Before `#finalize!` is called, an `IO::AwsS3` instance writes traces to a temporary file as well as the passed STDOUT. I chose a temporary file as the temporary store instead of a memory store like `StringIO` because I can't estimate the usage of memory.

The feature of uploading to S3 is currently optional. If you want to use this feature, `--output=S3_URI` option is required. Of course, you have to prepare an AWS S3 bucket and allow `Runners` to upload files.


test/cli_test.rb
```diff
@@ -14,7 +14,7 @@ def stderr
   end
 
   def test_parsing_options
-    cli = CLI.new(argv: %w(--analyzer=rubocop --base=base.tgz --head=https://example.com/head --base-key=basekey --head-key=headkey --ssh-key=id_rsa --working=/working/dir test-guid), stdout: stdout, stderr: stderr)
+    cli = CLI.new(argv: %w(--analyzer=rubocop --base=base.tgz --head=https://example.com/head --base-key=basekey --head-key=headkey --ssh-key=id_rsa --working=/working/dir --s3-uri=s3://bucket/a/b/c test-guid), stdout: stdout, stderr: stderr)
```
yykamei: > __[RuboCop]__ Metrics/LineLength: Line is too long. [242/200] ([link](https://sider.review/gh/repos/202486306/pulls/162#issue-4592632))

Close!

Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'bundler', '>= 1.12', '< 3.0'
 gem 'nokogiri', '>= 1.10.4'
```
ybiquitous: I removed Nokogiri via #161. Please rebase 🙏

test/aws_s3_io_test.rb
```diff
@@ -0,0 +1,36 @@
...
+    assert_equal 'abc!!!', io.stdout.tap(&:rewind).read
+  end
+
+  def test_finalize
+    mock_object = Object.new
+    mock(Aws::S3::Client).new(retry_limit: is_a(Numeric), retry_base_delay: is_a(Numeric)) { mock_object }
+    mock(mock_object).put_object(bucket: 'bucket_name', key: 'object_name', body: instance_of(Tempfile))
+    io = Runners::AwsS3IO.new(stdout, 's3://bucket_name/object_name')
+    io.finalize!
+  end
```
yykamei: Perhaps, we should test with an actual AWS S3 bucket. The current test mocks an AWS API call.

lib/runners/aws_s3_io.rb
```diff
@@ -0,0 +1,44 @@
+module Runners
+  # AwsS3IO expects AWS credentials set with environment variables or an instance profile.
+  # So, before using this, make sure AWS credentials available.
+  # Also, prepare the S3 bucket and allow this instance to upload an S3 object.
+  class AwsS3IO
```
ybiquitous: [nitpick] How about naming `Runners::IO::AwsS3`? Assuming `Runners::IO::GoogleCloud`, `Runners::IO::Azure`, and so on...

yykamei: I see. It makes sense :+1:

test/aws_s3_io_test.rb
ybiquitous: Will this test be slow if using actual AWS?

yykamei: Yes, I think so.

lib/runners/cli.rb
```diff
@@ -69,6 +74,7 @@ def validate_options!
       raise "base_key is given but base is missing" if !base && base_key
       raise "head_key is given but head is missing" if !head && head_key
       validate_analyzer!
+      AwsS3.parse_s3_uri!(s3_uri) if s3_uri
```
ybiquitous: I think this validation is unnecessary because `AwsS3#initialize` validates it later. 🤔

lib/runners/cli.rb
```diff
@@ -104,6 +111,7 @@ def run
             writer << Schema::Result.envelope.coerce(json)
           end
         end
+        io.finalize! if io.instance_of?(Runners::IO::AwsS3)
```
ybiquitous: How about this? If `Runners::IO::*` classes will be added, we should fix this code, but it's easy to forget. 😅 

```suggestion
        io.finalize! if io.respond_to?(:finalize!)
```

lib/runners/io/aws_s3.rb
```diff
@@ -0,0 +1,46 @@
...
+    # Also, prepare the S3 bucket and allow this instance to upload an S3 object.
+    class AwsS3
+      def self.parse_s3_uri!(s3_uri)
+        prefix = 's3://'
+        raise "The URI must starts with 's3://'. The actual value is '#{s3_uri}'" unless s3_uri.start_with?(prefix)
+        s3_uri.delete_prefix(prefix).split('/', 2).tap do |result|
+          bucket, object = result
+          raise "The specified S3 URI is not valid. The actual value is '#{s3_uri}'" unless bucket && object
+        end
+      end
```
ybiquitous: I think better to use `URI` class for readability.

For example:

```ruby
uri = URL(s3_uri)
bucket = uri.host
object = uri.path
if uri.scheme == "s3" && bucket && object
  [bucket, object]
else
  raise "..."
end
```

lib/runners/cli.rb
yykamei: Looks good. Thank you!

test/aws_s3_io_test.rb
ybiquitous: I'm worried if we should add actual tests because:
- Tests and CI will be slower.
- Setup and onboarding to this project will be harder. For example, AWS credentials will be needed.

Are the disadvantages bigger than the advantages?

lib/runners/io/aws_s3.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def finalize!
+        tempfile.rewind
+        client.put_object(
+          bucket: s3_bucket_name,
+          key: s3_object_name,
+          body: tempfile,
+        )
+      end
```
ybiquitous: [question] Will not `tempfile` be removed after `finalize!`

lib/runners/cli.rb
yykamei: I thought so too, but it might be better for us to know the error faster than the evaluation of `AwsS3#initialize.`
However, I agree with you. I'll fix it.

lib/runners/io/aws_s3.rb
yykamei: Yes, I read [this](https://ruby-doc.org/stdlib-2.6.4/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Explicit+close) reference.

test/aws_s3_io_test.rb
```diff
@@ -0,0 +1,36 @@
+require_relative "test_helper"
```
ybiquitous: How about renaming `test/aws_s3_io_test.rb` to `test/io/aws_s3_test.rb`?

lib/runners/io/aws_s3.rb
ybiquitous: I got it. 👍

test/aws_s3_io_test.rb
yykamei: I thought such an integration test should be run in only CI. But, to do this, we have to write extra code.

Like this implementation, I won't do it soon.

lib/runners/cli.rb
```diff
@@ -39,6 +40,10 @@ def initialize(argv:, stdout:, stderr:)
         opts.on("--working=WORKING") do |working|
           @working_dir = working
         end
+        opts.on("--s3-uri=S3_URI",
+                "AWS S3 URI (e.g. 's3://bucket/some/path'). Runners writes traces to the specified S3 object") do |s3_uri|
+          @s3_uri = s3_uri
+        end
```
ybiquitous: [IMO] I got just an idea, perhaps `--output` option may be more flexible, for example:

By default (when omitting), `stdout`:

```
$ runners [--output=stdout] ...
```

Explicit:

```
$ runners --output=stdout ...
$ runners --output=stderr ...
```

Multiple:

```
$ runners --output=s3://some_bucket/some/object --output=stdout
```

What do you think? Too complex?

test/aws_s3_io_test.rb
ybiquitous: Yes, let's think later.

ybiquitous:
The class names has been renamed, so how about changing the PR title like this?

```diff
-Add AwsS3IO to upload traces to AWS S3
+Upload traces to AWS S3
```

test/aws_s3_io_test.rb
yykamei: I'm doing it.

lib/runners/cli.rb
yykamei: It looks, but it's complex to me.

yykamei: I need more time to do it.

ybiquitous: OK, keep as-is. 👍

yykamei: I understand the behavior of `OptionParser`.

Given this file (`test.rb`):
```ruby
require 'optparse'

outputs = []
OptionParser.new do |opts|
  opts.on("--output=OUTPUT") do |output|
    outputs << output
  end
end.parse!(ARGV)

p outputs
```

I got such result:

```shell-session
$ ruby test.rb --output a --output b --output c
["a", "b", "c"]
```

lib/runners/io/aws_s3.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      # @param stdout [IO, StringIO]
+      # @param s3_uri [String]
+      def initialize(stdout, s3_uri)
+        @stdout = stdout
+        @s3_bucket_name, @s3_object_name = self.class.parse_s3_uri!(s3_uri)
+        @tempfile = Tempfile.new
+        @ios = [stdout, tempfile]
+        @client = Aws::S3::Client.new(retry_limit: 5, retry_base_delay: 1.2)
+      end
```
ybiquitous: [suggestion] About naming, I have some suggestions (but very personal opinion):

- Rename `stdout` -> `io` (using abstract `IO` interface)
- Remove `s3_` prefixes from methods because the class name `AwsS3` already works as some prefix.

lib/runners/cli.rb
ybiquitous: Nice! 👍 I think good to do this in another PR, but if you do this in this PR, I don't stop it. 😄

yykamei:
I will fix this PR with the `--output` option. It would be difficult to update after merging it.

yykamei:
@ybiquitous Before rebasing, I want you to review this PR which introduces the `--output` option :pray:

yykamei:
Thank you so much!

PR #192, opened 2019-09-24
Support ktlint

ybiquitous:
*ktlint* is a linter for Kotlin.
See <https://github.com/pinterest/ktlint>.

See also https://github.com/sider/sider-docs/pull/198


ybiquitous:
Thank you for your review!

> Will users set `gradle` and `maven` configs?

Honestly, I'm not sure how many users will set them. But I assume a build tool like Maven or Gradle is used in a large Java project.

yykamei:
I see. For such big projects, `gradle` and` maven` options seem useful and they can easily introduce Sider to their projects.

Issue #129, opened 2019-09-19
Manage Java dependencies via Maven


ybiquitous:
Like #99. This will make Dependabot auto-update available on tools' update!

See https://dependabot.com/java/

ybiquitous:
`mvn install` command is too slow. It does many tasks such as downloading all dependencies, compiling, and so on.

ybiquitous:
Close via #216, #222, #224

PR #216, opened 2019-09-27
Manage Checkstyle dependency via Maven

ybiquitous:
This is a challenge for #129.

### DIFF
--- a/lib/runners/processor/checkstyle.rb
+++ b/lib/runners/processor/checkstyle.rb
@@ -41,10 +41,6 @@ module Runners
       args
     end
 
-    def jar_path
-      ENV["CHECKSTYLE_JAR_PATH"] || "/usr/local/share/checkstyle-all.jar"
-    end
-
     def checkstyle(*args, config: nil, format: nil, excludes: nil, properties: nil)
       capture3("java", *checkstyle_common_options, *checkstyle_args(*args, config: config, format: format, excludes: excludes, properties: properties))
     end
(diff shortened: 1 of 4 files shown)


lib/runners/processor/checkstyle.rb
```diff
@@ -41,10 +41,6 @@ def checkstyle_args(*args, config:, format:, excludes:, properties:)
       args
     end
 
-    def jar_path
-      ENV["CHECKSTYLE_JAR_PATH"] || "/usr/local/share/checkstyle-all.jar"
```
ybiquitous: `CHECKSTYLE_JAR_PATH` is unused.

lib/runners/processor/checkstyle.rb
```diff
@@ -54,7 +50,11 @@ def checkstyle!(*args, config: nil, format: nil, excludes: nil, properties: nil)
     end
 
     def checkstyle_common_options
-      ["-Duser.country=#{locale.country}", "-Duser.language=#{locale.language}", "-jar", jar_path]
+      %W[
+        -Duser.country=#{locale.country}
+        -Duser.language=#{locale.language}
+        com.puppycrawl.tools.checkstyle.Main
```
ybiquitous: [note]
- Before: Using an executable JAR does not need the main class name.
- After: Using `CLASSPATH` needs the main class name (with multiple JAR files).

ybiquitous:
It works expectedly! 🎉

Issue #131, opened 2019-09-19
Drop RuboCop "<= 0.60.0"


ybiquitous:
We'd like to upgrade Ruby to version 2.6. Now, only RuboCop uses Ruby 2.5.

https://github.com/sider/runners/blob/15b8cc14510844c67fafe84acfe8cb60418412b3/images/rubocop/Dockerfile#L7-L16

---

> RuboCop versions less than v0.61.0 cannot use Psych.safe_load in Ruby 2.6.

--- https://github.com/sider/runners/issues/57#issuecomment-530858769

ybiquitous:
See #599

PR #228, opened 2019-09-30
Add deprecation warning for too old RuboCop versions

ybiquitous:
See #131


ybiquitous:
@yykamei Do you have any opinions about this PR?

PR #239, opened 2019-10-02
Use `master` tag for devon_rex images

ybiquitous:
These tags should be updated before release.


ybiquitous:
This PR will increase one task on release. What do you think? @yykamei

ybiquitous:
Ooops, I'll fix gometalinter's smoke tests. 😖

yykamei:
> This PR will increase one task on release.

What do you mean?

ybiquitous:
For example, following changes will be needed before release.

```diff
-FROM sider/devon_rex_ruby:master
+FROM sider/devon_rex_ruby:2.3.0
```

1. Change `FROM` of all `Dockerfile`s.
2. Open PR.
3. Release

yykamei:
After releasing, should we get the tag back to `master`? I'm not confident in doing the flow of operations without forgetting.

The purpose of this PR is to make the CI always test the devon_rex with `master`, right?

ybiquitous:
Hm, I think that using `master` instead of a tag` on production will make debugging hard.

This PR purpose is to run smoke tests of Runners using unreleased devon_rex images.

But, when we will release the new version of Runners, we should change tags of devon_rex images from `master` to the devon_rex images' latest tag (e.g. `2.3.0`).

The release workflow will be like this:

1. Change devon_rex images' tags from `master` to the latest tag (e.g. `2.3.0`).
2. Open a PR including the tag changes above.
3. CI pass.
4. Merge the PR.
5. Release a new version of Runners. (tagging and publishing)
6. Change back devon_rex images' tags from `2.3.0` to `master`. (back to developing)

We will want a new Rake task that automates the workflow.

yykamei:
Thanks, I understood.

I'm not sure what to do or how to do, but `devon_rex` and `runners` should be integrated into one repository if such needs exist. Of course, creating another Rake task would help us, but this still creates extra commits and takes a little long time to release.

ybiquitous:
Currently, making one repository will reduce complexity but CI build time will increase.
Perhaps, we need to consider more what pros and cons it will produce.

PR #249, opened 2019-10-03
Retrieve --head, --head-key, --base, and --base-key from env

### LINKED ISSUE #246: Receive sensitive data from environment variables
Like #205, runners should receive sensitive data from environment variables.
Specifically, `--head-key` and `--base-key` must not be exposed.

yykamei:
Resolve #246

`--head` and `--base` are also secure information in Sider.


yykamei:
@ybiquitous What do you think about this PR?

Issue #238, opened 2019-10-02
Lock MeowCop version for RuboCop runner


ybiquitous:
Currently, the latest version of MeowCop is installed and used at runtime.
This behavior can make tests unstable.

https://github.com/sider/runners/blob/b452d0eaa2e4d844e2ac9fe08be50786fb8daf45/lib/runners/processor/rubocop.rb#L80

ybiquitous:
Related to #237

ybiquitous:
The latest MeowCop requires usually the latest RuboCop. When a user's Rubocop version is not the latest, the installation of MeowCop 2.4.0 (requires RuboCop 0.75.0+) will fail.

https://github.com/sider/meowcop/blob/5522707261ce651ae95d6aa1d71642f648a4a805/meowcop.gemspec#L30

ybiquitous:
We cannot lock MeowCop versions now.

PR #298, opened 2019-10-09
Fix failed smoke test for RuboCop runner

ybiquitous:
The cause is that new MeowCop version 2.4.0 has been released.
See https://github.com/sider/meowcop/blob/master/CHANGELOG.md#240-2019-10-08

Related to #238


ybiquitous:
CI passed! 🎉

PR #306, opened 2019-10-10
Use timeout(1) to forcefully terminate too late Runners

yykamei:
Relevant PR is #305

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -153,6 +153,23 @@ namespace :docker do
     sh "bin/runners_smoke #{image_name} test/smokes/#{analyzer}/expectations.rb"
   end
 
+  desc 'Run timeout test'
+  task :timeout_test do
+    File.write('bin/runners', <<~EOF)
+      #!/usr/bin/env ruby
+      require "open3"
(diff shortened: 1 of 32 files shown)


Rakefile
```diff
@@ -153,6 +153,23 @@ namespace :docker do
...
+  task :timeout_test do
+    File.write('bin/runners', <<~EOF)
+      #!/usr/bin/env ruby
+      require "open3"
+      Open3.capture3('sleep 10')
+    EOF
+    ENV['RUNNERS_TIMEOUT'] = '5s'
+    Rake::Task['docker:build'].invoke
+    sh "docker run --rm #{image_name}" do |_ok, res|
+      res.exitstatus == 124 or abort "Expected exit status is 124, but the actual value is #{res.exitstatus}"
```
yykamei: > __[Querly]__ Process#exitstatus may be nil ([link](https://sider.review/gh/repos/202486306/pulls/306#issue-4683213))
>
> <details>
>
>
>
>
> </details>

When the value is `nil`, the CI fails expectedly. There's no problem.

PR #312, opened 2019-10-11
Notify exception if the result is "error"

yykamei:
The instance of `Runners::Results::Error` contains unhandled exceptions. I want Runners to notify them. So, I changed `Runners::CLI#run` to return a result.


yykamei:
Thank you!

PR #313, opened 2019-10-11
Reconsider issue data structure

ybiquitous:
Summary
-------

- Unify `Runners::Issues::Structured` and `Runners::Issues::Text` classes to `Runners:Issue` class.
- Validate issues via StrongJSON.
- Avoid message formatting for some runners. This should be done on the view layer.

Why?
----

I think that separating issues to "structured" and "text-based" is not very useful and maintainable.
Unifying classes to handle an issue will make it easier to handle issues on the Sider front-end layer.

To-do
-----

- [x] We should apply this PR's new data structure to Sider's front-end before this version will be released.


lib/runners/issue.rb
```diff
@@ -0,0 +1,76 @@
...
+      path.instance_of?(Pathname) or
+        raise ArgumentError, "#{path.inspect} must be a #{Pathname}"
+
+      (id && !id.empty?) or
+        raise ArgumentError, "#{id.inspect} must be required"
+
+      (message && !message.empty?) or
+        raise ArgumentError, "#{message.inspect} must be required"
+
+      verify_object!(object, schema)
```
ybiquitous: [note] This raises an error on `#initialize` method instead of `#valid?` method. Because these errors are not for users and are due to Runner's bugs.

ybiquitous:
@yykamei Could you please tell me your idea about this PR?

lib/runners/issue.rb
```diff
@@ -0,0 +1,76 @@
...
+        message: message,
+        links: links,
+        object: object,
+      )
+    end
+
+    private
+
+    def verify_object!(object, schema)
+      schemas = Array(schema)
```
yykamei: Why do you wrap the `schema` with `Array`? I think checking whether the `schema` is `nil` or not is adequate.

ybiquitous: This wrapping is to convert a string or a string array to an array (e.g. `verify_object!({}, [Schema.foo, Schema.bar])`).

https://github.com/sider/runners/blob/e3c742b4b2789309a4169450bcd86965d6f1793b/test/issue_test.rb#L77

But the case that `schema` is an array does not exist actually on production (only test).
So, I'll restrict `schema` is only a string or `nil`.

ybiquitous: Fixed via 45360a58cb30e465f449da07d320c5f65c8fee9b

PR #319, opened 2019-10-16
Fix Checkstyle output XML parse error

ybiquitous:
Checkstyle can output non-XML when some error occurs.
In the case, this change aims to handle the error correctly.

NOTE: `REXML::Document#root` returns `nil` instead of raising an exception as follows:

```ruby
require "rexml/document"
REXML::Document.new("invalid XML").root #=> nil
```


ybiquitous:
Smoke tests passed! 🎉

PR #320, opened 2019-10-16
Make timeout(1) call SIGUSR2 and trap it in bin/runners

yykamei:
Currently, Runners process will stop without notifying to Bugsnag when timeout(1) calls a signal.
This PR makes timeout(1) to call SIGUSR2 instead of SIGTERM (the default signal of timeout(1)), and trap SIGUSR2 in `bin/runners` and notify `RunnersTimeoutError`.

I don't want Runners to trap SIGTERM because Runners cannot tell the difference whether the signal is called by timeout(1) or not. So, I chose SIGUSR2 as a signal dispatched by timeout(1).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 require "rake/testtask"
 require 'erb'
 require "aufgaben/release"
+require_relative "lib/tasks/docker/timeout_test"
 
 Aufgaben::Release.new
 
@@ -20,7 +21,7 @@ task :default => [:test, :typecheck]
 
 task :typecheck do
   # FIXME: Must check all files.
-  files = Dir.glob("lib/**/*.rb").reject { |f| f.include? "processor/" }
+  files = Dir.glob("lib/runners/**/*.rb").reject { |f| f.include? "processor/" }
   files += Dir.glob("lib/**/cppcheck/**/*.rb")
   sh "steep", "check", *files
 end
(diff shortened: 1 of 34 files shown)


yykamei:
@ybiquitous What do you think about this PR? If you have any concerns, let me know them :smiley:

bin/runners
```diff
@@ -8,17 +8,27 @@ require "bugsnag"
...
 Bugsnag.configure do |config|
   config.app_version = ENV["RUNNERS_VERSION"]
 end
 
-at_exit do
-  if $!
-    Bugsnag.notify($!)
+def notify_with_bugsnag(exception)
+  Bugsnag.notify(exception) do |report|
+    report.add_tab(:guid, ENV["TASK_GUID"]) if ENV["TASK_GUID"]
```
yykamei: `:task_guid` is better. I'll fix it.

ybiquitous:
Sorry, we have already `test-timeout` on CI (but this is limited...)
Have you tested this on your local machine?

yykamei:
> Sorry, we have already `test-timeout` on CI (but this is limited...)
> Have you tested this on your local machine?

Yes, I have. :thinking: OK, I'll consider testing on CI.

Rakefile
```diff
@@ -155,19 +156,51 @@ namespace :docker do
...
+        ENV['RUNNERS_TIMEOUT'] = '5s'
+        Rake::Task['docker:build'].invoke
+      end
+      stdout, _, _ = Open3.capture3 "docker inspect -f '{{ range $item := .Config.Entrypoint }}{{ $item }} {{ end }}' #{image_name}"
+      entrypoint = stdout.split.inspect
+      sh "docker run --name #{container_name} --entrypoint bash #{image_name} gem install rr"
+      sh "docker commit --change='ENTRYPOINT #{entrypoint}' --change='CMD []' #{container_name} #{image_name}"
+      _, stderr, status = Open3.capture3 "docker run --rm #{image_name}"
+      status.exitstatus == 124 or abort "Expected exit status is 124, but the actual value is #{res.exitstatus}"
+      !stderr.include?('unexpected method invocation') or abort "Bugsnag is expected to be called"
```
yykamei: `timeout_test` became very ugly :cry: , but we can check `Bugsnag` is called using `rr` when timeout(1) calls a signal.

ybiquitous: How about extracting the test code to `lib/tasks/timeout_test.rb`?

yykamei: Make sense. Thank you. I try it.

yykamei: I fixed like this. 94ac89cc9cd651c6d952e5a5ed5289cf606dabfa

Rakefile
```diff
@@ -1,6 +1,8 @@
 require "rake/testtask"
 require 'erb'
 require "aufgaben/release"
+require "open3"
```
ybiquitous: [IMO] I think better to move `require "open3"` to `lib/tasks/docker/timeout_test.rb` because it shows an acual dependency.

yykamei: Thank you! I agree.

PR #322, opened 2019-10-16
Improve Checkstyle invalid XML error message

ybiquitous:
For example:

> REXML::ParseException · Start tag isn't ended Line: 436 Position: 90112 Last 80 unconsumed characters

But sadly I cannot verify with a smoke test...

### DIFF
--- a/lib/runners/processor/checkstyle.rb
+++ b/lib/runners/processor/checkstyle.rb
@@ -58,7 +58,14 @@ module Runners
 
         stdout, stderr, _ = capture3(analyzer_bin, *dir, *checkstyle_args(config: config_file, excludes: excludes, properties: properties))
 
-        xml_root = REXML::Document.new(stdout).root
+        begin
+          xml_root = REXML::Document.new(stdout).root
+        rescue REXML::ParseException => exn
+          message = exn.message
+          trace_writer.error "Invalid XML output: #{message}"
+          return Results::Failure.new(guid: guid, analyzer: analyzer, message: message)
+        end
+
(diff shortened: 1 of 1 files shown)


ybiquitous:
@yykamei There are no smoke tests for this, but could you review whether the code is correct?

lib/runners/processor/checkstyle.rb
```diff
@@ -58,7 +58,14 @@ def analyze(changes)
 
         stdout, stderr, _ = capture3(analyzer_bin, *dir, *checkstyle_args(config: config_file, excludes: excludes, properties: properties))
 
-        xml_root = REXML::Document.new(stdout).root
+        begin
+          xml_root = REXML::Document.new(stdout).root
+        rescue REXML::ParseException => exn
+          message = exn.message
+          trace_writer.error "Invalid XML output: #{message}"
+          return Results::Failure.new(guid: guid, analyzer: analyzer, message: message)
```
yykamei: OK, handling `REXML::ParseException` is good.

This error is caused by the code of users. So, that's why you use `Results::Failure` instead of `Results::Error`. Is it correct?

ybiquitous: Yes. Honestly, when should we use `Results::Error`?

yykamei: No, if it's obvious, `Failure` is appropriate.

ybiquitous: I see, thanks. 👍

PR #323, opened 2019-10-16
Improve PHPMD custom rule support

ybiquitous:
- Added `custom_rule` option.
- This option supports *glob* format.
- Files specified by this option are **always** analyzed.
- Unless this option, analyses not including custom rule files in a PR would fail.

Usage:

```yaml
linter:
  phpmd:
    rule: custom_ruleset.xml
    custom_rule_path:
      - NoFunctions.php
      - custom/**/*.php
```

### DIFF
--- a/lib/runners/processor/phpmd.rb
+++ b/lib/runners/processor/phpmd.rb
@@ -9,6 +9,7 @@ module Runners
                         suffixes: string?,
                         exclude: string?,
                         strict: boolean?,
+                        custom_rule: array?(string),
                         # DO NOT ADD ANY OPTIONS in `options` option.
                         options: optional(object(
                                             rule: string?,
@@ -46,7 +47,7 @@ module Runners
     end
 
     def prepare_analysis_files(changes, config)
-      delete_unchanged_files(changes, only: target_files(config))
+      delete_unchanged_files(changes, only: target_files(config), except: config[:custom_rule] || [])
 
       trace_writer.message "Changed files:" do
         changes.changed_files.each do |file|
--- a/lib/runners/processor/phpmd.rb
+++ b/lib/runners/processor/phpmd.rb
@@ -9,7 +9,7 @@ module Runners
                         suffixes: string?,
                         exclude: string?,
                         strict: boolean?,
-                        custom_rule: array?(string),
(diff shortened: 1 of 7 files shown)


lib/runners/processor/phpmd.rb
```diff
@@ -46,7 +47,7 @@ def target_files(config)
     end
 
     def prepare_analysis_files(changes, config)
-      delete_unchanged_files(changes, only: target_files(config))
+      delete_unchanged_files(changes, only: target_files(config), except: config[:custom_rule] || [])
```
ybiquitous: [question] Is the usage of `except:` expected one? (this cannot be verified via smoke tests 😓 )

lib/runners/processor/phpmd.rb
```diff
@@ -9,6 +9,7 @@ class Processor::Phpmd < Processor
                         suffixes: string?,
                         exclude: string?,
                         strict: boolean?,
+                        custom_rule: array?(string),
```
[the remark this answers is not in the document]
yykamei: At a glance, I thought `custom_rule` should be specified with rule names instead of file names. `custom_rule_path(s)` is better to me. What do you think?

lib/runners/processor/phpmd.rb
yykamei: Correct :+1:

yykamei: Without `except: config[:custom_rule]`, the `custom_rule` smoke test should fail, so I think it seems OK.

lib/runners/processor/phpmd.rb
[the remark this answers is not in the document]
ybiquitous: Good idea! 👍

lib/runners/processor/phpmd.rb
ybiquitous: Thank you!

lib/runners/processor/phpmd.rb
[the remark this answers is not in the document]
ybiquitous: Fixed via 770aa1a628d0af19feb68fe7e66482b1ea0920c9

PR #339, opened 2019-10-23
[Refactor] Simplify the internal code of GemInstaller

ybiquitous:


lib/runners/ruby/gem_installer.rb
```diff
@@ -68,11 +62,12 @@ def gemfile_content
           lines << "#{source} do"
           specs.each do |spec|
             versions = spec.version
-            sider_constraints = self.constraints[spec.name]
+            sider_constraints = self.constraints[spec.name] || []
+
+            trace_writer.message "Installing `#{spec.name}` gem from #{source}..."
+            trace_writer.message "  Specified version: #{versions.join(', ').presence || 'latest'}"
+            trace_writer.message "  Sider constraints: #{sider_constraints.join(', ').presence || 'none'}"
```
ybiquitous: [note] Use [`presence`](https://api.rubyonrails.org/classes/Object.html#method-i-presence) of ActiveSupport.

PR #345, opened 2019-10-23
Create pull requests to bump an analyzer via GitHub Actions

### LINKED ISSUE #130: Make update PR automtically
For tools without Dependabot update support (e.g. SwiftLint), I'd like to introduce some auto-update PR system.

Perhaps I think that writing own GitHub Actions can realize this.

ybiquitous:
## Summary

- Add a new Rake task `bump_analyzers`.
- Run the task on schedule.

## See also

- https://help.github.com/en/github/automating-your-workflow-with-github-actions/virtual-environments-for-github-actions#default-environment-variables
- https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions

Fixes #130

### DIFF
--- a/.github/workflows/bump_analyzers.yml
+++ b/.github/workflows/bump_analyzers.yml
@@ -1,0 +1,23 @@
+name: Bump analyzers
+
+on:
+  schedule:
+    - cron: "0 0 */1 * *"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
+      - run: |
+          gem install bundler --no-document
+          bundle install --jobs=4 --retry=3 --path=vendor/bundle
+      - run: |
+          bundle exec rake bump_analyzers
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_AUTHOR_NAME: GitHub Actions
+          GITHUB_AUTHOR_EMAIL: actions@github.com
--- a/lib/tasks/bump_analyzers.rb
+++ b/lib/tasks/bump_analyzers.rb
@@ -1,0 +1,192 @@
+require "net/http"
+require "json"
+
+desc "Bump analyzers and create new pull requests"
+task :bump_analyzers do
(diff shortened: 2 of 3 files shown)


ybiquitous:
Generated PRs: #351 #352

.github/workflows/bump_analyzers.yml
```diff
@@ -0,0 +1,23 @@
+name: Bump analyzers
+
+on:
+  schedule:
+    - cron: "0 0 */1 * *"
```
ybiquitous: [note] We should verify the setting after this PR will be merged. If succeeded, we should increase interval days like `0 0 */30 * *`.

lib/tasks/bump_analyzers.rb
```diff
@@ -0,0 +1,192 @@
...
+  end
+
+  # @see https://developer.github.com/v3/git/refs/#list-matching-references
+  def tags
+    @tags ||= call_github_api(:get, "/repos/#{github_repo}/git/refs/tags")
+      .map { |tag| tag.fetch(:ref).delete_prefix("refs/tags/") }
+  end
+
+  def latest_version
+    @latest_version ||= tags.last.delete_prefix("v")
```
yykamei: `tags.last` is really the latest version? I think we should consider pagination when calling GitHub APIs.

ybiquitous: Yes, ideally. But I think the pagination code will be too complex here. Do you have any solutions?

yykamei: Probably, the only solution is to use GitHub API v4... yes, I know it's also complex.

https://developer.github.com/v4/object/repository/

Screenshot from 2019-10-24 11-14-37

ybiquitous: Umm..., the current code seems to work enough, so how about considering after any problems will occur?
(It seems to me that Octokit.rb is easier than the raw GraphQL API 😅 )

yykamei: I thought all GitHub APIs have pagination, but I tried the same endpoint `refs` on my local machine, then GitHub responds without pagination information. It looks pagination does not affect this endpoint.

OK. It works well. Sorry for preventing you.

ybiquitous:
Thank you for reviewing! 😄

PR #359, opened 2019-10-24
Optimize generated Gemfile

ybiquitous:
Reduce duplicates. For example:

## Before

```ruby
source "https://rubygems.org"
source "https://rubygems.org" do
  gem("strong_json", "0.4.0", "<= 0.8")
end
source "https://rubygems.org" do
  gem("meowcop", "1.2.0")
end
```

## After

```ruby
source "https://rubygems.org"

gem "strong_json", "0.4.0", "<= 0.8"
gem "meowcop", "1.2.0"
```

## Why?

- Duplicate sources give a bad effect on downloading gems.
- Harder to debug.

## Note

`specs.group_by(&:source)` has not been working until now. Because the source classes do not have `#eql?` and `#hash` methods, which are needed to use these classes as a Hash key.
See <https://ruby-doc.org/core-2.6.5/Hash.html#class-Hash-label-Hash+Keys> for more details.

https://github.com/sider/runners/blob/72ca9767d87a69f92c6a4093e7fc84369c19e823/lib/runners/ruby/gem_installer.rb#L61

### DIFF
--- a/lib/runners/ruby/gem_installer.rb
+++ b/lib/runners/ruby/gem_installer.rb
@@ -25,7 +25,7 @@ module Runners
       end
 
       def install!
-        gemfile_path.write(gemfile_content.join("\n"))
+        gemfile_path.write(gemfile_content)
 
         trace_writer.header "Installing gems..."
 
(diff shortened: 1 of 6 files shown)


lib/runners/ruby/gem_installer.rb
```diff
@@ -56,39 +56,67 @@ def gemfile_content
...
-
-        specs.group_by(&:source).each do |source, specs|
-          lines << "#{source} do"
-          specs.each do |spec|
-            versions = spec.version
-            sider_constraints = self.constraints[spec.name] || []
-
-            trace_writer.message "Installing `#{spec.name}` gem from #{source}..."
-            trace_writer.message "  Specified version: #{versions.join(', ').presence || 'latest'}"
-            trace_writer.message "  Sider constraints: #{sider_constraints.join(', ').presence || 'none'}"
```
ybiquitous: [note] Instead of each version and constraint, this PR outputs the Gemfile's content once. I think it is easier.

ybiquitous:
@yykamei This PR may give an effect on the current Ruby analyses. Could you please review?

lib/runners/ruby/gem_installer.rb
```diff
@@ -56,39 +56,67 @@ def gemfile_content
...
-            lines << "  gem(#{spec.name.inspect}, #{constraints.map(&:inspect).join(", ")})"
+        lines = ["source #{DEFAULT_SOURCE.inspect}", ""]
+
+        group_specs.each do |source, specs|
+          if source.default?
+            specs.each do |spec|
+              lines << gem(spec, gem_constraints(spec, source))
+            end
+          else
+            lines << ""
```
yykamei: If `group_specs` does not contain default gems but includes only gems with sources, the trace output would write two blank lines in succession like this?

```
source "https://rubygems.org"


source "https://gems.sider.review" do
...
end
```

ybiquitous: Yes, but it's odd. I'll fix it. 💪

ybiquitous: I've pushed to rebased one commit. See below:

https://github.com/sider/runners/pull/359/files#diff-1e80d84d2a63b283b1cff0b5bf5bbe19R67

lib/runners/ruby/gem_installer.rb
```diff
@@ -56,39 +56,67 @@ def gemfile_content
...
-            lines << "  gem(#{spec.name.inspect}, #{constraints.map(&:inspect).join(", ")})"
+        lines = ["source #{DEFAULT_SOURCE.inspect}", ""]
+
+        group_specs.each do |source, specs|
+          if source.default?
+            specs.each do |spec|
+              lines << gem(spec, gem_constraints(spec, source))
+            end
+          else
+            lines << "" if lines.last.present?
```
ybiquitous: [note] Use ActiveSupport to avoid the following type-check error:

> lib/runners/ruby/gem_installer.rb:67:31: NoMethodError: type=(::String | nil), method=empty? (lines.last.empty?)

https://github.com/sider/runners/pull/359/files#diff-2b742c00c45e0f8c8d7f87113d3cbbbfR11

PR #367, opened 2019-10-25
Use pre-installed gems as possible

### LINKED ISSUE #237: Is the default version in Gemfile used?
See https://github.com/sider/runners/pull/231#issuecomment-536911447

ybiquitous:
Fix #237

- Define versions at runtime via `gem list` instead of static versions via `DEFAULT_GEMS`.
- Fix mismatch between `Gemfile` in `Dockerfile` and `DEFAULT_GEMS`.
- Stop [`bundle install --deployment`](https://bundler.io/man/bundle-install.1.html#DEPLOYMENT-MODE) because Runners is always executed in Docker containers (sandbox).

## See also

- https://help.sider.review/getting-started/custom-configuration#gems-option
- https://bundler.io/man/gemfile.5.html
- https://bundler.io/v2.0/man/bundle-install.1.html
- https://guides.rubygems.org/command-reference/#gem-list


images/rubocop/Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gem 'rubocop', '0.75.0'
+gem 'meowcop'
```
ybiquitous: [note] Using pre-installed cache. See:

https://github.com/sider/runners/pull/367/files#diff-2d8efbdb905f4de38c35365f48a62992R76

ybiquitous:
Thanks! 😊

PR #401, opened 2019-10-29
Empower ESLint runner

ybiquitous:
- Change the output from a file to STDOUT (seen in the log). This makes a debug easier.
- Use a custom ESLint formatter because the original JSON formatter includes source code of analyzed files.
- Save useful meta data of issues (link, severity, etc.)
- Remove a severity from a message (this should be formatted on the view layer).
- Not raise an analysis error when no linting files with ESLint v5+.
- Change the recommended minimum version from `4.19.1` to `5.0.0`.

### DIFF
--- a/images/eslint/custom-eslint-json-formatter.js
+++ b/images/eslint/custom-eslint-json-formatter.js
@@ -1,0 +1,32 @@
+// @see https://eslint.org/docs/developer-guide/working-with-custom-formatters
+module.exports = (results, data) => {
+  const meta = data ? data.rulesMeta : null;
+
+  const docs = ruleId => {
+    if (meta && meta[ruleId] && meta[ruleId].docs) {
+      const { category, recommended, url } = meta[ruleId].docs;
+      return { category, recommended, url };
+    } else {
+      return null;
+    }
+  };
+
+  const newResults = results.map(({ filePath, messages }) => ({
+    filePath,
+
+    messages: messages.map(
+      ({ ruleId, severity, message, line, column, endLine, endColumn }) => ({
+        ruleId,
+        severity,
+        message,
+        line,
+        column,
+        endLine,
+        endColumn,
+        docs: docs(ruleId)
+      })
+    )
+  }));
+
+  return JSON.stringify(newResults);
+};
--- a/lib/runners/processor/eslint.rb
+++ b/lib/runners/processor/eslint.rb
@@ -26,6 +26,12 @@ module Runners
                                             quiet: boolean?
                                           ))
                       })
+
+        let :issue, object(
+          severity: string,
+          category: string?,
+          recommended: boolean?,
+        )
       }
     end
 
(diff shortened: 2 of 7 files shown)


lib/runners/processor/eslint.rb
```diff
@@ -186,41 +187,39 @@ def run_analyzer(target_dir, additional_options)
...
         *target_dir
       )
-      # Print text of output.txt for debug
-      if output.file?
-        output_json = output.read
-        trace_writer.message 'Created output.json.'
-      end
 
-      if status.success? || output_json
+      if [0, 1].include? status.exitstatus
```
ybiquitous: > __[Querly]__ Process#exitstatus may be nil ([link](https://sider.review/gh/repos/202486306/pulls/401#issue-4769376))
>
> <details>
>
>
>
>
> </details>

Maybe no problem.

lib/runners/processor/eslint.rb
```diff
@@ -134,50 +140,45 @@ def quiet(config)
...
+          message = details[:message]
+
+          unless id
+            trace_writer.message("No ID found! - #{message}")
+            id = Digest::SHA1.hexdigest(message)
           end
+
+          yield Issue.new(
+            path: path,
+            location: Location.new(start_line: details[:line]),
```
yykamei: Are `endLine` and `endColumn` always set? If so, to set all parameters to `Location` would be better (e.g. We may introduce a new feature to show issues with a range of lines).

However, I'm not sure about `endLine` and `endColumn` properties of `message.` They sometimes become `undefined,` this suggestion doesn't make sense.

images/eslint/custom-eslint-json-formatter.js
```diff
@@ -0,0 +1,32 @@
...
+
+    messages: messages.map(
+      ({ ruleId, severity, message, line, column, endLine, endColumn }) => ({
+        ruleId,
+        severity,
+        message,
+        line,
+        column,
+        endLine: endLine,
+        endColumn: endColumn,
```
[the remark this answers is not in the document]
ybiquitous: Oh, my bad. I'll fix it 👍

lib/runners/processor/eslint.rb
ybiquitous: Nice question. I also want to set all parameters, but I'm not confident whether ESLint plugins satisfy the following constraint:

https://github.com/sider/runners/blob/113a394098834bd077b7adf49053cbe4a480fdea/lib/runners/schema/result.rb#L6-L8

yykamei: I see. So far, our system with only `start_line` works well. Let's keep it as is.

ybiquitous: If respecting the constraint above, I'll fix as follow:

```ruby
case
when start_line && start_column && end_line && end_column
  Location.new(start_line: start_line, start_column: start_column, end_line: end_line, end_column: end_column)
when start_line && end_line
  Location.new(start_line: start_line, end_line: end_line)
else
  Location.new(start_line: start_line)
end
```

yykamei: Looks greate!

ybiquitous: Thank you. I'll fix also smoke tests.

ybiquitous:
I've applied all reviews and squashed to one commit. When CI passes, I'll merge it.

PR #412, opened 2019-10-30
[Chore] Remove unused Ruby signatures

ybiquitous:
Steeep is under development. I think increasing more signatures is not good at least now.


Rakefile
```diff
@@ -33,7 +33,6 @@ task :default => [:test, :typecheck]
 task :typecheck do
   # FIXME: Must check all files.
   files = Dir.glob("lib/runners/**/*.rb").reject { |f| f.include? "processor/" }
-  files += Dir.glob("lib/**/cppcheck/**/*.rb")
```
ybiquitous: [note] Not working...

```
irb(main):001:0> Dir.glob("lib/**/cppcheck/**/*.rb")
=> []
```

PR #415, opened 2019-10-30
Add Options module and delete insecure options

### LINKED ISSUE #409: Deprecate insecure options and make it receive such information from ENV.
Currently, Runners support the following options, which are insecure because they are visible to the container hosts.

* `--head`
* `--head-key`
* `--base`
* `--base-key`
* `--ssh-key`

To securely run Runners, I want to deprecate them and introduce an interface to receive such information from the environment variable.
These parameters are passed via JSON string:

```
RUNNERS_OPTIONS={"head_url": "http://...", "head_url_key": "secret", ...}
```

Runners parse the environment variable like this:

```ruby
options = JSON.parse(ENV["RUNNERS_OPTIONS"])
...
```

Perhaps, the `--output` option also should be extracted from the variable.

yykamei:
Resolve #409

## BREAKING CHANGES

This includes breaking changes in the command line interface. These options are no longer available in favor of `ENV['RUNNERS_OPTIONS']`.

* `--base`
* `--base-key`
* `--head`
* `--head-key`
* `--ssh-key`
* `--working`
* `--output`

Instead, you have to set the environment variable `RUNNERS_OPTIONS` like this:

```shell
RUNNERS_OPTIONS='{"source": {"head": "http://example.com/head"},
"ssh_key": "ssh_key", "outputs": ["stdout", "s3://bucket/key"]}'
```

By the way, I also deleted the `--working` option because we have never used it.


lib/runners/cli.rb
```diff
@@ -6,87 +6,32 @@ class CLI
...
-      raise "head_key is given but head is missing" if !head && head_key
-      validate_analyzer!
-      self
-    end
-
-    def validate_analyzer!
       raise "--analyzer is required" unless analyzer
       raise "The specified analyzer is not supported" unless processor_class
+      self
     end
```
ybiquitous: [suggestion] Unifying the validation logic to one place seems better to me. For example:

```ruby
    def validate_options!
      raise "--analyzer is required" unless analyzer
      raise "The specified analyzer is not supported" unless processor_class

      "RUNNERS_OPTIONS".tap do |key|
        value = ENV[key]
        value.present? or raise "The environment variable #{key.inspect} is required"
        @options = Options.new(stdout, stderr, value)
      end

      self
    end
```

lib/runners/options.rb
```diff
@@ -0,0 +1,47 @@
...
+              end
+    end
+
+    private
+
+    def parse_options
+      ENV['RUNNERS_OPTIONS'].yield_self do |val|
+        Schema::Options.payload.coerce(JSON.parse(val, symbolize_names: true))
+      end
+    end
```
ybiquitous: [suggestion]

```ruby
    def parse_options(value)
      Schema::Options.payload.coerce(JSON.parse(value, symbolize_names: true))
    end
```

See https://github.com/sider/runners/pull/415/files#r340522613

lib/runners/options.rb
```diff
@@ -0,0 +1,47 @@
+module Runners
+  class Options
```
ybiquitous: Introducing `Options` class looks good! 👍

lib/runners/schema/options.rb
```diff
@@ -0,0 +1,18 @@
+module Runners
+  module Schema
+    Options = _ = StrongJSON.new do
+      # @type self: Types::Options
+
+      let :source, enum?(
```
ybiquitous: [question] `source` is required?

```suggestion
      let :source, enum(
```

lib/runners/options.rb
```diff
@@ -0,0 +1,47 @@
...
+                    stderr
+                  when /^s3:/
+                    Runners::IO::AwsS3.new(output)
+                  else
+                    raise "Invalid output option. You included '#{output}'"
+                  end
+                end
+                Runners::IO.new(*ios)
+              end
+    end
```
ybiquitous: [suggestion] I feel the `io` method is a bit complex. How about initializing in the constructor as follow?

```ruby
      outputs = options[:outputs] || []
      @io = if outputs.empty?
              Runners::IO.new(stdout)
            else
              ios = outputs.map do |output|
                case output
                when 'stdout'
                  stdout
                when 'stderr'
                  stderr
                when /^s3:/
                  Runners::IO::AwsS3.new(output)
                else
                  raise "Invalid output option. You included '#{output}'"
                end
              end
              Runners::IO.new(*ios)
            end
```

In the above example, validation timing will be the same as `CLI#validate_options!`.

lib/runners/testing/smoke.rb
```diff
@@ -114,11 +114,16 @@ def with_data_container
...
-        dir = data_smoke_path + name
-        commands = %W[docker run --rm --volumes-from #{data_container} #{docker_image} --head=#{dir.expand_path}]
-        ssh_key = config.ssh_key
-        commands << "--ssh-key=#{dir.expand_path + ssh_key}" if ssh_key
-        commands << "test-guid"
+        dir = data_smoke_path / name
+        ssh_key = config.ssh_key&.yield_self do |file|
+          Dir.mktmpdir do |tmp_dir|
+            tmp_ssh_key_path = Pathname(tmp_dir) / 'ssh_key'
+            system "docker cp #{data_container}:#{dir.expand_path / file} #{tmp_ssh_key_path}"
```
ybiquitous: [suggestion] I noticed it now but I think `system(*args, exception: true)` is safer.

How about adding the following helper method?

```ruby
def run!(command_args)
  system(*command_args, exception: true)
end

# usage
run! %W[docker cp #{data_container}:#{dir.expand_path / file} #{tmp_ssh_key_path}]
```

See http://ruby-doc.org/core-2.6.5/Kernel.html#method-i-system

test/cli_test.rb
```diff
@@ -14,49 +14,27 @@ def stderr
...
+    with_runners_options_env(source: { head: 'http://example.com/head' }) do
+      assert_raises RuntimeError do
+        # analyzer is missing
+        CLI.new(argv: %w(test-guid), stdout: stdout, stderr: stderr).validate_options!
+      end
+
+      assert_raises RuntimeError do
+        # Invalid analyzer
+        CLI.new(argv: %w(--analyzer=FOO test-guid), stdout: stdout, stderr: stderr).validate_options!
+      end
```
ybiquitous: [suggestion] I think it good to verify also an error message when using `assert_raises RuntimeError`, because we cannot know where `RuntimeError` is raised actually (maybe within libraries).

For example:

```ruby
      assert_raises RuntimeError do
        CLI.new(argv: %w(--analyzer=FOO test-guid), stdout: stdout, stderr: stderr).validate_options!
      end.tap do |error|
        assert_equal "Not found processor class with 'FOO'", error.message
      end
```

test/test_helper.rb
```diff
@@ -22,4 +22,12 @@ def data(file)
...
     Pathname(__dir__).join("incorrect_yarn_data", file)
   end
+
+  def with_runners_options_env(hash)
+    backup = ENV['RUNNERS_OPTIONS']
+    ENV['RUNNERS_OPTIONS'] = JSON.dump(hash)
+    yield
+  ensure
+    ENV['RUNNERS_OPTIONS'] = backup
+  end
```
ybiquitous: 👍

lib/runners/cli.rb
yykamei: Unifying seems good :+1: 

How about deleting `validate_options!`? I think it's sufficient to raise exceptions in `#initialize,` and it looks weird to set `@options` in `validate_options!`.
In other words, I mean writing such code:

```ruby
    def initialize(argv:, stdout:, stderr:)
      @stdout = stdout
      @stderr = stderr

      OptionParser.new do |opts|
        opts.on("--analyzer=ANALYZER") do |analyzer|
          @analyzer = analyzer
        end
      end.parse!(argv)
      @guid = _ = argv.shift

      raise "--analyzer is required" unless analyzer
      raise "The specified analyzer is not supported" unless processor_class
      @options = Options.new(stdout, stderr)
    end
```

lib/runners/testing/smoke.rb
yykamei: Thank you! It's beautiful.

yykamei: See b66a25b

test/cli_test.rb
yykamei: See 06bca80

lib/runners/options.rb
yykamei: I fixed it with fb99a28

yykamei: I'm dealing with the failure of CI...

lib/runners/cli.rb
ybiquitous: Agree 👍

yykamei:
I will rebase this PR after the CI on CircleCI succeeds.

lib/runners/cli.rb
```diff
@@ -6,87 +6,28 @@ class CLI
...
-        opts.on("--working=WORKING") do |working|
-          @working_dir = working
-        end
-        opts.on("--output=OUTPUT",
-                "The output destination. Currently, the valid formats are 'stdout', 'stderr', and 's3://BUCKET/KEY'") do |output|
-          @outputs << output
-        end
       end.parse!(argv)
-
       @guid = _ = argv.shift
```
ybiquitous: By the way, `guid` is required? is the check needed?

ybiquitous: Also, the assignment to `_` is needed?

lib/runners/testing/smoke.rb
```diff
@@ -105,22 +105,31 @@ def unify_result(result, pattern, out)
...
+          end
+        end
+        runners_options = JSON.dump(source: { head: dir.expand_path }, ssh_key: ssh_key)
+        commands = %W[docker run --rm --volumes-from #{data_container} --env RUNNERS_OPTIONS='#{runners_options}' #{docker_image} test-guid]
         commands.join(" ")
       end
 
+      def system!(*command_args)
+        system(*command_args, exception: true)
+      end
```
ybiquitous: 👍

lib/runners/cli.rb
yykamei: I try to delete `_` and validate it. Thank you.

sig/runners/options.rbi
```diff
@@ -0,0 +1,14 @@
...
+  attr_reader head: String
+  attr_reader head_key: String?
+  attr_reader base: String?
+  attr_reader base_key: String?
+  attr_reader outputs: Array<String>
+  attr_reader ssh_key: String?
+
+  def initialize: (::IO, ::IO) -> any
+  def io: () -> Runners::IO
+  def parse_options: () -> any
```
ybiquitous: How about this?

```suggestion
  def parse_options: () -> Hash<Symbol, any>
```

sig/polyfills/kernel.rbi
```diff
@@ -1,7 +1,7 @@
 extension Kernel (Polyfill)
   def abort: (String) -> void
   def exit: (Integer) -> void
-  def system: (String) -> any
+  def system: (*any) -> any
```
ybiquitous: How about this?

```suggestion
  def system: (*String) -> void
```

ybiquitous: How about this?

```suggestion
  def system: (*String) -> void
```

lib/runners/cli.rb
yykamei: `_` is still required because I want to make sure the `@guid` is `String` (not `String | nil`).

ybiquitous: I see 👍

PR #449, opened 2019-11-01
Add Runners version and set it to result

ybiquitous:
Why?
----

If Runners client can receive its version, the client can do something different according to the some versions.
Also, I think that saving Runners versions will be useful to debug.

Notes
-----

- The `rake release` task will update the version automatically.

### DIFF
--- a/lib/runners/version.rb
+++ b/lib/runners/version.rb
@@ -1,0 +1,3 @@
+module Runners
+  VERSION = "0.8.1".freeze
+end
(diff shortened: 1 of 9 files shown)


lib/runners/version.rb
```diff
@@ -0,0 +1,3 @@
+module Runners
+  VERSION = "0.7.5".freeze
```
ybiquitous: [note] Please update here when releasing a new version!

ybiquitous:
@yykamei What do you think about this?

PR #497, opened 2019-11-15
Flush content whenever its size reaches the buffer max limit

yykamei:
Currently, the trace of the analysis is flushed at the end of the
analysis. However, I want to see the on-going trace log even though the
analysis has not yet finished. Also, I want to know whether the analysis
has started or not, so at the beginning of the analysis, I inserted the
`io.flush!` procedure.

### DIFF
--- a/lib/runners/cli.rb
+++ b/lib/runners/cli.rb
@@ -38,6 +38,7 @@ module Runners
     end
 
     def run
+      io.flush! # Write or upload empty content to let the caller of Runners notice the beginning of the analysis.
       with_working_dir do |working_dir|
         writer = JSONSEQ::Writer.new(io: io)
         trace_writer = TraceWriter.new(writer: writer)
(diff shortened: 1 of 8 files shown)


lib/runners/cli.rb
```diff
@@ -38,6 +38,7 @@ def processor_class
     end
 
     def run
+      io.finalize! # Write or upload empty content to let the caller of Runners notice the beginning of the analysis.
```
ybiquitous: [IMO] `finalize!` seems inaccurate to me... (not final)

yykamei: I agree. How about adding `alias flush! finalize!`?

ybiquitous: Adding alias seems confusing to me? How about renaming `finalize!` to `flush!`?

yykamei: Good :+1: I will fix it like that.

PR #501, opened 2019-11-18
Wrap Workspace with Harness

yykamei:
`Workspace.open` sometimes raises the error `DownloadError`.
When it happens, the result won't be output in the trace.

This PR changed `Harness` and wrapped `Workspace` with `Harness`.
Since `Harness#ensure_result` catches any errors, `DownloadError`
will be handled and the `trace_writer` of `Harness` surely writes
the result.

I think it's easy to read this PR with the settings "Hide whitespace changes."

### DIFF
--- a/test/cli_test.rb
+++ b/test/cli_test.rb
@@ -120,6 +120,25 @@ class CLITest < Minitest::Test
     end
   end
 
+  def test_run_when_download_error_happens
+    mktmpdir do |head_dir|
+      with_runners_options_env(source: { head: head_dir }) do
+        cli = CLI.new(argv: ["--analyzer=rubocop", "test-guid"], stdout: stdout, stderr: stderr)
+        cli.instance_variable_set(:@processor_class, TestProcessor)
+        mock(Runners::Workspace).open.with_any_args { raise Runners::Workspace::DownloadError }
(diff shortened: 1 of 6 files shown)


yykamei:
@ybiquitous What do you think about this PR?

test/cli_test.rb
```diff
@@ -120,6 +120,25 @@ def test_run_when_sider_yml_is_broken
...
+
+        output = stdout.string
+        reader = JSONSEQ::Reader.new(io: StringIO.new(output), decoder: -> (string) { JSON.parse(string, symbolize_names: true) })
+        objects = reader.each_object.to_a
+
+        assert objects.find { |hash| hash.dig(:result, :type) == 'error' }
+        assert objects.find { |hash| hash.dig(:result, :class) == 'Runners::Workspace::DownloadError' }
+      end
+    end
+  end
```
yykamei: I added an additional test.

By the way,  [`#test_run_when_sider_yml_is_broken`](https://github.com/sider/runners/pull/501/files#diff-7f1b37e427aa850ea29f1674870e5977L102) assumes the result is "error," but I think it should be "failure." I created an issue [here](#503).

PR #504, opened 2019-11-19
Mark the result "failure" when `sider.yml` is broken

### LINKED ISSUE #503: The result should be "failure" instead of "error" when `sider.yml` is broken
Current implementation marks the result "error" when broken `sider.yml` is included in users' source tree. I believe this is not Sider's fault. So, the result should be "failure."

yykamei:
Fix #503


test/cli_test.rb
```diff
@@ -113,6 +113,7 @@ def test_run_when_sider_yml_is_broken
         reader = JSONSEQ::Reader.new(io: StringIO.new(output), decoder: -> (string) { JSON.parse(string, symbolize_names: true) })
         objects = reader.each_object.to_a
 
+        assert objects.find { |hash| hash.dig(:result, :type) == 'failure' }
```
yykamei: The test would fail. I got the following message, but I can't find out why it happens :scream: 

```
#<NoMethodError: undefined method `CIConfigBroken' for #<Runners::Processor::TestProcessor:0x0000555f09ea7708>\nDid you mean?  ci_config
```

ybiquitous: ??? 👇 

```diff
-raise CIConfigBroken(message)
+raise CIConfigBroken.new(message)
```

yykamei: Thank you!

lib/runners/processor.rb
```diff
@@ -26,8 +28,17 @@ def initialize(guid:, working_dir:, git_ssh_path:, trace_writer:)
...
-            trace_writer.error("Parse error occurred: #{exn}")
-            raise
+            message = <<~MESSAGE
+              Parse error occurred: #{exn}
+
+              Your configuration file is:
+
+              ```
+              #{ci_config_path.read}
+              ```
```
ybiquitous: [question] The Sider current log page does NOT render in markdown. Will you change also front-end code?

yykamei: No. I thought the message would be rendered as Markdown. Thank you for telling that.

lib/runners/processor.rb
```diff
@@ -26,8 +28,9 @@ def initialize(guid:, working_dir:, git_ssh_path:, trace_writer:)
               trace_writer.ci_config(conf)
             end
           rescue Psych::SyntaxError => exn
-            trace_writer.error("Parse error occurred: #{exn}")
-            raise
+            message = "Parse error occurred: #{exn}"
+            trace_writer.error(message)
```
ybiquitous: [question] Will the same messages of `trace_writer.error` and `Results::Failure#message` be output to the log page?

lib/runners/harness.rb
```diff
@@ -78,6 +78,8 @@ def ensure_result
             raise InvalidResult.new(result: result)
           end
         end
+      rescue Runners::Processor::CIConfigBroken => exn
+        Results::Failure.new(guid: guid, message: exn.message)
```
ybiquitous: [question] Will you add a test in `harntess_test.rb` to verify the behavior?

lib/runners/processor.rb
yykamei: No, I'll fix it.

lib/runners/harness.rb
yykamei: I'll address it.

lib/runners/processor.rb
yykamei: Thank you for commenting. I fixed it with 96dcb9c.

lib/runners/harness.rb
yykamei: Is 411599f proper?

ybiquitous: Thanks. In addition, how about adding an assertion for `result.message`?

Also, will `exn.message` be seen by users on the log page?

yykamei: Thank you. How about 84b24ca?

ybiquitous: I think not good to show an exception class to users. So how about as follow?

```ruby
rescue Psych::SyntaxError => exn
  message = "Your `#{ci_config_name}` file may be broken (line #{exn.line}, column #{exn.column})."
  trace_writer.error message #=> This call may need to output the log page?
  raise CIConfigBroken.new(message)
end
```

http://ruby-doc.org/stdlib-2.6.5/libdoc/psych/rdoc/psych/SyntaxError.html

test/harness_test.rb
```diff
@@ -134,6 +134,20 @@ def test_run_when_root_dir_not_found
...
+  def test_run_when_ci_config_is_broken
+    with_working_dir do |working_dir|
+      (working_dir / "sider.yml").write('1: 1:')
+      with_options do |options|
+        harness = Harness.new(guid: SecureRandom.uuid, processor_class: TestProcessor,
+                              options: options, working_dir: working_dir, trace_writer: trace_writer)
+
+        result = harness.run
+        assert_instance_of Results::Failure, result
+        assert result.message =~ /Parse error occurred/
```
ybiquitous: Maybe in this case, `assert_match %r{Parse error occurred}, result.message` is better for a friendly assertion message.

If a file name its unnecessary, I think better to use `assert_equal`.

lib/runners/harness.rb
```diff
@@ -78,6 +78,8 @@ def ensure_result
             raise InvalidResult.new(result: result)
           end
         end
+      rescue Runners::Processor::CIConfigBroken => exn
```
ybiquitous: The suggestion is out of the scope of this PR, but we may introduce a hierarchy of error classes as follow:

```ruby
# lib/runners/errors.rb
module Runners
  class Error < StandardError; end
  class UserError < Error; end
  class SystemError < Error; end
end
```

By doing this, we will not refer to concrete error classes:

```diff
-rescue Runners::Processor::CIConfigBroken => exn
+rescue UserError => exn
```

lib/runners/harness.rb
yykamei: > Also, will `exn.message` be seen by users on the log page?

I thought your suggestion is "Do not show any messages on trace," but probably I misunderstood. I'll fix it as you said.

ybiquitous: Thanks. I think better not to show internal Ruby exception classes to users **as possible**.
The information seems almost unuseful to users and may cause security issues.

lib/runners/harness.rb
yykamei: I created an issue [here](#506).

ybiquitous: Thank you!

test/harness_test.rb
yykamei: Thank you. `assert_equal` is better. I fixed with it.

lib/runners/harness.rb
yykamei: I agree. I followed the original implementation without considering.

PR #516, opened 2019-11-22
Refactor Runners::Workspace

yykamei:
I redefined `Workspace.open` as `Workspace#open` because I wanted to
reduce parameters of methods and wanted to use instance variables.


yykamei:
With the exception of SonarCloud, the CI passed. Do you have any comments on this PR? @ybiquitous

lib/runners/workspace.rb
```diff
@@ -1,85 +1,82 @@
...
 
             trace_writer.message "Copying head to working dir..." do
               FileUtils.copy_entry(head_path, working_dir)
             end
 
-            yield new(working_dir: working_dir, base_dir: base_path, head_dir: head_path, git_ssh_path: ssh_path)
+            trace_writer.header "Calculating changes between head and base"
+            trace_writer.message "Running..." do
+              yield git_ssh_path, Changes.calculate(base_dir: base_path, head_dir: head_path, working_dir: working_dir)
+            end
```
ybiquitous: [question] Do not we need to call `Changes.calculate` within `Processor#push_root_dir` block?

lib/runners/workspace.rb
```diff
@@ -1,85 +1,82 @@
...
 
-          config_path = ssh_dir + "/config"
-          key_path = ssh_dir + "/key"
-          git_ssh_path = ssh_dir + "/run.sh"
+          config_path = Pathname(dir) / 'config'
+          key_path = Pathname(dir) / 'key'
+          script_path = Pathname(dir) / 'run.sh'
 
-          File.write(config_path, <<~SSH_CONFIG)
+          config_path.write(<<~SSH_CONFIG)
```
ybiquitous: [question] Is `perm: 0600` necessary also here?

```suggestion
          config_path.write(<<~SSH_CONFIG, perm: 0600)
```

lib/runners/workspace.rb
yykamei: I don't think `Processor#push_root_dir` is necessary to call `Changes.calculate`. If it were necessary, `changes_test.rb` also should be written with `#push_root_dir`.

ybiquitous: OK 👌

lib/runners/workspace.rb
yykamei: I don't know why it's been working, but to set permission is required according to ssh_config(5). I'll set it.

lib/runners/workspace.rb
```diff
@@ -89,7 +86,7 @@ def self.prepare_in_dir(type, source, key, dir, trace_writer:)
         Retryable.retryable(
           tries: 5,
           on: [OpenSSL::Cipher::CipherError, OpenSSL::SSL::SSLError, DownloadError, Net::OpenTimeout, Errno::ECONNRESET, SocketError, Net::HTTPForbidden],
-          sleep: -> (n) { n + 1 },
+          sleep: retryable_sleep,
```
ybiquitous: [question]

1. Do we need to wrap with [`method`](http://ruby-doc.org/core-2.6.5/Object.html#method-i-method) as follow?

```suggestion
          sleep: method(:retryable_sleep),
```

2. I feel that `attr_reader :retryable_sleep` is a bit verbose and confusing. How about adding a private method like this?

```ruby
def retryable_sleep(n)
  n + 1
end
```

lib/runners/workspace.rb
ybiquitous: [suggestion] This trace block seems incorrect to me and "Running..." seems a bit verbose. How about this?

```ruby
changes = trace_writer.message "Calculating changes between head and base..." do
  Changes.calculate(base_dir: base_path, head_dir: head_path, working_dir: working_dir)
end

yield git_ssh_path, changes
```

Here is the current output:
image

lib/runners/workspace.rb
yykamei: If I continue using the `:retryable_sleep` as a `Proc` value, it seems to work without `method`, but declaring it as an instance method looks good. I'll fix it as you said.

lib/runners/workspace.rb
yykamei: OK

lib/runners/workspace.rb
```diff
@@ -1,85 +1,80 @@
...
         Dir.mktmpdir do |base_dir|
           base_path = Pathname(base_dir)
           Dir.mktmpdir do |head_dir|
             head_path = Pathname(head_dir)
 
             trace_writer.header "Setting up source code"
 
-            prepare_in_dir(:base, base, base_key, base_path, trace_writer: trace_writer) if base
-            prepare_in_dir(:head, head, head_key, head_path, trace_writer: trace_writer)
+            if base = options.base
```
ybiquitous: [IMO] This is my personal opinion. I feel that `if a = b` format is confusing. For example, `if a = b` and `if a == b` is very similar, so reading the code intention is a bit hard.

lib/runners/workspace.rb
ybiquitous: Is calling `yield` within the block `trace_writer.message "Calculating changes between head and base..." do` incorrect?

This `yield` can be a long process...

lib/runners/workspace.rb
```diff
@@ -1,85 +1,80 @@
...
     end
 
-    def self.prepare_ssh(ssh_key, trace_writer:)
-      trace_writer.header "Setting up SSH config"
+    private
 
-      if ssh_key
-        Dir.mktmpdir do |ssh_dir|
+    def prepare_ssh
+      if ssh_key = options.ssh_key
```
ybiquitous: Same here as https://github.com/sider/runners/pull/516/files#r349983381

sig/runners/workspace.rbi
```diff
@@ -0,0 +1,17 @@
+type prepare_type = :base | :head
+
+class Runners::Workspace
+  attr_reader options: Options
+  attr_reader working_dir: Pathname
+  attr_reader trace_writer: TraceWriter
+  attr_reader retryable_sleep: ^(Integer) -> Integer
```
ybiquitous: Change to `def retryable_sleep`?

lib/runners/workspace.rb
yykamei: I don't have any opinions about that. I'll adjust it to your suggestion.

lib/runners/workspace.rb
ybiquitous: What do you think?

yykamei: Sorry, I overlooked it. I think it's not incorrect, but it looks easier to get it out of the block.

yykamei: Fixed with 92aa445

yykamei:
I ignore the status by SonarCloud because Duplication Code problems are the issues for the smoke tests. Of course, our smoke test contains many duplication codes, and there are no problems. I wanted to tweak the SonarCloud, but I don't have enough permission to do that.

PR #520, opened 2019-11-25
[Refactor] Add Workspace::HTTP and Workspace::File

yykamei:
We're planning to implement a feature that is fetching source code
via Git. Before doing that, I want to separate the current
implementation with two part, `Workspace::HTTP` and `Workspace::File`.


lib/runners/options.rb
```diff
@@ -1,18 +1,33 @@
...
-      @head = options.dig(:source, :head)
-      @head_key = options.dig(:source, :head_key)
-      @base = options.dig(:source, :base)
-      @base_key = options.dig(:source, :base_key)
+      source_params = options[:source]
+      @source = if source_params.key?(:git_http_url)
+                  GitSource.new(**source_params)
+                else
+                  ArchiveSource.new(**source_params)
+                end
```
yykamei: :thinking: Perhaps, `@source` don't have to be a union type. I will change it and declare `@archive_source` and `@git_source`.

yykamei: I think the current implementation is better than the one wich `@archive_source` and `@git_source` because the will be an optional value and we have to consider they are `nil` or not. I won't fix it.

lib/runners/workspace.rb
```diff
@@ -23,12 +41,7 @@ def open
             head_path = Pathname(head_dir)
 
             trace_writer.header "Setting up source code"
-
-            base = options.base
-            if base
-              prepare_in_dir(:base, base, options.base_key, base_path)
-            end
-            prepare_in_dir(:head, options.head, options.head_key, head_path)
+            prepare_source(head_path, base_path)
```
ybiquitous: [IMO] I feel it understandable to split each method call for head and base.

For example:

```ruby
if options.source.base?
  trace_writer.message "Preparing base commit tree..."
  prepare_source(base_path)
end

trace_writer.message "Preparing head commit tree..."
prepare_source(head_path)
```

lib/runners/workspace.rb
```diff
@@ -2,6 +2,24 @@ module Runners
...
+      when Runners::Options::ArchiveSource
+        case
+        when source.http?
+          Workspace::HTTP.new(options: options, working_dir: working_dir, trace_writer: trace_writer)
+        when source.file?
+          Workspace::File.new(options: options, working_dir: working_dir, trace_writer: trace_writer)
+        else
+          raise ArgumentError, "The specified options #{options.inspect} is not supported"
+        end
+      when Runners::Options::GitSource
```
ybiquitous: [nitpick] It seems better to omit the top-level module `Runners::` if possible.

```diff
-when Runners::Options::GitSource
+when Options::GitSource
```

lib/runners/workspace.rb
```diff
@@ -46,6 +59,15 @@ def open
...
 
+    def as_archive_source
+      source = options.source
+      if source.instance_of?(Runners::Options::ArchiveSource)
+        source
+      else
+        raise "#{source.inspect} is not ArchiveSource"
+      end
+    end
+
```
ybiquitous: [nitpick] How about doing this?

```ruby
def archive_source
  @archive_source ||=
    begin
      source = options.source
      if source.instance_of?(Runners::Options::ArchiveSource)
        source
      else
        raise "#{source.inspect} is not ArchiveSource"
      end
    end
end

# usage
head_src = archive_source.head.yield_self(&convert_to_pathname)
base_src = archive_source.base&.yield_self(&convert_to_pathname)
```

sig/runners/workspace.rbi
```diff
@@ -5,13 +5,25 @@ class Runners::Workspace
   attr_reader working_dir: Pathname
   attr_reader trace_writer: TraceWriter
 
+  def self.prepare!: (options: Options, working_dir: Pathname, trace_writer: TraceWriter) -> instance
```
ybiquitous: [IMO] It seems a bit inconsistent that the only `prepare!` method ends with `!`. Because other methods like `prepare_ssh` or `prepare_source` do not end with it.

sig/runners/options.rbi
```diff
@@ -1,7 +1,7 @@
 class Runners::Options
   attr_reader stdout: ::IO
   attr_reader stderr: ::IO
-  attr_reader head: String
+  attr_reader source: Runners::Options::GitSource | Runners::Options::ArchiveSource
```
ybiquitous: [nitpick] Can omit `Runners::` as follow:

```suggestion
  attr_reader source: Options::GitSource | Options::ArchiveSource
```

sig/runners/workspace.rbi
```diff
@@ -5,13 +5,25 @@ class Runners::Workspace
...
   def initialize: (options: Options, working_dir: Pathname, trace_writer: TraceWriter) -> any
   def open: <'a> () { (Pathname?, Changes) -> 'a } -> 'a
   def prepare_ssh: <'a> () { (Pathname?) -> 'a } -> 'a
-  def prepare_in_dir: (prepare_type, String, String?, Pathname) -> void
+  def prepare_source: (Pathname, Pathname) -> void
   def decrypt: <'a> (Pathname, String?) { (Pathname) -> 'a } -> 'a
   def decrypt_by_openssl: (Pathname, String, Pathname) -> void
   def extract: (Pathname, Pathname) -> void
+  def as_archive_source: () -> Runners::Options::ArchiveSource
```
ybiquitous: [nitpick] Also here:

```suggestion
  def as_archive_source: () -> Options::ArchiveSource
```

yykamei:
Thank you so much :smiley: I'll address them :muscle:

sig/runners/workspace.rbi
yykamei: Fixed with 8dad8f5

sig/runners/workspace.rbi
yykamei: Nice. I'll rename it with `prepare` :+1:

lib/runners/workspace.rb
yykamei: I agree with you. To call the method for each source code is easy to understand. However,
When `prepare_source(head_path)` is run, `prepare_source` doesn't know which source code it should use because the passed arg is just a `Pathname`. To make sure that, `prepare_source` have to be declared with additional arguments, like `head` and `key`.

I'm going to implement `Workspace::Git`, and `prepare_source` of this new class cannot receive the same parameters with `Workspace::HTTP` or `Workspace::File`. For example, `key` parameter is not able to be passed to `Workspace::Git#prepare_source` since it doesn't decrypt files. So, I decided to call `prepare_source(head_path, base_path)` in Workspace#open`.

lib/runners/workspace.rb
yykamei: Beautiful :tulip:

yykamei: Replaced it via a97442b

lib/runners/workspace.rb
ybiquitous: Then, how about splitting the method?

```ruby
if options.source.base?
  trace_writer.message "Preparing base commit tree..."
  prepare_base_source(base_path)
end

trace_writer.message "Preparing head commit tree..."
prepare_head_source(head_path)
```

yykamei: With your suggestion, the implementation in `Workspace::HTTP` would be like this. Is it correct?

```ruby
def prepare_head_source(head_dest)
  head_uri = URI.parse(archive_source.head)
  provision(:head, head_uri, head_dest, archive_source.head_key)
end

def prepare_base_source(base_dest)
  base_uri = archive_source.base&.yield_self { |v| URI.parse(v) }
  raise "base_uri must not be nil" unless base_uri
  provision(:base, base_uri, base_dest, archive_source.base_key)
end
```

In my opinion, it might be redundant to use condition not only when calling `prepare_base_source` but also at the `prepare_base_source` like `unless base_uri`. However, if the above implementation of `prepare_head_source` and `prepare_base_source` are reasonable, I'll change my code the way your suggestion :smiley:

ybiquitous: Maybe I think the implementation will be following ⬇️ 

```ruby
def prepare_head_source(head_dest)
  provision(URI.parse(archive_source.head), head_dest, archive_source.head_key)
end

def prepare_base_source(base_dest)
  provision(URI.parse(archive_source.base), base_dest, archive_source.base_key)
end
```

I think it unnecessary to check whether `archive_source.base` is empty because an error will occur somewhere on downloading.
Checking is a caller's responsibility.

yykamei: :thinking: `URI.parse(archive_source.base)` would be issued by Steep, wouldn't it?

ybiquitous: Umm..., checking seems necessary for Steeep 😢 

```ruby
base = archive_source.base or raise "base must not be nil"
URI.parse(base)
```

yykamei: Thank you. Looks good :+1:

yykamei: I changed my code and ran `typecheck`.

```ruby
    def prepare_base_source(dest)
      base = archive_source.base or raise "base must not be nil"
      provision(Pathname(URI.parse(base).path).realpath, dest, archive_source.base_key)
    end
```

:thinking: `or raise ...` doesn't work in the local variable assignment?

```
lib/runners/workspace/file.rb:7:35: ArgumentTypeMismatch: receiver=::URI.class constructor, expected=::String, actual=(::String | nil) (base)
```

yykamei: This `base = _ = archive_source.base` works, but I know it's weird.

ybiquitous: The Steep's type inference for `or raise` does not work...
I agree `_ = ` is weird. Using `unless` may be better as you suggest as follow:

```ruby
raise "base must not be nil" unless base
```

yykamei: Oh, this still doesn't work.

```ruby
base = archive_source.base
raise "base must not be nil" unless base
provision(URI.parse(base), dest, archive_source.base_key)
```

But, the following works...

```ruby
base = archive_source.base
if base
  provision(URI.parse(base), dest, archive_source.base_key)
else
  raise "base must not be nil"
end
```

So, I will implement with the latter.

yykamei: Finally, I fixed with ab792d5.

lib/runners/workspace/http.rb
```diff
@@ -0,0 +1,64 @@
+module Runners
+  class Workspace::HTTP < Workspace
+    private
+
+    def prepare_base_source(dest)
+      base = archive_source.base
+      if base
+        provision(URI.parse(base), dest, archive_source.base_key)
+      else
+        raise "base must not be nil" unless base
```
ybiquitous: 😅 

```suggestion
        raise "base must not be nil"
```

PR #537, opened 2019-11-28
Add a feature to mask secure strings on TraceWriter

yykamei:
We are planning to fetch source code via Git. To get source code
from a private repository, we must set a token and perform command like
this:

```shell-session
$ git config http.extraheader "Authorization: Basic token"
```

I'm going to run the command with `Shell#capture3!`. However, this
command writes its passed command line via `TraceWriter`. So, we have to
mask the secure string and prevent it from being output in traces.

Secure strings also should be masked on stdout or stderr, so I
implemented the feature on `TraceWriter` instead of `Shell`.

### DIFF
--- a/goodcheck.yml
+++ b/goodcheck.yml
@@ -29,5 +29,13 @@ rules:
     pass: |
       class FooError < UserError; end
 
+  - id: sider.runners.trace_writer.masked_string
+    glob: lib/runners/trace_writer.rb
+    message: |
+      Use `#masked_string` if the passed string has possibly sensitive values
+    justification:
+      - When the passed value is not sensitive
+      - When your change is irrelevant to `:writer`
+
 import:
   - https://raw.githubusercontent.com/sider/goodcheck-rules/master/rules/typo.yml
--- a/lib/runners/trace_writer.rb
+++ b/lib/runners/trace_writer.rb
@@ -1,14 +1,14 @@
 module Runners
   class TraceWriter
-    # @dynamic writer
-    attr_reader :writer
+    attr_reader :writer, :sensitive_strings
 
-    def initialize(writer:)
+    def initialize(writer:, sensitive_strings: [])
       @writer = writer
+      @sensitive_strings = sensitive_strings
     end
 
     def command_line(command_line, recorded_at: Time.now)
-      self << { trace: 'command_line', command_line: command_line, recorded_at: recorded_at.utc.iso8601 }
+      self << { trace: 'command_line', command_line: command_line.map { |v| masked_string(v) }, recorded_at: recorded_at.utc.iso8601 }
     end
 
     def status(status, recorded_at: Time.now)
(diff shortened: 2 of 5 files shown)


lib/runners/trace_writer.rb
```diff
@@ -1,14 +1,15 @@
 module Runners
   class TraceWriter
     # @dynamic writer
```
ybiquitous: [nitpick] Let's remove the annotation!

lib/runners/trace_writer.rb
```diff
@@ -46,19 +47,19 @@ def message(message, recorded_at: Time.now, max_length: 4_000)
...
+      self << { trace: 'warning', file: file, message: masked_string(message), recorded_at: recorded_at.utc.iso8601 }
     end
 
     def ci_config(content, recorded_at: Time.now)
       self << { trace: 'ci_config', content: content, recorded_at: recorded_at.utc.iso8601 }
     end
 
     def error(message, recorded_at: Time.now, max_length: 4_000)
-      each_slice(message, size: max_length) do |text|
+      each_slice(masked_string(message), size: max_length) do |text|
```
ybiquitous: [question] How about masking within the `#<<(object)` method? If we will add a new method in the future, I'm concerned that we will forget masking... 🤔

yykamei: I intended to do that, but `#<<` expects any `object` values instead of `String`, so I gave up to do this. Do you have any ideas?

yykamei:
> [question] What values do you assume when setting `secure_strings` actually?

I want to hide the value of `GitSource#token`.

lib/runners/trace_writer.rb
ybiquitous: Umm..., the following seems possible but not good. Especially, `each_slice` is difficult.

```ruby
def <<(object)
  writer << Schema::Trace.anything.coerce(object).yield_self do |obj|
    obj.each do |key, value|
      obj[key] = masked_string(value) if value.is_a? String
    end
  end
end
```

OK, let's keep them.

Also, how about adding a Goodcheck rule to enforce using `masked_string`?

yykamei: I'll do it :+1:

yykamei:
I renamed the attribute to `sensitive_strings` because I want to hide "secret" or "credentials," and they are not secure.

goodcheck.yml
```diff
@@ -29,5 +29,13 @@ rules:
...
       class FooError < UserError; end
 
+  - id: sider.runners.trace_writer.masked_string
+    glob: lib/runners/trace_writer.rb
+    message: |
+      Use `#masked_string` if the passed string has possibly sensitive values
+    justification:
+      - When the passed value is not sensitive
+      - When your change is irrelevant to `:writer`
+
```
ybiquitous: 👏

lib/runners/trace_writer.rb
yykamei: It's working :tada: 
https://sider.review/gh/repos/202486306/pulls/537#issue-4881731

PR #548, opened 2019-12-02
Add Workspace::Git to fetch source code via Git

yykamei:
`Workspace::Git` provides the feature to allow us to fetch source code
via Git. Previously, Runners depended on the uploaded and encrypted
tarballs, so the caller of Runners should have prepared source code
before calling Runners. This made our system complicated and make its
maintainability difficult.

From this change, we can delegate the responsibility of fetching source
code to Runners.


yykamei:
:thinking: We should not call the actual git-fetch(1)?

https://app.circleci.com/jobs/github/sider/runners/51358

yykamei:
I got it. git(1) on CircleCI is so old that it doesn't support `--no-auto-gc` option.

ybiquitous:
Should we migrate to GitHub Actions?

ybiquitous:
Or need only `apt-get install git`?

ybiquitous:
Here seems the images repository of CircleCI:
https://github.com/circleci-public/circleci-dockerfiles

yykamei:
I think we can alternatively use `git config gc.auto 0`, and probably it'll work. The change is small like this:

```diff
diff --git a/lib/runners/workspace/git.rb b/lib/runners/workspace/git.rb
index 74d4076..838dc2b 100644
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -26,6 +26,7 @@ module Runners
       @git_directory ||= Pathname(Dir.mktmpdir).tap do |path|
         shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
         shell.capture3!("git init")
+        shell.capture3!("git config gc.auto 0")
         shell.capture3!("git remote add origin #{remote_url}")
         shell.capture3!("git", *git_fetch_args)
         at_exit { FileUtils.remove_entry(path) }
@@ -40,7 +41,7 @@ module Runners
     end

     def git_fetch_args
-      @git_fetch_args ||= %w[fetch --no-tags --no-auto-gc --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/*].tap do |command|
+      @git_fetch_args ||= %w[fetch --no-tags --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/*].tap do |command|
         pull_number = git_source.pull_number
         command << "+refs/pull/#{pull_number}/head:refs/remotes/pull/#{pull_number}/head" if pull_number
       end
```

It's easier to use git-config(1) than moving to GitHub Actions. What do you think?

ybiquitous:
Using `git config gc.auto 0` sounds good to me! Let's continue it on this PR.

About the outdated git(1) on Circle CI, it may cause another problem in the future. So ideally I think it better to update git(1) on CircleCI as possible if it would be an easy task.

But it does not look necessary to do it on this PR. 😄

lib/runners/workspace/git.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def git_directory
+      @git_directory ||= Pathname(Dir.mktmpdir).tap do |path|
+        shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
+        shell.capture3!("git init")
+        shell.capture3!("git config gc.auto 0")
+        shell.capture3!("git remote add origin #{remote_url}")
+        shell.capture3!("git", *git_fetch_args)
+        at_exit { FileUtils.remove_entry(path) }
```
yykamei: I want to cache the Git repository during the analysis, so I used `Dir.mktmpdir` without a block and implemented cleaning up the procedure manually.

.circleci/config.yml
```diff
@@ -50,6 +50,9 @@ jobs:
       - checkout
       - install_gems
       - install_npm_and_yarn
+      - run:
+          name: Upgrade git(1)
+          command: sudo apt-get update && sudo apt-get upgrade git
```
ybiquitous: [question] After all, this is necessary?

image

https://circleci.com/gh/sider/runners/51486

yykamei: No, but I agree with your comment, so I added it to the "Steps."

>  So ideally I think it better to update git(1) on CircleCI as possible if it would be an easy task.

ybiquitous: What do you mean on the following comment?

> git(1) on CircleCI is so old that it doesn't support `--no-auto-gc` option.

https://github.com/sider/runners/pull/548#issuecomment-560229658

yykamei: git(1) on CircleCI is "2.20.1" and it looks lack of the "--[no-]auto-gc" option, so I changed this PR not to use `git fetch --no-auto-gc`. Consequently, running `apt-get upgrade git` became unnecessary. Without git-config(1), we will still get hit an error on CirlceCI.

lib/runners/workspace/git.rb
ybiquitous: I feel a bit worried about adding a global event handler like `at_exit` in the program's deeper location. 😓 

So, how about adding a root temp directory for `Workspace`, for example?

```ruby
class Workspace
  def root_tmp_dir
    @root_tmp_dir ||= Dir.mktmpdir
  end

  def open
    # ...
  ensure
    FileUtils.remove_entry root_tmp_dir
  end
end
```

.circleci/config.yml
ybiquitous: I got it. If `apt-get upgrade git` does nothing actually, I think better to remove this `run`. It looks confusing to me.
I thought `apt-get upgrade git` could resolve the lack of `--auto-gc` option. 😅

lib/runners/workspace/git.rb
yykamei: Where do I call `Workspace#open`?

ybiquitous: I think `Harness#run` is good as until now.

yykamei: OK. I'll do it.

By the way, I'm not worried about using `at_exit` in the short-running and also not library process like Runners.  I agree that adding a global state might cause trouble, but I believe we can handle it when we use the method appropriately within our application program only.

ybiquitous: Yes, this is just my opinion. I actually don't verify if it causes an issue.

.circleci/config.yml
yykamei: Thank you. I [Reverted it](51176a2).

lib/runners/workspace/git.rb
```diff
@@ -0,0 +1,49 @@
...
+      FileUtils.copy_entry(git_directory, dest)
+      FileUtils.remove_entry(dest / ".git")
+    end
+
+    def git_directory
+      @git_directory ||= root_tmp_dir.tap do |path|
+        shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
+        shell.capture3!("git init")
+        shell.capture3!("git config gc.auto 0")
+        shell.capture3!("git remote add origin #{remote_url}")
```
ybiquitous: [question] Is there no problems when executing in shell mode?

```suggestion
        shell.capture3!("git", "remote", "add", "origin", remote_url)
```

lib/runners/schema/options.rb
```diff
@@ -4,7 +4,7 @@ module Schema
       # @type self: Types::Options
 
       let :source, enum(
-        object(head: string, base: string?, git_http_url: string, owner: string, repo: string, token: string?, pull_number: number?),
+        object(head: string, base: string?, git_http_url: string, owner: string, repo: string, http_userinfo: string?, pull_number: number?),
```
ybiquitous: [IMO] It looks a bit inconsistent for naming and a bit confusing to me.

- `git_http_url` - with `git`
- `http_userinfo` - without `git`

lib/runners/workspace/git.rb
yykamei: I don't know, but I will adjust to other implementations.

lib/runners/schema/options.rb
yykamei: Any ideas?

yykamei: Done with e10ef81

lib/runners/workspace/git.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+    end
+
+    def prepare_head_source(dest)
+      provision(git_source.head, dest)
+    end
+
+    def provision(commit_hash, dest)
+      shell = Shell.new(current_dir: git_directory, trace_writer: trace_writer, env_hash: {})
+      shell.capture3!("git checkout #{commit_hash}")
```
ybiquitous: Is this also forgotten?

```suggestion
      shell.capture3!("git", "checkout", commit_hash)
```

sig/polyfill.rbi
```diff
@@ -77,17 +77,21 @@ class FileUtils
...
+                   | () -> String
 end
 
 class URI
   def self.parse: (String) -> URI
+  def self.join: (*String) -> URI
   def path: -> String
   def scheme: -> String
   def host: -> String
+  def userinfo=: (String) -> String
```
ybiquitous: [question] Does this setter return `String`? Is it `void` appropriate?

```suggestion
  def userinfo=: (String) -> void
```

test/workspace/git_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  def test_remote_url
+    with_workspace(head: "commit",
+                   git_http_url: "https://github.example.com", owner: "foo", repo: "bar") do |workspace|
+      assert_instance_of Runners::Workspace::Git, workspace
+      mktmpdir do |dest|
+        assert_equal URI("https://github.example.com/foo/bar"), workspace.send(:remote_url)
+      end
```
ybiquitous: [question] The `dest` argument looks unused. Is the `mktmpdir` block needed?

test/workspace/git_test.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def test_remote_url_with_token
+    with_workspace(head: "commit",
+                   git_http_url: "https://github.com", owner: "foo", repo: "bar",
+                   git_http_userinfo: "x-access-token:v1.aaabbbcccdddeeefffgggbc06400127f248a82ac") do |workspace|
+      assert_instance_of Runners::Workspace::Git, workspace
+      mktmpdir do |dest|
+        assert_equal URI("https://x-access-token:v1.aaabbbcccdddeeefffgggbc06400127f248a82ac@github.com/foo/bar"), workspace.send(:remote_url)
+      end
```
ybiquitous: Also here.

lib/runners/workspace/git.rb
yykamei: Thank you :+1:

sig/polyfill.rbi
yykamei: I followed the actual behavior, but yes, I agree that `#userinfo=` should return nothing.

```
irb(main):002:0> uri = URI('https://example.com')
=> #<URI::HTTPS https://example.com>
irb(main):003:0> uri.userinfo = 'username'
=> "username"
```

test/workspace/git_test.rb
yykamei: You're right. I'm going to remove it.

PR #554, opened 2019-12-04
Add cpplint to `rake bump:analyzers`

ybiquitous:
Also, this changes the cron schedule on GitHub Actions to verify the task's behavior.

### DIFF
--- a/.github/workflows/bump_analyzers.yml
+++ b/.github/workflows/bump_analyzers.yml
@@ -2,7 +2,7 @@ name: Bump analyzers
 
 on:
   schedule:
-    - cron: "0 0 1 * *"
+    - cron: "0 0 1,5,15,25 * *"
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


lib/tasks/bump/analyzers.rb
```diff
@@ -46,14 +46,17 @@ def initialize(response)
     end
   end
 
+  ANALYZERS_UNSUPPORTED_BY_DEPENDABOT = {
```
ybiquitous: [note] `ANALYZERS` conflicts with `Rakefile`:

https://github.com/sider/runners/blob/8e54d2e52b6fb34df8ca96e68cfe760fd9112589/Rakefile#L21

.github/workflows/bump_analyzers.yml
```diff
@@ -2,7 +2,7 @@ name: Bump analyzers
 
 on:
   schedule:
-    - cron: "0 0 1 * *"
+    - cron: "0 0 1,5,15,25 * *"
```
ybiquitous: image

PR #566, opened 2019-12-09
[Reek] Improve issue message

ybiquitous:
- More simple (trimming a rule ID).
- Wrap `context` with back-quotes (it is almost a class or method identifier).


ybiquitous:
Thanks!

PR #569, opened 2019-12-09
[Chore] Pass multiple arguments to `sh` method in Rake tasks

ybiquitous:


querly.yml
```diff
@@ -20,3 +20,12 @@ rules:
     examples:
       - before: trace_writer.status(status)
       - before: "trace_writer.status(status, recorded_at: Time.now)"
+  - id: sider.runners.sh-util-in-rake-tasks
+    pattern:
+      - sh(_)
+    message: |
+      Consider passing multiple arguments instead of a single argument to `sh` in Rake tasks.
+
+      When passing a single argument, it runs in shell-mode.
```
yykamei: :thinking: The current Ruby implementation looks to see whether the argument contains Sell words or not. So, passing a single argument does not always invoke sh(1).

https://github.com/ruby/ruby/blob/7a9b2039b7b82f081f2cab40a4fbbc8b01231ca4/process.c#L2895

> In case of including meta-characters or any shell specific words,
> passing multiple arguments is recommended.

By the way, I like writing with an only single argument for `#sh` method in `Rakefile`, but to reduce human error, to compel to pass single arguments seems better.

ybiquitous: Yes, I think that this implementation rule may be hard to understand for many people including me (which is shell-mode?). 😅 

For example, my Docker password includes the meta-characters, so the `sh "docker login -u #{DOCKER_USER} -p #{DOCKER_PASSWORD}` call failed and I didn't understand why the command failed for a while. 😢 

I think the shell-mode is unnecessary in most cases, so I add the rule to reduce errors and confusion. I welcome to revet the rule if it is not so useful. 

See also <https://ruby-doc.org/core-2.6.5/Kernel.html#method-i-exec>.

yykamei: Sorry, I mistyped

Before:

>  to compel to pass single arguments seems 

After:

> to compel to pass multiple arguments seems 

In other words, I agree with you :smile: but how about including the description like this?

> In case of including meta-characters or any shell specific words,
> passing multiple arguments is recommended.

ybiquitous: Make sense. I will improve the message. Thank you!

ybiquitous: Thanks for your feedback! I rewrite the rule (squashing to one commit).

PR #575, opened 2019-12-10
Show runtime version as possible

ybiquitous:


lib/runners/processor/scss_lint.rb
```diff
@@ -26,6 +28,10 @@ def self.ci_config_section_name
       'scss_lint'
     end
 
+    def analyzer_version
+      @analyzer_version ||= extract_version! analyzer_bin
+    end
```
ybiquitous: [note] `Ruby` module overrides `#analyzer_version`, so this override is needed.

PR #577, opened 2019-12-11
[PHPMD] Output report to file instead of STDOUT

ybiquitous:
Why? - Because STDOUT may include any non-XML string. This uses a more reliable way.

### DIFF
--- a/lib/runners/processor/phpmd.rb
+++ b/lib/runners/processor/phpmd.rb
@@ -123,13 +123,26 @@ module Runners
     end
 
     def run_analyzer(changes, targets, rule, options)
+      report_file = working_dir / "phpmd-report-#{Time.now.to_i}.xml"
+
       # PHPMD exits 1 when some violations are found.
       # The `--ignore-violation-on-exit` will exit with a zero code, even if any violations are found.
-      # See. https://phpmd.org/documentation/index.html
-      commandline = [analyzer_bin, targets, 'xml', rule, '--ignore-violations-on-exit'] + options
-      stdout, _ = capture3!(*commandline)
+      # See https://phpmd.org/documentation/index.html
+      command_line = [
+        analyzer_bin, targets, "xml", rule, "--ignore-violations-on-exit",
+        "--report-file", report_file.to_s, *options,
+      ]
+      capture3!(*command_line)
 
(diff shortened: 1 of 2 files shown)


lib/runners/processor/phpmd.rb
```diff
@@ -123,13 +123,26 @@ def strict(config)
     end
 
     def run_analyzer(changes, targets, rule, options)
+      report_file = Pathname(Dir.tmpdir) / "phpmd-report.xml"
```
yykamei: How about using `working_dir`?  `working_dir` will be deleted after the analysis has finished.

ybiquitous: Nice! I used the following code as a reference. Is `working_dir` different from a project root dir?

https://github.com/sider/runners/blob/9ed55326e3ebc3ee8491e512a7cbdad7c48c9c18/lib/runners/processor/eslint.rb#L213

yykamei: Yes, I think so. For example, if `root_dir` is `frontend`, the analysis will be done in `frontend`, but the output file will not be created in `frontend` like this:

```
.
├── README.md
├── app/
├── config/
├── frontend/    <-- root_dir
├── (output.json) <-- `working_dir / "output.json"`
└── goodcheck.yml
```

ybiquitous: Thanks. I'm concerning about give an effect on a user's project tree, but is this my concern too much?

ybiquitous: So, I will add a random string to an output filename like this.

```ruby
report_file = working_dir / "#{Time.now.to_i}-phpmd-report.xml"
```

Is this OK?

yykamei: I think there are no concerns in the current implementation. However, this concern makes sense when we directly see the entries of the source code and do something with them. Fortunately, we have `Changes` class.

ybiquitous: Thank you. I've pushed and squashed the fixation:

```ruby
report_file = working_dir / "phpmd-report-#{Time.now.to_i}.xml"
```

https://github.com/sider/runners/pull/577/files#diff-45bfa428c2ae719ccc23a7f61b3f794fR126

PR #584, opened 2019-12-12
Filter issues with the "diff lines"

### LINKED ISSUE #574: Filter issues with the "diff lines"
The current implementation filters the issues only with the file paths like this:

https://github.com/sider/runners/blob/5d6ced4f784971ea45db06903d0f7e66f87b6576/lib/runners/harness.rb#L52

Now, I want to filter them with the "diff lines." For example, I want to exclude the line which does not exist in the diff from the original issues.

```diff
diff --git a/a.py b/a.py
index 7e24076..5724d52 100644
--- a/a.py
+++ b/a.py
@@ -17,4 +17,10 @@ def foo():
     some bad code  # <--- Issue1: I want to exclude this line


+    pass
+
+
+def bad_method(): # <--- Issue2
+    pass
+def bad_method2(): # <--- Issue3
     pass
```

yykamei:
Resolve #574

Using `Workspace::Git`, we can detect the issues shows up in the change or not.
So, I updated the filtering process with `GitDiffParser`.

### DIFF
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -3,11 +3,13 @@ module Runners
     attr_reader :changed_paths
     attr_reader :unchanged_paths
     attr_reader :untracked_paths
+    attr_reader :patches
 
-    def initialize(changed_paths:, unchanged_paths:, untracked_paths:)
+    def initialize(changed_paths:, unchanged_paths:, untracked_paths:, patches:)
       @changed_paths = changed_paths
       @unchanged_paths = unchanged_paths
       @untracked_paths = untracked_paths
+      @patches = patches
     end
 
     def delete_unchanged(dir:, except: [], only: [])
@@ -29,7 +31,23 @@ module Runners
       FileUtils.rm(files_to_delete.map(&:to_s))
     end
 
-    def self.calculate(base_dir:, head_dir:, working_dir:)
+    def include?(issue)
+      gdp = patches # NOTE: This assignment is required for typecheck by Steep
+      if gdp
+        location = issue.location
+        return true unless location
+        patch = gdp.find_patch_by_file(issue.path.to_s)
+        if patch && location
+          patch.changed_lines.one? { |line| location.start_line == line.number }
+        else
+          false
+        end
+      else
+        Set.new(changed_paths).member?(issue.path)
+      end
+    end
+
+    def self.calculate(base_dir:, head_dir:, working_dir:, patches:)
       changed_paths = []
       unchanged_paths = []
       untracked_paths = []
@@ -65,9 +83,10 @@ module Runners
         end
       end
 
-      new(changed_paths: changed_paths.sort,
+      new(changed_paths: changed_paths.sort!,
           unchanged_paths: unchanged_paths.sort!,
-          untracked_paths: untracked_paths.sort!)
+          untracked_paths: untracked_paths.sort!,
+          patches: patches)
     end
   end
 end
--- a/lib/runners/workspace.rb
+++ b/lib/runners/workspace.rb
@@ -52,7 +52,7 @@ module Runners
             end
 
             changes = trace_writer.message "Calculating changes between head and base..." do
-              Changes.calculate(base_dir: base_path, head_dir: head_path, working_dir: working_dir)
+              Changes.calculate(base_dir: base_path, head_dir: head_path, working_dir: working_dir, patches: patches)
             end
 
             yield git_ssh_path, changes
@@ -171,6 +171,10 @@ module Runners
         out, status = Open3.capture2e("tar", "xf", archive.to_s, chdir: dir)
         raise "Extracting archive failed - #{out}" unless status.success?
       end
+    end
+
+    def patches
+      nil
     end
   end
 end
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -45,5 +45,19 @@ module Runners
         command << "+refs/pull/#{pull_number}/head:refs/remotes/pull/#{pull_number}/head" if pull_number
       end
     end
+
+    def patches
+      base = git_source.base
+      head = git_source.head
+      if base && head
+        git_directory.yield_self do |path|
+          shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
+          stdout, _ = shell.capture3!("git", "diff", base, head)
+          GitDiffParser.parse(stdout)
+        end
+      else
+        nil
+      end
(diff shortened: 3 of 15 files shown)


lib/runners/changes.rb
```diff
@@ -29,7 +31,26 @@ def delete_unchanged(dir:, except: [], only: [])
...
 
-    def self.calculate(base_dir:, head_dir:, working_dir:)
+    def include?(issue)
+      gdp = patches
+      location = issue.location
+      if gdp
+        patch = gdp.find_patch_by_file(issue.path.to_s)
+        case
+        when patch && location
+          patch.changed_lines.one? { |line| location.start_line == line.number }
```
yykamei: :thinking: I don't know why `patch` is assumed to be an optional value. In this context, `patch` and `location` are not `nil`.

https://app.circleci.com/jobs/github/sider/runners/55575
```
lib/runners/changes.rb:41:10: NoMethodError: type=(::GitDiffParser::Patch | nil), method=changed_lines (patch.changed_lines)
```

ybiquitous: Need to use `@type`?

ybiquitous: Or, need `if location` block?

yykamei: Oh, using `if` block instead of `case`, the `typecheck` succeeded. It's strange to me, but I will change the code using that.

```diff
diff --git a/lib/runners/changes.rb b/lib/runners/changes.rb
index d7668a7..a08964c 100644
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -36,11 +36,9 @@ module Runners
       location = issue.location
       if gdp
         patch = gdp.find_patch_by_file(issue.path.to_s)
-        case
-        when patch && location
+        if patch && location
           patch.changed_lines.one? { |line| location.start_line == line.number }
-        when patch && !location
-          # NOTE: issues without location should be included
+        elsif patch && !location
           true
         else
           false
```

ybiquitous: i don't know why but looks good!

btw, `gdp.find_patch_by_file` looks useless if location is nil.

yykamei: I see. Fixed with 26f55da

lib/runners/changes.rb
```diff
@@ -29,7 +31,23 @@ def delete_unchanged(dir:, except: [], only: [])
       FileUtils.rm(files_to_delete.map(&:to_s))
     end
 
-    def self.calculate(base_dir:, head_dir:, working_dir:)
+    def include?(issue)
+      gdp = patches
```
ybiquitous: `gdp` is a bit confusing to me. Is this assignment needed?

lib/runners/changes.rb
```diff
@@ -29,7 +31,23 @@ def delete_unchanged(dir:, except: [], only: [])
...
+      location = issue.location
```
ybiquitous: Can we move this line into `if gdp` block?

lib/runners/changes.rb
```diff
@@ -29,7 +31,23 @@ def delete_unchanged(dir:, except: [], only: [])
...
+      if gdp
+        return true unless location
+        patch = gdp.find_patch_by_file(issue.path.to_s)
+        if patch && location
```
ybiquitous: Is this `&& location` needed?

lib/runners/workspace.rb
```diff
@@ -52,7 +52,7 @@ def open
             end
 
             changes = trace_writer.message "Calculating changes between head and base..." do
-              Changes.calculate(base_dir: base_path, head_dir: head_path, working_dir: working_dir)
+              Changes.calculate(base_dir: base_path, head_dir: head_path, working_dir: working_dir, patches: patches)
```
ybiquitous: I feel more simple to set `patches: nil`.

lib/runners/changes.rb
yykamei: Yes..., without assigning, I've got the following message:

```
lib/runners/changes.rb:38:16: NoMethodError: type=(::GitDiffParser::Patches | nil), method=find_patch_by_file (patches.find_patch_by_file(issue.path.to_s))
rake aborted!
```

lib/runners/changes.rb
yykamei: Yes, I will do that.

lib/runners/changes.rb
yykamei: Unfortunately, yes...

lib/runners/workspace.rb
yykamei: If the instance is the one of `Workspace::Git#open`, I want to set the object of `GitDiffParser::Patches` to `:patches`, so `patches` method seems necessary.

lib/runners/changes.rb
ybiquitous: Umm, I don't know why. 😢 

Here is a solution by adding the following helper method (but quite ugly):

```ruby
def patches!
  patches or raise "Must not be nil!"
end

def include?(issue)
  if patches
    patches!.find_patch_by
  end
end
```

lib/runners/workspace.rb
ybiquitous: OK, I see. 👍

lib/runners/workspace/git.rb
```diff
@@ -45,5 +45,19 @@ def git_fetch_args
...
       end
     end
+
+    def patches
+      base = git_source.base
+      head = git_source.head
+      if base && head
+        git_directory.yield_self do |path|
+          shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
+          stdout, _ = shell.capture3!("git", "diff", base, head)
```
ybiquitous: Do we need to consider `ignore-space-*` options?

yykamei: Currently, Sider seems not to use this option.

I think it's better not to include the option since some tools issue about white spaces.

lib/runners/changes.rb
yykamei: :thinking: Hmmm, I would prefer assigning...

lib/runners/workspace/git.rb
ybiquitous: Make sense.

lib/runners/changes.rb
ybiquitous: Then, I need comments to justify the Steep hacks like the useless assignment.

yykamei: Thank you. I fixed it with 1323cb4

PR #587, opened 2019-12-12
[HAML-Lint] Add warnings if RuboCop fails

ybiquitous:
## Why?

HAML-Lint exists successfully even if RuboCop fails, so users may notice such errors. By adding warnings, this aims to make users noticing the errors.

See https://github.com/sds/haml-lint/issues/317

### DIFF
--- a/lib/runners/processor/haml_lint.rb
+++ b/lib/runners/processor/haml_lint.rb
@@ -218,11 +218,8 @@ module Runners
     #
     # @see https://github.com/sds/haml-lint/issues/317
     def add_rubocop_warnings_if_exists(stderr)
-      stderr.each_line(chomp: true) do |line|
-        line.match(/\b(cannot load such file -- [\w-]+)\b/) do |match|
-          error_message, = match.captures
-          add_warning error_message
-        end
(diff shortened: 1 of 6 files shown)


lib/runners/processor/haml_lint.rb
```diff
@@ -205,9 +207,23 @@ def run_analyzer(targets, options)
...
       Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
         parse_result(stdout).each { |v| result.add_issue(v) }
       end
     end
+
+    # NOTE: HAML-Lint exits successfully even if RuboCop fails.
+    #
+    # @see https://github.com/sds/haml-lint/issues/317
+    def add_rubocop_warnings_if_exists(stderr)
+      stderr.each_line(chomp: true) do |line|
```
yykamei: [nitpick] I think `stderr.scan(/\b(cannot load such file -- [\w-]+)\b/)` would work, too.

ybiquitous: Thanks! Very nice idea! I've fixed via 98532159b060b482761a767946269ab5a844eda6

PR #593, opened 2019-12-13
Use Yarn when duplicate lock files are found

ybiquitous:
When duplicate lock files `package-lock.json` and `yarn.lock` are found,

- Before: Analysis failed.
- After: Show a warning and use Yarn.

Why?
----

In a few projects, there is a case that the duplicate lock files exist at the same time.
(for example, during migration from npm to Yarn)

In such case, failing analysis can be too strict.


lib/runners/nodejs.rb
```diff
@@ -205,8 +196,10 @@ def yarn_install(option)
     end
 
     def check_installed_nodejs_deps(constraints, default_dependency)
-      # NOTE: `npm ls` fails when any peer dependencies are missing. Also, the command output can be too long.
-      stdout, _, _ = capture3 "npm", "ls", "--depth=0", "--json", trace_stdout: false
+      # NOTE: `npm ls` fails when any peer dependencies are missing.
+      #        Also, this scans `node_modules/` without `package-lock.json`.
+      #        Also, the command output can be too long.
+      stdout, _, _ = capture3 "npm", "ls", "--depth=0", "--json", "--package-lock=false", trace_stdout: false
```
ybiquitous: [note] Added `--package-lock=false` option. Because `npm ls` scans also `package-lock.json` unless the option.

When the installation is executed by Yarn, `npm ls` scanning `package-lock.json` can output the following error:

```
npm ERR! missing: eslint@6.0.1
```

test/smokes/eslint/expectations.rb
```diff
@@ -673,3 +673,13 @@
 }, {
   warnings: [{ message: /The required dependency `eslint` may not have been correctly installed/, file: "package.json" }],
 })
+
+Smoke.add_test("duplicate_lock_files", {
+  guid: "test-guid",
+  timestamp: :_,
+  type: "success",
+  issues: [],
+  analyzer: { name: "ESLint", version: "5.16.0" },
```
ybiquitous: [note] `eslint@5.1.0` is defined in `package-lock.json`:

https://github.com/sider/runners/pull/593/files#diff-9b6c762daf6969b97723fe31938fc4a3R256

PR #615, opened 2019-12-19
Use `bundle config without` instead of `--without` CLI option

ybiquitous:
The CLI option is deprecated. See below:

https://github.com/bundler/bundler/blob/v2.1.1/UPGRADING.md#cli-deprecations

Here is the `docker build` log:

```
Step 4/16 : RUN cd ${RUNNERS_DIR} &&     bundle config --global jobs 4 &&     bundle config --global retry 3 &&     bundle config --local without development:test &&     bundle config &&     bundle install
 ---> Running in 466cf01a4480
You are replacing the current local value of without, which is currently nil
Settings are listed in order of priority. The top value will be used.
jobs
Set for the current user (/home/analyzer_runner/.bundle/config): 4

retry
Set for the current user (/home/analyzer_runner/.bundle/config): 3

without
Set for your local app (/home/analyzer_runner/runners/.bundle/config): [:development, :test]
```


ybiquitous:
Thanks!

PR #617, opened 2019-12-19
Set default `TAG`

ybiquitous:
Why?

- Reduce a project-setup work.
- Even if the `dev` tag is pushed to Docker Hub, it does not affect our service (just removing it).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -101,14 +101,7 @@ namespace :docker do
   end
 
   def tag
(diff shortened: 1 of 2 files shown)


ybiquitous:
@yykamei What do you think? Is there any influences which I miss?

ybiquitous:
Thanks for your comment. 😄

PR #623, opened 2019-12-24
Fix Bundler warning

ybiquitous:
> [DEPRECATED] `Bundler.with_clean_env` has been deprecated in favor of `Bundler.with_unbundled_env`. If you instead want the environment before bundler was originally loaded, use `Bundler.with_original_env`

See also:
- https://github.com/bundler/bundler/blob/v2.1.2/lib/bundler.rb#L376
- https://github.com/bundler/bundler/blob/v2.1.2/lib/bundler.rb#L381

### DIFF
--- a/lib/runners/ruby/gem_installer.rb
+++ b/lib/runners/ruby/gem_installer.rb
@@ -29,7 +29,7 @@ module Runners
 
         trace_writer.header "Installing gems..."
 
-        Bundler.with_clean_env do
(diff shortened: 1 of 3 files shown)


ybiquitous:
Oh, we need to continue using also Bundler v1 for a while... 😰 

https://github.com/sider/devon_rex/blob/796b99356412224eed23cf56c5fa0b11bffd78f5/base/bin/build-ruby#L40

Issue #581, opened 2019-12-11
Support hadolint


sumyapp:
https://github.com/hadolint/hadolint

Looks like https://github.com/sider/runners/pull/417 for Docker(hadolint)
That is cpplint support pull request. In that time, we published help page https://help.sider.review/tools/cplusplus/cpplint

```
$ hadolint -h
hadolint - Dockerfile Linter written in Haskell

Usage: hadolint [-v|--version] [-c|--config FILENAME] [-f|--format ARG]
                [DOCKERFILE...] [--ignore RULECODE]
                [--trusted-registry REGISTRY (e.g. docker.io)]
  Lint Dockerfile for errors and best practices

Available options:
  -h,--help                Show this help text
  -v,--version             Show version
  -c,--config FILENAME     Path to the configuration file
  -f,--format ARG          The output format for the results [tty | json |
                           checkstyle | codeclimate | codacy] (default: tty)
  --ignore RULECODE        A rule to ignore. If present, the ignore list in the
                           config file is ignored
  --trusted-registry REGISTRY (e.g. docker.io)
                           A docker registry to allow to appear in FROM
                           instructions

```

## TODO
- processor
- Dockerfile
- Smoke test
- (optional) Write document ref: https://github.com/sider/sider-docs

ybiquitous:
via #625

PR #625, opened 2019-12-24
add hadolint

mpls104:
## Issue
https://github.com/sider/runners/issues/581

## Remains todo
- [x] Write smoke test(with some Option)
- [x] Add pattern for success end, waring exit, error exit
- [x] Add config option
- [x] Add test for error exit
- [x] Write document
- [x] Add CI setting
- [ ] Feedback about hadolint
- [ ] Fix developer guid if needed

## Things want to confirm
- [x] Do we need to use config file `.hadolint.yaml`?
- [ ] Do we need -trusted-registry option?
- [x] Which version should be extracted from `Haskell Dockerfile Linter 3fab409` ?
- How should be object, schema, link in Issue?
- [x] Should I clean up commit logs? -> Use squash merge


images/hadolint/Dockerfile.erb
```diff
@@ -0,0 +1,20 @@
+FROM sider/devon_rex_haskell:2.4.0
+
+ARG HADOLINT_VERSION="v1.17.30"
+
+USER root
+RUN git clone https://github.com/hadolint/hadolint tags/${HADOLINT_VERSION} --depth 1 && \
```
yykamei: This doesn't seem to lock the version of "hadolint." I think `--branch` option is required. See our SwiftLint runners implementation [here](https://github.com/sider/runners/blob/ef34276cdc76ce0791f673f59ace8bec5242512b/images/swiftlint/Dockerfile.erb#L8) :smiley:

images/hadolint/Dockerfile.erb
```diff
@@ -0,0 +1,20 @@
+FROM sider/devon_rex_haskell:2.4.0
+
+ARG HADOLINT_VERSION="v1.17.30"
+
+USER root
```
yykamei: :thinking: I couldn't figure out why changing the user here. The following `RUN` instruction doesn't seem to require the permission of superuser.

images/hadolint/Dockerfile.erb
ybiquitous: Can we use precompiled binaries on the [release](https://github.com/hadolint/hadolint/releases/tag/v1.17.3) page?
If we can, the Docker build time can be reduced.

ybiquitous:
Please add hadolint to README. 🙏 
https://github.com/sider/runners/blob/master/README.md#supported-analyzers

images/hadolint/Dockerfile.erb
yykamei: I think we can :+1:

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,93 @@
+module Runners
+  class Processor::Hadolint < Processor
+    Schema = StrongJSON.new do
+      let :runner_config, Schema::RunnerConfig.base.update_fields { |fields|
```
ybiquitous: > __[RuboCop]__ Parenthesize the param `Schema::RunnerConfig.base.update_fields { |fields| ([see on Sider](https://sider.review/gh/repos/202486306/pulls/625#issue-4989187))
>
> <details>
>
>         fields.merge!({
>                         target: enum?(string, array(string)),
>                         ignore: boolean?,
>                       })
>       }` to make sure that the block will be associated with the `Schema::RunnerConfig.base.update_fields` method call.
>
> | Rule | Severity |
> | - | - |
> | Lint/AmbiguousBlockAssociation | warning |
>
> References:
> - <https://rubystyle.guide#syntax>
> - <https://github.com/rubocop-hq/rubocop/blob/v0.78.0/manual/cops_lint.md#lintambiguousblockassociation>
>
> </details>

Near to a stylistic issue. I think we do not need to fix.

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+        Array(config[:target] || DEFAULT_TARGET).each do |target|
+          opts << target
+        end
+      end
+    end
+
+    def run_analyzer
+      stdout, stderr, status = capture3 analyzer_bin, *analyzer_options
+      if status.exitstatus == 1 && stdout.strip == "Please provide a Dockerfile"
```
ybiquitous: > __[Querly]__ Process#exitstatus may be nil ([see on Sider](https://sider.review/gh/repos/202486306/pulls/625#issue-4989186))
>
> <details>
>
>
>
> </details>

I think no problem as-is.

README.md
```diff
@@ -23,6 +23,7 @@ Each analyzer can be available as a Docker image.
 - [Golint](https://hub.docker.com/r/sider/runner_golint)
 - [Go Meta Linter](https://hub.docker.com/r/sider/runner_gometalinter)
 - [Goodcheck](https://hub.docker.com/r/sider/runner_goodcheck)
+- [Hadolint](https://hub.docker.com/r/sider/runner_hadolint)
```
ybiquitous: I think the official name is **hadolint**, not *Hadolint*.
Because the Wiki says "hadolint": https://github.com/hadolint/hadolint/wiki

ybiquitous:
```suggestion
- [hadolint](https://hub.docker.com/r/sider/runner_hadolint)
```

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    def analyzer_options
+      [].tap do |opts|
+        opts << "--format=json"
+        opts << "--trusted-registry=#{config[:trusted_registry]}" if config[:trusted_registry]
+
+        Array(config[:ignore]).each do |ignore|
+          opts << "--ignore=#{ignore}"
+        end
+        Array(config[:target] || DEFAULT_TARGET).each do |target|
```
ybiquitous: [suggestion] How about supporting **glob** format for `target` opion like ShellCheck?

https://github.com/sider/runners/blob/29303582982bfa8b961d14128c6627782ae93751/lib/runners/processor/shellcheck.rb#L91-L93

For example, `DEFAULT_TARGET = "**/Dockerfile{,.*}"`.

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,93 @@
...
+
+        line = file[:line]
+        id = file[:code]
+
+        Issue.new(
+          path: path,
+          location: Location.new(start_line: line),
+          id: id,
+          message: file[:message],
+        #todo object? schema?
```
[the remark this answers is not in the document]
ybiquitous: And, please add `links` like ShellCheck:

https://github.com/sider/runners/blob/29303582982bfa8b961d14128c6627782ae93751/lib/runners/processor/shellcheck.rb#L155

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,93 @@
...
+    private
+
+    def config
+      @config or raise "Must be initialized!"
+    end
+
+    def analyzer_options
+      [].tap do |opts|
+        opts << "--format=json"
+        opts << "--trusted-registry=#{config[:trusted_registry]}" if config[:trusted_registry]
```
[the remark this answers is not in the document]
mpls104: Yes. It supports multiple values so it has to be string or array of string.

https://github.com/hadolint/hadolint#configure
https://github.com/hadolint/hadolint/wiki/DL3026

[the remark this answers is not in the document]
mpls104: it's confusional that ruby cannot user `-` as symbol so need to user `trusted_registry` as key.
However, hadolint will accept `--trusted-registry`

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,93 @@
...
+    def run_analyzer
+      stdout, stderr, status = capture3 analyzer_bin, *analyzer_options
+      if status.exitstatus == 1 && stdout.strip == "Please provide a Dockerfile"
+        add_warning "No Docker file provided"
+        return Results::Success.new(guid: guid, analyzer: analyzer)
+      end
+
+      unless stderr.empty?
+        return Results::Failure.new(guid: guid, analyzer: analyzer, message: stderr)
+      end
```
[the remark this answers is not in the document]
mpls104: As far as I investigated, The relationship with stdout, stderr and exit status is as below.

Dockerfile not found
```
=> ["", "hadolint: test/smokes/hadolint/default/Dockerfil: openBinaryFile: does not exist (No such file or directory)\n", #<Process::Status: pid 24460 exit 1>]
```

Wrong Option
```
=> ["", "Invalid option `--ignor2'\n\nUsage: hadolint [-v|--version] [-c|--config FILENAME] [-f|--format ARG]\n                [DOCKERFILE...] [--ignore RULECODE]\n                [--trusted-registry REGISTRY (e.g. docker.io)]\n  Lint Dockerfile for errors and best practices\n", #<Process::Status: pid 24484 exit 1>]
```

There are point out from hadolint
```
=> ["[{\"line\":3,\"code\":\"DL3002\",\"message\":\"Last USER should not be root\",\"column\":1,\"file\":\"test/smokes/hadolint/default/Dockerfile\",\"level\":\"warning\"}]\n", "", #<Process::Status: pid 24478 exit 1>]
```

No error detected
```
=> ["[]\n", "", #<Process::Status: pid 24475 exit 0>]
```

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,93 @@
+module Runners
+  class Processor::Hadolint < Processor
+    Schema = StrongJSON.new do
+      let :runner_config, Schema::RunnerConfig.base.update_fields { |fields|
+        fields.merge!({
+                        target: enum?(string, array(string)),
+                        ignore: enum?(string, array(string)),
+                        trusted_registry: enum?(string, array(string)),
```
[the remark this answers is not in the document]
mpls104: I think so, I will add!

images/hadolint/Dockerfile.erb
ybiquitous: @yykamei Isn't that because only the root user has write permission to `/usr/local/bin`?

lib/runners/processor/hadolint.rb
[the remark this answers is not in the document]
ybiquitous: Make sense. 👍 
Let's change to `trusted-registry` like ESLint!
https://github.com/sider/runners/blob/0bd76fd06df3ab216d4744fb6c3bc59692f16ad5/lib/runners/processor/eslint.rb#L11

images/hadolint/Dockerfile.erb
yykamei: That's right if the directory `tags/${HADOLINT_VERSION}` is located in `/usr/local/bin` (And, I know the `git clone URL DIR` command is ran in `/` because `WORKDIR` is not set in the previous instructions). However, this `RUN` instruction doesn't need to modify the directory which requires the `root` permission. I think it's not required to `USER` instruction if we run git-clone(1) in `/tmp` directory.

lib/runners/processor/hadolint.rb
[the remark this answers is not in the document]
ybiquitous: Umm..., the hadolint behavior seems not enough for us.

First, how about just rescuing JSON parse error for error handling like this?
I think an invalid JSON output implies an unexpected error and it's more accurate than depending on STDERR output.

```ruby
begin
  end
rescue JSON::ParserError
  Results::Failure.new(guid: guid, analyzer: analyzer, message: "Unexpected error occurred")
end
```

Second, about the case of "Dockerfile not found", how about doing this?
We have checked already the message "Please provide a Dockerfile".

```ruby
if status.exitstatus == 1 && stderr.strip == "openBinaryFile: does not exist (No such file or directory)"
  return Results::Failure.new(guid: guid, analyzer: analyzer, message: "No Docker files found")
end
```

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,105 @@
...
+        id = file[:code]
+
+        Issue.new(
+          path: path,
+          location: Location.new(start_line: line),
+          id: id,
+          message: file[:message],
+          object: {
+            severity: file[:level],
+          },
```
ybiquitous: Please add also `schema:` like this:

```ruby
object: {
  severity: file[:level],
},
schema: Schema.issue,
```

lib/runners/processor/hadolint.rb
```diff
@@ -0,0 +1,105 @@
...
+          links: link_to_wiki(id),
+        )
+      end
+    end
+
+    def link_to_wiki(id)
+      if id.match(/^DL/)
```
ybiquitous: [nitpick] I recommend using [`start_with?`](http://ruby-doc.org/core-2.6.5/String.html#method-i-start_with-3F) for more readability.

```suggestion
      if id.start_with? "DL"
```

images/hadolint/Dockerfile.erb
ybiquitous: Oh, git-clone(1) has replaced with curl(1).

yykamei: I didn't notice the changes. Thank you. I should have marked resolved.

test/smokes/hadolint/expectations.rb
```diff
@@ -0,0 +1,58 @@
...
+  analyzer: {name: "hadolint", version: '1.17.4'},
+})
+
+Smoke.add_test("no_dockerfile", {
+  guid: "test-guid",
+  timestamp: :_,
+  type: "failure",
+  message: "hadolint: Dockerfile: openBinaryFile: does not exist (No such file or directory)\n",
+  analyzer: {name: "hadolint", version: '1.17.4'},
+})
```
ybiquitous: Please add a test case about [`.hadolint.yaml`](https://github.com/hadolint/hadolint#configure)

lib/runners/processor/hadolint.rb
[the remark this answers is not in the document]
mpls104: By the way, I don't come up with the use case of trusted-registry option in sider.
I think if user want to use hadolint for Dockerfile in some location, but in github, they don't use sider.
Do you think we do need to support this option?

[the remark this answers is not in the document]
yykamei: :thinking: I think it may be useful. For example, given such a `Dockerfile`:

```dockerfile
FROM 123.dkr.ecr.us-east-1.amazonaws.com/foo:1.2.3
RUN build something

FROM malicious.example.com/foo:1.2.3
CMD ["do", "something"]
```

The Dockerfile author accidentally wrote such a Dockerfile,
which contains `FROM malicious.example.com`.
Without the `--trusted-registry` option, hadolint reports nothing:

```shell-session
$ docker run --rm -i hadolint/hadolint hadolint - < Dockerfile
(nothing)
```

If users use this option like this, they can notice the malicious registry is written in Dockerfile. This might be an advantage for them, and they can modify the Dockerfile before deploying.

```shell-session
$ docker run --rm -i hadolint/hadolint hadolint --trusted-registry 123.dkr.ecr.us-east-1.amazonaws.com --trusted-registry docker.io - < Dockerfile
/dev/stdin:4 DL3026 Use only an allowed registry in the FROM image
```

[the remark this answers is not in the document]
mpls104: Oh I misunderstood the function of this option 😓  It's working as so to speak white list so it make sense. I will add this option 👍

test/smokes/hadolint/config_option/sider.yml
```diff
@@ -0,0 +1,3 @@
+linter:
+  hadolint:
+    config: .hadolint.yml
```
ybiquitous: [nitpick] Thank you for your adding the test case!

`.hadolint.yml` is **not** supported by default (correct is `.hadolint.yml`), so this file name is a bit confusing. 😅 

How about renaming like this?

```suggestion
    config: custom-hadolint.yml
```

See https://github.com/hadolint/hadolint#configure

ybiquitous:
@yykamei Could you please review this again?

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ All notable changes to this project will be documented in this file.
 
 - Renewal trace [#624](https://github.com/sider/runners/pull/624)
 
+- [hadolint] New support [#625](https://github.com/sider/runners/pull/625)
+
```
yykamei: The position of this entry should be in "Unreleased."

ybiquitous: Fixed!

ybiquitous:
I've pushed a bit refactoring via 0d3c2d4659aefe07bf57889d57e687e5aaabebb2.

ybiquitous:
I will merge after the CI passes.

ybiquitous:
Ignore Codacy.

ybiquitous:
Thank you! @mpls104 😊

PR #630, opened 2019-12-26
Add git_blame_info field in the Issue schema

yykamei:
Sider needs to detect the issues reported by tools are really "new
issues" or not.  To help Sider easily determine that, I implemented to
include "git_blame_info" in the Issue schema.

After this implementation, the succeeded result will be like that:

```json
{
  "result": {
    "issues": [
      {
        "id": "ID",
        "path": "app/models/user.rb",
        "git_blame_info": {
          "commit": "commit hash",
          "original_line": 3,
          "final_line": 5,
          "line_hash": "a56cfa66045cc9bb9983be19974153631bbce34a"
        },
        ...
      }
    ],
    ...
  }
}

```

* [x] Fix smoke tests


lib/runners/workspace/git.rb
```diff
@@ -59,5 +62,22 @@ def patches
         nil
       end
     end
+
+    def git_blame_infos(path_string, start_line, end_line)
+      base = git_source.base
+      head = git_source.head
+      if base && head
+        shell = Shell.new(current_dir: git_directory, trace_writer: trace_writer, env_hash: {})
+        stdout, _ = shell.capture3!("git", "blame", "-p", "-L", "#{start_line},#{end_line}", "#{base}...#{head}", "--", path_string)
```
yykamei: git-blame(1) with `-p` returns like this:

```
0d315f31deae73da5fbed5bfce1337f726b3b11f 3 3 1
author Masafumi Koba
author-mail <ybiquitous@gmail.com>
author-time 1568699162
author-tz +0900
committer Masafumi Koba
committer-mail <ybiquitous@gmail.com>
committer-time 1568707259
committer-tz +0900
summary Rename namespace: `NodeHarness` -> `Runners`
previous 97231097a0a0feeaf8b022b08ef8f6dd950b65aa lib/node_harness/cli.rb
filename lib/runners/cli.rb
	module Runners
763058476826086985fe3e9f541a34a43e7ae86a 4 4 1
	  class CLI
763058476826086985fe3e9f541a34a43e7ae86a 6 5 3
	    attr_reader :stdout
763058476826086985fe3e9f541a34a43e7ae86a 7 6
	    attr_reader :stderr
763058476826086985fe3e9f541a34a43e7ae86a 8 7
	    attr_reader :guid
```

yykamei:
@ybiquitous If you have time, I want you to review this PR. This is too big, but important parts are `lib` and `sig` directory :pray:

lib/runners/workspace/git.rb
ybiquitous: [suggestion]

- How about adding the link https://www.git-scm.com/docs/git-blame as a source comment?
- How about extracting a `GitBlameParser` class? I think it's easier to test.
  - E.g. `stdout, _ = shell.capture3!("git", "blame", ...); GitBlameParser.parse(stdout)`

ybiquitous: The format is defined on https://www.git-scm.com/docs/git-blame#_the_porcelain_format

yykamei: Yes, that's right. I'll add it. Thank you.

yykamei: I'm going to implement it like that because I want to the class is like a Rails model.

```ruby
module Runners
  class GitBlameInfo
    def self.parse(stdout)
      ...
    end

    def initialize(commit:, ...)
      @commit = commit
      ...
    end
  end
end
```

ybiquitous: 🙆‍♂

yykamei: Fixed with 37cc846

lib/runners/git_blame_info.rb
```diff
@@ -0,0 +1,44 @@
...
+    def eql?(other)
+      self == other
+    end
+
+    def ==(other)
+      self.class == other.class &&
+        commit == self.commit &&
+        original_line == self.original_line &&
+        final_line == self.final_line &&
+        line_hash == self.line_hash
```
ybiquitous: `line_hash == self.line_hash` always is true???

yykamei: I was clumsy :man_shrugging:

ybiquitous: Using `Struct` may be easier...

```ruby
GitBlameInfo = Struct.new(...) do
  def self.parse
    # ...
  end
end
```

yykamei: :man_shrugging: OK

yykamei: Here 7ad295f

yykamei:
Thank you for your time :pray:

ybiquitous:
No worry! 😃

PR #632, opened 2019-12-27
[Chore] Reformat test/smokes/**/expectations.rb

yykamei:
It was hard for me to add the new result field in all `expectations.rb`
files on #630. So, I just ran the following command using [Prettier for Ruby](https://github.com/prettier/plugin-ruby).

```shell-session
$ rbprettier --prefer-single-quotes false --print-width 120 \
    --write test/smokes/**/expectations.rb
```

I know the preference of styles is different between us,
but I think all of us prefer the `Hash` with `:` instead of the `Hash`
rocket.


ybiquitous:
What do you think about running the formatter on pre-commit hooke?

yykamei:
> What do you think about running the formatter on pre-commit hooke?

I like this idea because we don't have to discuss the styles with it.

ybiquitous:
Nice. I think [`left-hook`](https://github.com/Arkweid/lefthook) is suitable because we are using on sider-docs.

For example:

```yaml
pre-commit:
  commands:
    format-smoke-tests:
      glob: "test/smokes/**/expectations.rb"
      run: bundle exec rbprettier --write {all_files}
```

PR #633, opened 2019-12-27
Add leftfook and run prettier on pre-commit

yykamei:
See https://github.com/sider/runners/pull/632#issuecomment-569180123

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -1,0 +1,6 @@
+pre-commit:
+  parallel: true
+  commands:
+    prettier:
+      glob: "test/smokes/**/expectations.rb"
+      run: bundle exec rbprettier --write {staged_files} && git add {staged_files}
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+test/smokes/**/*.rb
+!test/smokes/**/expectations.rb
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,0 +1,5 @@
+overrides:
+  - files: "**/*.rb"
+    options:
+      preferSingleQuotes: false
+      printWidth: 120
--- a/Gemfile
+++ b/Gemfile
@@ -22,4 +22,6 @@ group :development, :test do
   gem 'rubocop', require: false
   gem 'meowcop', require: false
   gem 'aufgaben', git: 'https://github.com/ybiquitous/aufgaben.git', tag: '0.4.0'
+  gem 'lefthook', require: false
+  gem 'prettier', require: false
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,6 +46,7 @@ GEM
     jmespath (1.4.0)
     jsonseq (0.2.0)
     language_server-protocol (3.14.0.1)
+    lefthook (0.7.0)
     listen (3.2.1)
       rb-fsevent (~> 0.10, >= 0.10.3)
       rb-inotify (~> 0.9, >= 0.9.10)
@@ -57,6 +58,7 @@ GEM
     parallel (1.19.1)
     parser (2.7.0.0)
       ast (~> 2.4.0)
+    prettier (0.17.0)
     pry (0.12.2)
       coderay (~> 1.1.0)
       method_source (~> 0.9.0)
@@ -111,10 +113,12 @@ DEPENDENCIES
   bundler (>= 1.12, < 3.0)
   git_diff_parser
   jsonseq
+  lefthook
   locale
   meowcop
   minitest
   parallel
+  prettier
   querly
   rake
   retryable


lefthook.yml
```diff
@@ -0,0 +1,6 @@
+pre-commit:
+  parallel: true
+  commands:
+    prettier:
+      glob: "*.rb"
```
ybiquitous: Should we exclude Ruby files in smoke tests (except for `expectations.rb`)?

Also, I think it's better to format other `*.rb` files on one commit like #632 because `git blame` will be complex.

lefthook.yml
```diff
@@ -0,0 +1,6 @@
...
+      run: bundle exec rbprettier --print-width 120 --prefer-single-quotes false --write {staged_files} && git add {staged_files}
```
yykamei: I chose `{staged_files}` instead of `{all_files}`, and I think it's good to format all Ruby files.
I just copied the YAML file from [here](https://github.com/sider/sider-docs/blob/ed83c41facf985481cfc408e32bf9cc1bfad58e2/lefthook.yml#L4-L6).

lefthook.yml
yykamei: Oh, yes we should

lefthook.yml
ybiquitous: Nice 👍 

In addition, how about adding [`.prettierrc.yml`](https://prettier.io/docs/en/configuration.html)? I'd like to format via my editor.

yykamei: That's good!

ybiquitous: It seems good to use `overrides` option.

lefthook.yml
yykamei: I added `.prettierignore` file. What do you think?

lefthook.yml
yykamei: :question: 2387358 is not sufficient?

ybiquitous: This configuration file is applied to any files not only `*.rb` files.

So, I think safer to format gradually and carefully. If not, unexpected changes will be committed.

ybiquitous: For example:

```yaml
overrides:
  - files: "test/smokes/**/expectations.rb"
    options:
      preferSingleQuotes: false
      printWidth: 120
```

yykamei: :bulb: I see.

lefthook.yml
ybiquitous: As I commented on https://github.com/sider/runners/pull/633#discussion_r361579608, I think `glob: "test/smokes/**/expectaions.rb"` is safer.

ybiquitous:
When will you format `*.rb` except for `test/smokes/**/expectations.rb`? On this PR?

lefthook.yml
ybiquitous: > I added `.prettierignore` file. What do you think?

Good! 👍

yykamei:
I wasn't planning. Is it better to do before merging this PR?

ybiquitous:
No, let's keep this PR. 👍 

Let's format other `*.rb` files on another PR via one commit!

yykamei:
I ran `bundle exec rbprettier --write lib/**/*.rb test/**/*.rb` and got the result against the `lib/runners/io/aws.rb`

```diff
diff --git a/lib/runners/io/aws_s3.rb b/lib/runners/io/aws_s3.rb
index 3f6bc45..2d5a92c 100644
--- a/lib/runners/io/aws_s3.rb
+++ b/lib/runners/io/aws_s3.rb
@@ -26,17 +26,18 @@ module Runners
       @tempfile = Tempfile.new
       @written_items = 0

-      args = {
-        retry_limit: 5,
-        retry_base_delay: 1.2,
-        instance_profile_credentials_retries: 5,
-        instance_profile_credentials_timeout: 3,
-      }.tap do |hash|
-        if ENV["S3_ENDPOINT"]
-          hash[:endpoint] = ENV["S3_ENDPOINT"]
-          hash[:force_path_style] = true
+      args =
+        {
+          retry_limit: 5,
+          retry_base_delay: 1.2,
+          instance_profile_credentials_retries: 5,
+          instance_profile_credentials_timeout: 3
+        }.tap do |hash|
+          if ENV["S3_ENDPOINT"]
+            hash[:endpoint] = ENV["S3_ENDPOINT"]
+            hash[:force_path_style] = true
+          end
         end
-      end
       @client = Aws::S3::Client.new(**args)
     end

@@ -62,11 +63,7 @@ module Runners

     def flush_to_s3!
       tempfile.rewind
-      client.put_object(
-        bucket: bucket_name,
-        key: object_name,
-        body: tempfile.read, # A stream object is slower than a string object, so the read string is passed here.
-      )
+      client.put_object()
       @written_items = 0
       tempfile.seek 0, ::IO::SEEK_END
     end
```

The arguments of `#pub_object` have disappeared!
I won't format the production code for now...

ybiquitous:
😨

ybiquitous:
I think valuable even if only `expecations.rb`. 😅

yykamei:
Yes, I agree. I want to tackle the issue of Prettier for Ruby if I have time.

PR #654, opened 2020-01-08
Include :git_blame_info in `Issue` instances

yykamei:
Make Runners add `:git_blame_info` for every issue.

Also, I renamed the API `Workspace#git_blame_info` to
`Workspace#range_git_blame_info`, and declare `Processor#git_blame_info`
because I want to relate the method name with the `GitBlameInfo` class.

### DIFF
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -6,7 +6,6 @@ module Runners
 
     delegate :push_dir, :current_dir, :capture3, :capture3!, :capture3_trace, :capture3_with_retry!, to: :shell
     delegate :env_hash, :push_env_hash, to: :shell
-    delegate :git_blame_info, to: :workspace
 
     def initialize(guid:, workspace:, git_ssh_path:, trace_writer:)
       @guid = guid
@@ -238,6 +237,10 @@ module Runners
 
     def show_runtime_versions
       # noop by default
+    end
+
+    def git_blame_info(path_string, start_line)
+      workspace.range_git_blame_info(path_string, start_line, start_line).last
     end
   end
 end
--- a/lib/runners/processor/brakeman.rb
+++ b/lib/runners/processor/brakeman.rb
@@ -61,7 +61,8 @@ module Runners
             location: loc,
             id: "#{warning[:warning_type]}-#{warning[:warning_code]}",
             message: warning[:message],
-            links: [warning[:link]]
+            links: [warning[:link]],
+            git_blame_info: git_blame_info(warning[:file], line),
           )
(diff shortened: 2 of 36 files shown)


yykamei:
@ybiquitous I added `git_blame_info` in each processor. What do you think?

ybiquitous:
Cannot you calculate `git_blame_info` within `Issue` class?

yykamei:
Maybe. I have to make `Issue` depend on `Workspace` if so.

ybiquitous:
Yes, for example:

```ruby
class Issue
  def initialize(path:, location:, workspace:)
    @git_blame_info = workspace.range_git_blame_info(path, location.start_line, location.end_line || location.start_line).last
  end
end
```

I think each processor can decrease its implementation cost.

```ruby
Issue.new(
  path: path,
  location: loc,
  id: "#{warning[:warning_type]}-#{warning[:warning_code]}",
  message: warning[:message],
  links: [warning[:link]],
  git_blame_info: git_blame_info(warning[:file], line), #<= Read attributes mutiple times. Not DRY.
)
```

yykamei:
I try.

yykamei:
Do you have an idea about passing an instance of `Workspace` to `Issue`?

ybiquitous:
I think as follow:

```ruby
Issue.new(
  path: path,
  location: loc,
  id: "#{warning[:warning_type]}-#{warning[:warning_code]}",
  message: warning[:message],
  links: [warning[:link]],
  workspace: workspace,
)
```

ybiquitous:
I'm worried also if a performance issue will raise, which is calling `shell.capture3!("git", "blame")` for each issue.
It is similar to the N+1 problem and may overflow trace logs.

yykamei:
Yes, I thought the performance issues, too. However, I cannot estimate how much the problem is.

By the way, I'm worried about `Issue` depends on `Workspace`.  Anyway, I'll do another implementation. Close this PR.

ybiquitous:
> I'm worried about `Issue` depends on `Workspace`.

Surely, I think also it's not so good design 😅

yykamei:
Thank you for your review!

yykamei:
I might be able to mitigate the N+1 problem by calling git-blame(1) for filtered issues after calling `Results::Success#filter_issues`. It's not a perfect solution to avoid N+1, but I could reduce the traces with it. I implemented this with #655.

ybiquitous:
> It's not a perfect solution to avoid N+1

Yes, it seems no other way... 🤷‍♂

PR #655, opened 2020-01-08
Add #add_git_blame_info

yykamei:
Make Runners add `:git_blame_info` for every issue via
`Results::Success#add_git_blame_info`.

Also, I renamed the API `Workspace#git_blame_info` to
`Workspace#range_git_blame_info`, and declare `Processor#git_blame_info`
because I want to relate the method name with the `GitBlameInfo` class.

See also #654

### DIFF
--- a/lib/runners/results.rb
+++ b/lib/runners/results.rb
@@ -68,6 +68,12 @@ module Runners
         end
         result
       end
+
+      def add_git_blame_info(workspace)
+        issues.each do |issue|
+          issue.add_git_blame_info(workspace)
+        end
+      end
     end
 
     # Result to indicate that processing failed by some error.
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -1,5 +1,17 @@
 module Runners
   class Workspace::Git < Workspace
+    def range_git_blame_info(path_string, start_line, end_line)
+      base = git_source.base
+      head = git_source.head
+      if base && head
+        shell = Shell.new(current_dir: git_directory, trace_writer: trace_writer, env_hash: {})
+        stdout, _ = shell.capture3!("git", "blame", "-p", "-L", "#{start_line},#{end_line}", "#{base}...#{head}", "--", path_string, trace_stdout: false, trace_stderr: false)
+        GitBlameInfo.parse(stdout)
+      else
+        []
+      end
+    end
+
     private
 
     def prepare_base_source(dest)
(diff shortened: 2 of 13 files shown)


yykamei:
@ybiquitous What do you think? I think this implementation reduces the call of git-blame(1).

https://github.com/sider/runners/pull/654#issuecomment-571953818

lib/runners/workspace/git.rb
```diff
@@ -1,5 +1,17 @@
 module Runners
   class Workspace::Git < Workspace
+    def range_git_blame_info(path_string, start_line, end_line)
+      base = git_source.base
+      head = git_source.head
+      if base && head
+        shell = Shell.new(current_dir: git_directory, trace_writer: trace_writer, env_hash: {})
+        stdout, _ = shell.capture3!("git", "blame", "-p", "-L", "#{start_line},#{end_line}", "#{base}...#{head}", "--", path_string)
```
ybiquitous: How about suppressing verbose output to trace as follow?

```suggestion
        stdout, _ = shell.capture3!("git", "blame", "-p", "-L", "#{start_line},#{end_line}", "#{base}...#{head}", "--", path_string, trace_stdout: false, trace_stderr: false)
```

lib/runners/results.rb
```diff
@@ -68,6 +68,24 @@ def filter_issues(changes)
...
+              message: issue.message,
+              links: issue.links,
+              object: issue.object,
+              workspace: workspace,
+            )
+          )
+        end
+        result
+      end
     end
```
ybiquitous: Looks good 👍 

BTW, how about changing the `#issues` array instead of creating a new instance of `Result`?
This class has already the mutating method `#add_issue`, so it seems unneeded to me to ensure immutability for `#add_git_blame_info` and `#filter_issues` methods. 🤔 
Perhaps I think we can get the advantages of fewer code lines and less memory usage if doing so.

For example:

```ruby
def add_git_blame_info(workspace)
  issues.each do |issue|
    issue.add_git_blame_info(workspace)
  end
end
```

lib/runners/workspace/git.rb
yykamei: Oh, I forgot the arguments `:trace_stdout` and `:trace_stderr`. Thank you.

lib/runners/results.rb
yykamei: Yes, I agree. I thought a similar method like `#add_git_blame_info` at first, but I followed `#filter_issues` and implemented immutability as possible. I'll do that.

yykamei: Fixed with d564870

ybiquitous: Good! How about refactoring also `#filter_issues`?

lib/runners/workspace/git.rb
```diff
@@ -1,5 +1,17 @@
...
+      head = git_source.head
+      if base && head
+        shell = Shell.new(current_dir: git_directory, trace_writer: trace_writer, env_hash: {})
+        stdout, _ = shell.capture3!("git", "blame", "-p", "-L", "#{start_line},#{end_line}", "#{base}...#{head}", "--", path_string, trace_stdout: false, trace_stderr: false)
+        GitBlameInfo.parse(stdout)
+      else
+        []
+      end
+    end
+
```
ybiquitous: If a performance issue would raise in the future, we might need to cache the result of the method. But it is not needed to consider now.

lib/runners/results.rb
yykamei: OK, looks good. I'll refactor it in another PR.

PR #658, opened 2020-01-09
Refactor Results::Success#filter_issues to mutate its attribute

yykamei:
See https://github.com/sider/runners/pull/655#discussion_r364537300

### DIFF
--- a/lib/runners/harness.rb
+++ b/lib/runners/harness.rb
@@ -49,8 +49,9 @@ module Runners
               case result
               when Results::Success
                 trace_writer.message "Removing issues from unchanged or untracked files..." do
-                  result.filter_issues(changes).tap do |new_result|
-                    new_result.add_git_blame_info(workspace)
+                  result.tap do |r|
+                    r.filter_issues(changes)
+                    r.add_git_blame_info(workspace)
                   end
                 end
               else
--- a/lib/runners/results.rb
+++ b/lib/runners/results.rb
@@ -62,11 +62,9 @@ module Runners
       end
 
       def filter_issues(changes)
-        result = self.class.new(guid: guid, analyzer: analyzer)
-        issues.each do |issue|
-          result.add_issue(issue) if changes.include?(issue)
+        issues.select! do |issue|
+          changes.include?(issue)
         end
-        result
       end
 
       def add_git_blame_info(workspace)
--- a/sig/runners.rbi
+++ b/sig/runners.rbi
@@ -65,7 +65,7 @@ class Runners::Results::Success < Runners::Results::Base
 
   def initialize: (guid: String, analyzer: Analyzer) -> any
   def add_issue: (Runners::Issue) -> void
-  def (constructor) filter_issues: (Changes) -> instance
+  def filter_issues: (Changes) -> void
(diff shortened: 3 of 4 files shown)


lib/runners/harness.rb
```diff
@@ -49,8 +49,9 @@ def run
               case result
               when Results::Success
                 trace_writer.message "Removing issues from unchanged or untracked files..." do
-                  result.filter_issues(changes).tap do |new_result|
-                    new_result.add_git_blame_info(workspace)
+                  result.tap do |r|
+                    r.filter_issues(changes)
+                    r.add_git_blame_info(workspace)
```
ybiquitous: `tap` is needed?

```ruby
result.filter_issues(changes)
result.add_git_blame_info(workspace)
```

ybiquitous: Sorry, it's needed.

lib/runners/results.rb
```diff
@@ -62,11 +62,9 @@ def add_issue(issue)
       end
 
       def filter_issues(changes)
-        result = self.class.new(guid: guid, analyzer: analyzer)
-        issues.each do |issue|
-          result.add_issue(issue) if changes.include?(issue)
+        issues.reject! do |issue|
```
ybiquitous: How about this? See http://ruby-doc.org/core-2.7.0/Array.html#//dash_ref_method%252Di%252Dfilter%252D21/Method/filter!/0

```suggestion
        issues.filter! do |issue|
```

yykamei: Oh, I mistakenly used `reject!`! I should have chosen `filter!`. Thank you!

yykamei: I replaced `reject!` with `filter!` and did git-commit(1) with `--amend`.

yykamei: Sorry, reject! is correct. I'm looking into the error on CI.

yykamei: `filer!` is correct...

sig/polyfill.rbi
```diff
@@ -60,6 +60,7 @@ extension Array (Polyfill)
   def +: <'x> (Array<'x>) -> Array<'a | 'x>
        | (self) -> self
   def filter: { ('a) -> bool } -> self
+  def filter!: { ('a) -> bool } -> void
```
yykamei: `Array#filter!` returns `nil` or its instance, but we should not depend on the response of this method, so I set the return type `void`.

https://ruby-doc.org/core-2.7.0/Array.html#method-i-filter-21

ybiquitous:
Why...? 

> NoMethodError: undefined method `filter!' for #<Array:0xXXXXXX>

ybiquitous:
`#Array:0xXXXXXX` does not seem an `Array`... 🤔

yykamei:
I understood. `Array#filter!` does not exist in Ruby 2.5. I'll use `Array#select! instead.

sig/polyfill.rbi
ybiquitous: Is this unneeded?

yykamei: Yes. Thank you

PR #683, opened 2020-01-28
Add more plugins to OPTIONAL_GEMS

### LINKED ISSUE #682: Let users use RuboCop plugins without sider.yml
Our runner_rubocop tries to install RuboCop plugins based on the following declarations,
but this lacks some plugins even though these are maintained in [rubocop-hq](https://github.com/rubocop-hq).

https://github.com/sider/runners/blob/0b931988f23abe3324ed4dc262ec0d60c54a46ae/lib/runners/processor/rubocop.rb#L27-L52

We should add such plugins (and, we should remove the comment `DEPRECATED` because this list is necessary for Sider).

yykamei:
Close #682

### DIFF
--- a/test/smokes/rubocop/rails_and_performance/sider.yml
+++ b/test/smokes/rubocop/rails_and_performance/sider.yml
@@ -1,5 +1,0 @@
-linter:
-  rubocop:
-    gems:
-      - { name: rubocop-rails, version: 2.3.2 }
-      - { name: rubocop-performance, version: 1.5.0 }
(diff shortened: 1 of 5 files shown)


test/smokes/rubocop/rails_and_performance/sider.yml
```diff
@@ -1,5 +0,0 @@
-linter:
-  rubocop:
-    gems:
-      - { name: rubocop-rails, version: 2.3.2 }
-      - { name: rubocop-performance, version: 1.5.0 }
```
yykamei: We should continue analyses without `sider.yml` if the `Gemfile` contains our supported plugins.

yykamei:
Thank you :pray:

PR #702, opened 2020-02-04
Add warnings if the analysis tries to use Python 2

yykamei:
I updated the documentation [here](https://help.sider.review/tools/python/flake8) to notify users about the deprecation of Python 2.
Runners also should warn about the deprecation.


yykamei:
Ideally, I think `Smoke.add_test` should receive only one `Hash` instance without options, and it always assumes the parameter is `Runners::Schema::Types::Result.envelope`.

PR #721, opened 2020-02-12
Drop older versions of ESLint

### LINKED ISSUE #715: Drop ESLint "< 5.0.0"
Drop the support of the older version of ESLint.

https://help.sider.review/news/2019#update-languages-and-tools-on-dec-17-2019
> [Announcement] We are planning to drop the support of ESLint and RuboCop older versions on January 31, 2020. After the deadline, all analyses by the older versions will fail. If you are using the older versions, please upgrade them by the deadline. Here are new versions:
> * ESLint 5.0.0+
> * RuboCop 0.61.0+

yykamei:
Close #715


test/smokes/eslint/expectations.rb
```diff
@@ -30,44 +30,6 @@
   }
 )
 
-Smoke.add_test(
-  "sideci_valid_npm_install_option",
```
yykamei: I think this test is no longer necessary because other tests are checking the validity of `sider.yml`.

test/smokes/eslint/expectations.rb
```diff
@@ -78,157 +40,37 @@
...
-    ESLint couldn't find the plugin "eslint-plugin-filenames".
+          ESLint couldn't find the plugin "eslint-plugin-filenames".
   MESSAGE
       ),
     analyzer: { name: "ESLint", version: "6.8.0" }
   }
 )
 
 Smoke.add_test(
-  "dir_option_is_array",
```
yykamei: This test should be written as a unit test. In the first place, testing this feature is overplaying to me.

test/smokes/eslint/expectations.rb
```diff
@@ -78,157 +40,37 @@
...
+  }
 )
 
 Smoke.add_test(
   "no_files",
   { guid: "test-guid", timestamp: :_, type: "success", issues: [], analyzer: { name: "ESLint", version: "6.8.0" } }
 )
 
-Smoke.add_test(
-  "no_files_eslint4",
```
yykamei: This test is satisfied with `pinned_old_eslint`.

lib/runners/processor/eslint.rb
```diff
@@ -59,7 +58,6 @@ def setup
           return Results::Failure.new(guid: guid, message: exn.message, analyzer: nil)
         end
         analyzer
-        add_warning_if_deprecated_version(minimum: RECOMMENDED_MINIMUM_VERSION, file: "package.json", deadline: Time.new(2020, 1, 31))
         yield
       end
     end
```
ybiquitous: It seems we can fix also this:

https://github.com/sider/runners/blob/027d777ca046d671f32671d764cddb78997569b7/lib/runners/processor/eslint.rb#L246-L249

yykamei: Thank you. Is [this](7772312) correct?

ybiquitous: Looks good!

BTW, `[this](7772312)` link points to "404 Not Found" 😅

PR #731, opened 2020-02-12
Add FxCop support

### LINKED ISSUE #724: Support FxCop analyzers
## Information

- Website: https://docs.microsoft.com/en-us/visualstudio/code-quality/install-fxcop-analyzers?view=vs-2019
- Source code:https://github.com/dotnet/roslyn-analyzers
- Language: C#

## Why do you need the tool?

- Because FxCop is official

## Additional context

None

## See also

None

gnakagaw:
Fix #724 

## ToDo
- [x] Add CI configuration
- [x] modify documents (CHANGELOG, README)
- [x] write images/fxcop_analyzers/Dockerfile.erb
- [x] add 'require' declaration to lib/runners.rb
- [x] implement lib/runners/processor/fxcop_analyzers.rb
- [x] implement smoke tests


lib/runners/processor/fxcop_analyzers.rb
```diff
@@ -0,0 +1,113 @@
...
+            end
+            rval.append(
+              {
+                "file": file,
+                "rule_id": rule_id,
+                "message": message,
+                "start_line": loc_info.fetch('startLine'),
+                "start_column": loc_info.fetch('startColumn'),
+                "end_line": loc_info.fetch('endLine'),
+                "end_column": loc_info.fetch('endColumn')
```
yykamei: You can write the Hash keys without quotation.

test/smokes/fxcop_analyzers/expectations.rb
```diff
@@ -0,0 +1,56 @@
...
+    type: 'success',
+    issues: [
+      {
+        path: 'Program.cs',
+        location: { start_line: 9, start_column: 31, end_line:9, end_column:45 },
+        id: 'CA1303',
+        message: "Method 'void Program.Main(string[] args)' passes a literal string as parameter 'value' of a call to 'void Console.WriteLine(string value)'. Retrieve the following string(s) from a resource table instead: \"Hello World!\".",
+        object: nil,
+        git_blame_info: nil,
+        links: []
```
yykamei: It's useful to see the links about the rule ID. This CA1303 seems explained here.

https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1303?view=vs-2019

I think the function which generates the link is required like this:

https://github.com/sider/runners/blob/af837af5de18374bdfc7ed40fb9407e32af70487/lib/runners/processor/rubocop.rb#L243-L246

test/smokes/fxcop_analyzers/expectations.rb
```diff
@@ -0,0 +1,56 @@
...
+    guid: 'test-guid',
+    timestamp: :_,
+    type: 'success',
+    issues: [
+      {
+        path: 'Program.cs',
+        location: { start_line: 9, start_column: 31, end_line:9, end_column:45 },
+        id: 'CA1303',
+        message: "Method 'void Program.Main(string[] args)' passes a literal string as parameter 'value' of a call to 'void Console.WriteLine(string value)'. Retrieve the following string(s) from a resource table instead: \"Hello World!\".",
+        object: nil,
```
yykamei: FxCopAnalyzers looks to have information about "level". It's convenient for users to see the level (or severity) of the rule.

For example, runner_eslint provides extra information like this:

https://github.com/sider/runners/blob/af837af5de18374bdfc7ed40fb9407e32af70487/lib/runners/processor/eslint.rb#L188-L192

lib/runners/processor/fxcop_analyzers.rb
```diff
@@ -0,0 +1,113 @@
...
+      unless status == 0
+        msg = <<~EOS
+        Failed to restore .NET Core Project. You have to put project file (.csproj) under the analysis root. Please check project structure and Sider config file (sider.yml).
+        Note: We only support a project managed with MSBuild. No support for other build systems. (e.g. Nuke, Cake)
+        EOS
+        return Results::Failure.new(guid: guid, message: msg, analyzer: analyzer)
+      end
+
+      # install FxCop Analyzers from nuget Repository
+      cmdline_add_analyzer = ['dotnet', 'add', 'package', 'Microsoft.CodeAnalysis.FxCopAnalyzers', '--version', FXCOP_ANALYZER_VERSION]
```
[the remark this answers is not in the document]
gnakagaw: >  it's impossible to install FxCopAnalyzers globally. Is it correct?
Yes, it is. We have to add a Nuget dependency for each target project. (Nuget is a dependency management system such as npm).

lib/runners/processor/fxcop_analyzers.rb
```diff
@@ -0,0 +1,113 @@
...
+      _, _, status = capture3(*cmdline_add_analyzer)
```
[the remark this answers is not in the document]
gnakagaw: You are right. This is NOT user's fault. I'll modify it.

lib/runners/processor/fxcop_analyzers.rb
```diff
@@ -0,0 +1,113 @@
...
+      yield
+    end
+
+    def analyze(changes)
+      # run 'dotnet build' with static analysis module
+      cmdline_run_analyzer = ['dotnet', 'build', '--no-incremental', "-property:errorlog=#{ANALYSIS_LOGFILE_PATH}"]
+      capture3!(*cmdline_run_analyzer)
+
+      # parse static analysis results
+      unless File.exist?(ANALYSIS_LOGFILE_PATH)
```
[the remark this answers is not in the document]
gnakagaw: `ANALYSIS_LOGFILE_PATH` is not a parameter from user's. It is static constant parameter defined in this file. Thus, any failures caused by user don't happen.

See: https://github.com/gnakagaw/runners/blob/add_fx_cop_analyzer_support/lib/runners/processor/fxcop_analyzers.rb#L9

lib/runners/processor/fxcop_analyzers.rb
gnakagaw: Thx!

test/smokes/fxcop_analyzers/expectations.rb
gnakagaw: A output file from the analyzer contains helpUri information. I've modified get it.

test/smokes/fxcop_analyzers/expectations.rb
gnakagaw: I've added "level" information from the analyzers to outputs.

gnakagaw:
@ybiquitous FxCop Analyzers is NOT identical with FxCop strictly. FxCop is a old obsoleted analysis tool for .NET environment. FxCop Analyzer is a porting of FxCop for Roslyn compiler platform. 

But, I think that it's ok to call this runner as fxcop because its rule is identical to FxCop.

FxCop (Wikipedia): https://en.wikipedia.org/wiki/FxCop
FxCop document (last update: 2007) https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-3.0/bb429476(v=vs.80)?redirectedfrom=MSDN

FxCop Analyzers documents say that "This is the primary analyzer package for this repo that contains all the ported FxCop code analysis rules (CAxxxx). "
https://github.com/dotnet/roslyn-analyzers#microsoftcodeanalysisfxcopanalyzers

ybiquitous:
@gnakagaw Thanks for the detail explanation. Please rename it to `fxcop`!

For users configuring this runner via `sider.yml`, I think `fxcop` is more useful: 😃 

```yaml
linter:
  fxcop:  # nice! :)
    # ...
  fxcop_analyzers:  # too long... :(
    # ...
```

lib/runners/processor/fxcop_analyzers.rb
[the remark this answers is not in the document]
yykamei: Thank you for your explanation. How about using cache? Nuget seems to have cache system.

https://docs.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders

By the way, I said directly, but I'm curious what happens if users have already added `Microsoft.CodeAnalysis.FxCopAnalyzers`.

lib/runners/processor/fxcop_analyzers.rb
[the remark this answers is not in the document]
yykamei: Yes, I know the constant. So, it's our fault if the file does not exist after running `dotnet build`. It's not necessary to check the `status`, and we should use `capture3!`.

lib/runners/processor/fxcop_analyzers.rb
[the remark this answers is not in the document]
gnakagaw: > How about using cache?

I add a caching process to image building process. 

See: https://github.com/sider/runners/pull/731/commits/3a12a8eb199f37d0c7b61bfa4e76fc2b72c60d21

> but I'm curious what happens if users have already added Microsoft.CodeAnalysis.FxCopAnalyzers.

No problem. If ever I execute "dotnet add ..." command multiple times, there are any duplicated the dependency configurations in the project settings file(.csproj). I wrote a new smoke test case.

See: https://github.com/sider/runners/pull/731/commits/20a13db64d532f9d502337c5984831ade87e2b9a

I also test a case that the target repo has a configuration for old version FxCop Analyzers. It's OK. We can overwrite user defined version with our version. I added a smoke test case for that. 

See: https://github.com/sider/runners/pull/731/commits/b9a971b5be24c503c5ccd2db6851da08e58abd68

gnakagaw:
@ybiquitous I replaced 'fxcop_analyzers' -> 'fxcop'.
See: https://github.com/sider/runners/pull/731/commits/736cab4550bb18efeae407436451c8875e3de40c and https://github.com/sider/runners/pull/731/commits/dd23b85ef446dd11250d3145b80e567050bcf863

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 [Full diff](https://github.com/sider/runners/compare/0.20.0...HEAD)
 
 - Show stderr of git-blame(1) [#756](https://github.com/sider/runners/pull/756)
+- [fxcop] New support [#731](https://github.com/sider/runners/pull/731)
```
ybiquitous: Thanks for renaming! Here is a small suggestion: 

```suggestion
- [FxCop] New support [#731](https://github.com/sider/runners/pull/731)
```

images/fxcop/Dockerfile.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<%= render_erb 'images/Dockerfile.base.erb' %>
+
+# cache target analyzer
+USER $RUNNER_USER
+RUN cd && \
+    mkdir dummy_project && \
+    cd dummy_project && \
+    dotnet new console && \
+    dotnet add package Microsoft.CodeAnalysis.FxCopAnalyzers --version 2.9.8 && \
```
ybiquitous: The global installation is nice! 👍 
In addition, what about using `ARG` like this for better git history?

```dockerfile
ARG FXCOP_VERSION=2.9.8

RUN dotnet add package Microsoft.CodeAnalysis.FxCopAnalyzers --version ${FXCOP_VERSION}
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
+module Runners
+  # A processor class for FxCop Analyzers
+  # https://docs.microsoft.com/en-us/visualstudio/code-quality/install-fxcop-analyzers
+  # This is a beta release. We support only repositories that has .csproj file directly under analysis root directory.
+  class Processor::FxCop < Processor
+
+    #parameters
+    FXCOP_ANALYZER_VERSION = '2.9.8'.freeze
```
ybiquitous: We want to make the version definition DRY, so can we get the version at runtime?

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+        i.fetch('results').each do |i2|
+          rule_id = i2.fetch('ruleId')
+          message = i2.fetch('message')
+          level = i2.fetch('level')
+          loc_info = i2.fetch('locations').fetch(0).fetch('resultFile').fetch('region')
+          file = i2.fetch('locations').fetch(0).fetch('resultFile').fetch('uri').sub(/^file:\/\//, '')
+          link = rules.fetch(rule_id).fetch('helpUri')
+          # skip issues if the rule id is NOT for FxCop Analyzers
+          unless rule_id =~ RULE_ID_PATTERN
+            continue
```
ybiquitous: Is it correct?

```suggestion
            next
```

lib/runners/processor/fxcop_analyzers.rb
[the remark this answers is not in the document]
gnakagaw: I've understood wrong your comment. I got it. 

I removed the file existing checking and the raising Result::Failure because we can check it and raise a exception in `File.open("#{ANALYSIS_LOGFILE_PATH}")`.  
See: https://github.com/sider/runners/pull/731/commits/f2775d7aca20e789a4b1fbdb972808d6089e2c2e

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      yield
+    end
+
+    def analyze(changes)
+      # run 'dotnet build' with static analysis module
+      cmdline_run_analyzer = ['dotnet', 'build', '--no-incremental', "-property:errorlog=#{ANALYSIS_LOGFILE_PATH}"]
+      capture3!(*cmdline_run_analyzer)
+
+      issues = parse_result_file(File.open("#{ANALYSIS_LOGFILE_PATH}"))
```
ybiquitous: [suggestion] `#read` ensures closing, so it is better:

```suggestion
      issues = parse_result_file(File.read(ANALYSIS_LOGFILE_PATH))
```

See https://ruby-doc.org/core-2.7.0/IO.html#method-c-read

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+            end_column: issue[:end_column]
+          )
+          result.add_issue Issue.new(
+            path: relative_path(issue[:file]),
+            location: loc,
+            id: issue[:rule_id],
+            message: issue[:message],
+            links: [issue[:link]],
+            object: {
+              level: issue[:level]
```
ybiquitous: Can you please use `severity` instead of `level` for consistency with other tools?

```ruby
severity: issue[:level]
```

images/fxcop/Dockerfile.erb
gnakagaw: I fixed it.
https://github.com/sider/runners/pull/731/commits/2140d790eb76d742f67e64304723f266b00b4526

lib/runners/processor/fxcop.rb
gnakagaw: I fixed it.
https://github.com/sider/runners/pull/731/commits/2140d790eb76d742f67e64304723f266b00b4526

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      # parse analysis results and extract information
+      rval = []
+      json.fetch('runs').each do |i|
+        i.fetch('results').each do |i2|
+          rule_id = i2.fetch('ruleId')
+          message = i2.fetch('message')
+          level = i2.fetch('level')
+          loc_info = i2.fetch('locations').fetch(0).fetch('resultFile').fetch('region')
+          file = i2.fetch('locations').fetch(0).fetch('resultFile').fetch('uri').sub(/^file:\/\//, '')
```
yykamei: Instead of using `#sub`, how about using `URI`? For me, regular expressions are hard to read.

```ruby
i2.fetch('locations').fetch(0).fetch('resultFile').fetch('uri').yield_self { |v| URI.parse(v).path }
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      yield
+    end
+
+    def analyze(changes)
+      # run 'dotnet build' with static analysis module
+      cmdline_run_analyzer = ['dotnet', 'build', '--no-incremental', "-property:errorlog=#{ANALYSIS_LOGFILE_PATH}"]
+      capture3!(*cmdline_run_analyzer)
+
+      issues = parse_result(File.read(ANALYSIS_LOGFILE_PATH))
```
ybiquitous: [suggestion] I suggest a refactoring idae for temp output file:

```ruby
output_file = Tempfile.new("fxcop-")

cmdline_run_analyzer = ['dotnet', 'build', '--no-incremental', "-property:errorlog=#{output_file.path}"]

issues = parse_result(output_file.read)
```

[`Tempfile.new`](https://ruby-doc.org/stdlib-2.7.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-new) creates a temporary file to be deleted automatically when the program exits.

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    # parse error log file(static analysis log file) from .NET Core Compilers
+    def parse_result(f)
+      json = JSON.parse(f)
+
+      # parse rule information
+      rules = {}
+      json.fetch('runs').each do |i|
+        rules.merge!(i.fetch('rules'))
+      end
```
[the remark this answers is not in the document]
ybiquitous: Does `runs[n].rules` have always the same value in the same `runs`?
If so, we may be able to do as follows:

```ruby
rules = json.dig('runs', 0, 'rules')
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      # parse analysis results and extract information
+      rval = []
+      json.fetch('runs').each do |i|
+        i.fetch('results').each do |i2|
```
ybiquitous: [suggestion] When using [`flat_map`](https://ruby-doc.org/core-2.7.0/Enumerable.html#method-i-flat_map), `filter` and `map` instead of nested `each`, we can write it more simply:

```ruby
# skip issues if the rule id is NOT for FxCop Analyzers
by_rule = ->(res) { res.fetch('ruleId') =~ RULE_ID_PATTERN }

json.fetch('runs').flat_map do |i|
  i.fetch('results').filter(&by_rule).map do |i2|
    {
      file: file,
      rule_id: rule_id,
      # ...
    }
  end
end
```

We don't need to use a local variable such as `rval`.

Also, `i` and `i2` seem uncommunicative to me, so what about renaming like this?

```ruby
json.fetch('runs').flat_map do |run|
  run.fetch('results').filter(&by_rule).map do |res|
    # ...
  end
end
```

ybiquitous: Or, when using `yield`, `flat_map` and `map` is unneeded:

```ruby
  parse_result(output_file.read) do |issue|
    result.add_issue(issue)
  end
end

# ...

json.fetch('runs').each do |run|
  run.fetch('results').filter(&by_rule).each do |res|
    yield Issue.new(...)
  end
end
```

FYI. https://github.com/sider/runners/blob/f043b545358d710209f3eed3d2dc508825d813d9/lib/runners/processor/cpplint.rb#L100

lib/runners/processor/fxcop.rb
ybiquitous: And, we need passing a `schema`:

```ruby
object: {
  # ...
},
schema: Schema.issue,
```

FYI. https://github.com/sider/runners/blob/f043b545358d710209f3eed3d2dc508825d813d9/lib/runners/processor/cpplint.rb#L108

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+    FXCOP_ANALYZER_VERSION = ENV['FXCOP_VERSION']
+    ANALYSIS_LOGFILE_PATH = '/tmp/sider_analysis_out.json'.freeze
+    RULE_ID_PATTERN = /CA[0-9]+/.freeze
+
+    def self.ci_config_section_name
+      'fxcop'
+    end
+
+    def analyzer_name
+      'fxcop'
```
ybiquitous: Can you change it?

```suggestion
      'FxCop'
```

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
gnakagaw: > Does runs[n].rules have always the same value in the same runs?

It is difficult to answer. According to the specification of analysis output, it may be possible take another value.
http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html#_Toc516224702

Actually, I have only see outputs that runs has only single element (e.g. n = 0). I think that it's ok to use `Hash#dig`. But, if we obey the spec., we should check each elements in runs array.

lib/runners/processor/fxcop.rb
gnakagaw: I was looking for such method!! Thx!

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
ybiquitous: Make sense. Let's keep it as-is. 👍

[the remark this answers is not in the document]
ybiquitous: I also agree with using `each_with_object`. 👍

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+      FXCOP_ANALYZER_VERSION
+    end
+
+    def setup
+      # restore project (install dependencies)
+      cmdline_install_dependency = %w(dotnet restore)
+      _, _, status = capture3(*cmdline_install_dependency)
+      unless status == 0
+        msg = <<~EOS
+          Failed to restore .NET Core Project. You have to put project file (.csproj) under the analysis root. Please check project structure and Sider config file (sider.yml).
```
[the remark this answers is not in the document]
gnakagaw: > See also Runners::Processor#ci_config_path_name method.

ok. I fixed it.

> currently can we configure FxCop via sider.yml? 

We can use 'root_dir' option without `Schema` definition. 
https://help.sider.review/getting-started/custom-configuration#root_dir-option

I modified the message to make clear that.
```
Failed to restore .NET Core Project. You have to put project file (.csproj) under the analysis root. Please check the project structure and 'root_dir' parameter in #{ci_config_path_name}.
See: https://help.sider.review/getting-started/custom-configuration#root_dir-option
Note: We only support a project managed with MSBuild. No support for other build systems. (e.g. Nuke, Cake)
```

lib/runners/processor/fxcop.rb
gnakagaw: cool solution! Thx!

lib/runners/processor/fxcop.rb
gnakagaw: I modified how to manage output file.
https://github.com/sider/runners/pull/731/commits/9de36274546eb4eabbb0232d0d168e583b3e1c04

lib/runners/processor/fxcop.rb
gnakagaw: I fixed it.
https://github.com/sider/runners/pull/731/commits/afe2785678699e670f4eb285f9025d9c78b466c8

lib/runners/processor/fxcop.rb
gnakagaw: I refactored `parse_result` method. I use `Hash#dig` as possible.

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
gnakagaw: Fix: https://github.com/sider/runners/pull/731/commits/d6d814a218cd7f44b8b2ff8dd0754e2ae6a7a7fb

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    def analyzer_version
+      FXCOP_ANALYZER_VERSION
+    end
+
+    def setup
+      # restore project (install dependencies)
+      cmdline_install_dependency = %w(dotnet restore)
+      _, _, status = capture3(*cmdline_install_dependency)
+      unless status == 0
```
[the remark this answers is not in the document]
gnakagaw: Fix: https://github.com/sider/runners/pull/731/commits/5d5859ec7a7e1e39e2766b040f3aa314c6c68985

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+    def analyzer_name
+      'fxcop'
+    end
+
+    def analyzer_bin
+      'dotnet'
+    end
+
+    def analyzer_version
+      FXCOP_ANALYZER_VERSION
```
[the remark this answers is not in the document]
gnakagaw: Fix: https://github.com/sider/runners/pull/731/commits/80f49b63d05a0ee65efa2419e34ed9280a29eb38

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+      'fxcop'
+    end
+
+    def analyzer_name
+      'fxcop'
+    end
+
+    def analyzer_bin
+      'dotnet'
+    end
```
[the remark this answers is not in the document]
gnakagaw: FiX: https://github.com/sider/runners/pull/731/commits/85d376c5c5c58c7ec3c9e76f44328f8bf7124b50

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    def analyzer_version
+      FXCOP_ANALYZER_VERSION
+    end
+
+    def setup
+      # restore project (install dependencies)
+      cmdline_install_dependency = %w(dotnet restore)
+      _, _, status = capture3(*cmdline_install_dependency)
```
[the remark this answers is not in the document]
gnakagaw: https://github.com/sider/runners/pull/731/commits/ca7898ddde6b2ec4d0876afb8ff9917f684377cf

lib/runners/processor/fxcop.rb
gnakagaw: Fix: https://github.com/sider/runners/pull/731/commits/9de36274546eb4eabbb0232d0d168e583b3e1c04

lib/runners/processor/fxcop.rb
gnakagaw: Cool. Ruby is so powerful programming language...

Fix: https://github.com/sider/runners/pull/731/commits/1aee8cc11c90889f818b26bbbfa7b529f004c47d

lib/runners/processor/fxcop.rb
gnakagaw: Fix: https://github.com/sider/runners/pull/731/commits/4e073b8ddf33ab84d94d3e92ac2ee94c4e3ddc08

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
+module Runners
+  # A processor class for FxCop Analyzers
+  # https://docs.microsoft.com/en-us/visualstudio/code-quality/install-fxcop-analyzers
+  # This is a beta release. We support only repositories that has .csproj file directly under analysis root directory.
+  class Processor::FxCop < Processor
+    #parameters
+    ANALYSIS_LOGFILE_PATH = '/tmp/sider_analysis_out.json'.freeze
```
ybiquitous: Unused?

```suggestion
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
...
+    def self.ci_config_section_name
+      'fxcop'
+    end
+
+    def analyzer_name
+      'FxCop'
+    end
+
+    def analyzer_version
+      ENV['FXCOP_VERSION']
```
ybiquitous: Safer. If the name is not found, it raises `KeyError`.

```suggestion
      ENV.fetch('FXCOP_VERSION')
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    def setup
+      # restore project (install dependencies)
+      _, _, status = capture3('dotnet', 'restore')
+      unless status.success?
+        msg = <<~EOS
+          Failed to restore .NET Core Project. You have to put project file (.csproj) under the analysis root. Please check the project structure and 'root_dir' parameter in #{ci_config_path_name}.
+          See: https://help.sider.review/getting-started/custom-configuration#root_dir-option
+          Note: We only support a project managed with MSBuild. No support for other build systems. (e.g. Nuke, Cake)
+        EOS
```
ybiquitous: Nice! 👍

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
...
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+        parse_result(output_file.read) do |issue|
+          result.add_issue(issue)
+        end
+      end
+    end
+
+    # parse static analysis log from .NET Core Compilers and generate Issue instance
+    # Output format is SARIF format 1.0
+    # http://json.schemastore.org/sarif-1.0.0
```
ybiquitous: [nitpick]

```suggestion
    # @see http://json.schemastore.org/sarif-1.0.0
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
...
+    def parse_result(f)
+      json = JSON.parse(f, {:symbolize_names => true})
```
ybiquitous: [nitpick]

```suggestion
      json = JSON.parse(f, symbolize_names: true)
```

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
yykamei: I agree with using `ENV.fetch`. `analyzer_version` is not tested, so the value would possibly be accidentally output as `nil` without `#fetch`.

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
...
+            path: relative_path(file_path),
+            location: Location.new(
+              start_line: loc_info[:startLine],
+              start_column: loc_info[:startColumn],
+              end_line: loc_info[:endLine],
+              end_column: loc_info[:endColumn]
+            ),
+            id: result[:ruleId],
+            message: result[:message],
+            links: [rules[result[:ruleId].intern][:helpUri]],
```
ybiquitous: [nitpick] More readable.

```suggestion
            links: [rules[result[:ruleId].to_sym][:helpUri]],
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
...
+        msg = <<~EOS
+          Failed to restore .NET Core Project. You have to put project file (.csproj) under the analysis root. Please check the project structure and 'root_dir' parameter in #{ci_config_path_name}.
+          See: https://help.sider.review/getting-started/custom-configuration#root_dir-option
+          Note: We only support a project managed with MSBuild. No support for other build systems. (e.g. Nuke, Cake)
+        EOS
+        return Results::Failure.new(guid: guid, message: msg, analyzer: analyzer)
+      end
+
+      # install FxCop Analyzers from nuget Repository
+      cmdline_add_analyzer = ['dotnet', 'add', 'package', 'Microsoft.CodeAnalysis.FxCopAnalyzers', '--version', analyzer_version]
```
yykamei: To directly pass arguments to `capture3!` looks shorter.

```suggestion
      capture3!('dotnet', 'add', 'package', 'Microsoft.CodeAnalysis.FxCopAnalyzers', '--version', analyzer_version)
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,98 @@
...
+      capture3!(*cmdline_add_analyzer)
+
+      yield
+    end
+
+    def analyze(changes)
+      # run 'dotnet build' with static analysis module
+      output_file = Tempfile.new("fxcop-")
+      cmdline_run_analyzer = ['dotnet', 'build', '--no-incremental', "-property:errorlog=#{output_file.path}"]
```
yykamei:
```suggestion
      capture3!('dotnet', 'build', '--no-incremental', "-property:errorlog=#{output_file.path}")
```

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,97 @@
...
+        parse_result(output_file.read) do |issue|
+          result.add_issue(issue)
+        end
+      end
+    end
+
+    # parse static analysis log from .NET Core Compilers and generate Issue instance
+    # Output format is SARIF format 1.0
+    # @see http://json.schemastore.org/sarif-1.0.0
+    def parse_result(f)
```
yykamei: I'm not RuboCop, but the argument `f` is too short for this method. `output` seems better.

lib/runners/processor/fxcop.rb
```diff
@@ -0,0 +1,97 @@
...
+      capture3!(*cmdline_add_analyzer)
+
+      yield
+    end
+
+    def analyze(changes)
+      # run 'dotnet build' with static analysis module
+      output_file = Tempfile.new("fxcop-")
+      capture3!('dotnet', 'build', '--no-incremental', "-property:errorlog=#{output_file.path}")
+      capture3!(*cmdline_run_analyzer)
```
yykamei: oops :sweat_smile:

gnakagaw: I fixed it :smile:

ybiquitous:
@yykamei It seems better to squash and merge this PR because of many commits, are you OK?

lib/runners/processor/fxcop.rb
gnakagaw: fixed it

yykamei:
Sorry, I forgot to squash this PR...

ybiquitous:
Don't mind 😃

PR #744, opened 2020-02-17
Show Runners version and GUID to trace

ybiquitous:
This aims to make it easier to know the running version from traces.

### DIFF
--- a/lib/runners/workspace.rb
+++ b/lib/runners/workspace.rb
@@ -126,7 +126,6 @@ module Runners
           yield script_path
         end
       else
-        trace_writer.message "Nothing to do"
         yield nil
       end
     end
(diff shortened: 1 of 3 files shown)


lib/runners/workspace.rb
```diff
@@ -126,7 +126,6 @@ def prepare_ssh
           yield script_path
         end
       else
-        trace_writer.message "Nothing to do"
```
ybiquitous: I think "Nothing to do" is ambigous.

ybiquitous:
> By the way, it's useful for me to include GUID in traces because I can search traces with a specific guid on CloudWatch logs.

That's a nice idea! I'll add it. 💪

ybiquitous:
@yykamei I've done via 68b6876bc5639b63ae705eca70edf30d8d1187af. Is it OK?

PR #749, opened 2020-02-18
Support detekt

### LINKED ISSUE #704: Support detekt
**Is your feature request related to a problem? Please describe.**
Support detekt (https://github.com/arturbosch/detekt)

**Describe the solution you'd like**
implementing a new runner for detekt

**Describe alternatives you've considered**
No alternatives

**Additional context**
Nothing

## Support [detekt]

- Source code URL: https://github.com/arturbosch/detekt
- Official Site/Documentation: https://arturbosch.github.io/detekt/

### Supplementary information

This issue is similar to these completed issues.

- ktlint support
  - https://github.com/sider/runners/pull/192
  - https://help.sider.review/tools/kotlin/ktlint

## TODO

- [ ] Write smoke test(with some linter's Option). Add pattern for success end, waring exit, error exit
- [ ] Add config option
- [ ] (Optional) Add unit tests for the new analysis tool's `Processor` if needed.
- [ ] Write documentation accordingly
- [ ] Add CI setting (https://github.com/sider/runners/blob/master/.circleci/config.yml)
- [ ] Discuss this Lint tool in this Issue
- [ ] Fix developer guide if needed (https://github.com/sider/runners#developer-guide)

hiroko3:
Fix #704 

## TODO
- [x] Add CI configuration
- [x] modify documents (CHANGELOG, README)
- [x] write images/detekt/Dockerfile.erb
- [x] add 'require' declaration to lib/runners.rb
- [x] implement lib/runners/processor/detekt.rb
- [x] implement smoke tests


lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        })
+      end
+
+      let :cli_config, object(
+        auto_correct: boolean?,
+        baseline_path: string?,
+        classpath: array?(string),
+        config: array?(string),
+        config_resource: array?(string),
+        disable_default_rulesets: boolean?,
```
yykamei:
```suggestion
        "disable-default-rulesets": boolean?,
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        excludes: array?(string),
+        includes: array?(string),
+        input: array?(string),
+        language_version: number?,
```
yykamei:
```suggestion
        "language-version": number?,
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        jvm_target: number?,
```
yykamei: :thinking: Our `devon_rex_java` has only [one](https://github.com/sider/devon_rex/blob/9b18513a83ec6835d195c2350287f5072ce457fb/java/Dockerfile#L24) JVM. I think this option won't work.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        plugins: array?(string),
```
yykamei: Can we use `plugins` when using the `cli` option?

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        report_id: string?,
```
yykamei: We should not provide this option because it's difficult to maintain our code like [this](https://github.com/sider/runners/pull/749/files#diff-bbab3614b708c2b1186739c430978219R279). Users don't care about what type of output is generated on `runner_detekt`. You can set an arbitrary `rule-id` to easily parse the output.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        report_path: string?
```
yykamei: You don't have to provide `report_path`.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+      )
+
+      let :gradle_config, object(
+        task: string,
+        report_id: string?,
```
yykamei: Unnecessary

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        report_path: string?
```
yykamei: It's also unnecessary.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+      end
+    end
+
+    def raise_err_unknown_report_id
+      msg = "Unknown 'report type' was supplied. Processing cannot continue."
+      trace_writer.error msg
+      raise msg
+    end
+
+    def gradle_option_quiet
```
yykamei: I think this method does not provide DRY, it's enough to inject `"--quiet"` in each `capture3` or `capture3!`.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
+module Runners
+  class Processor::Detekt < Processor
+    include Java
+
+    Schema = StrongJSON.new do
+
+      def self.array_or?(type)
```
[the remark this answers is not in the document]
hiroko3: I forgot to erase it. Delete it.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+          case value
+          when Array
+            array
+          else
+            type
+          end
+        })
+      end
+
+      def self.array_or(type)
```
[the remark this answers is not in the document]
hiroko3: I forgot to erase it. Delete it.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+
+    def analyzer_version
+      unknown_version = "0.0.0"
+
+      @analyzer_version ||=
+        case
+        when gradle_config
+          extract_detekt_version_for_gradle || unknown_version
+        when maven_config
+          extract_detekt_version_for_maven(current_dir) || unknown_version
```
yykamei: `#current_dir` is defined in `Processor`, so you don't have to pass it as an argument.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        when cli_config
+          pom_dir = Pathname(ENV['RUNNER_USER_HOME']) / analyzer_name
+          extract_detekt_version_for_maven(pom_dir) || unknown_version
```
yykamei: Can we use `extract_version!`? This is defined in `Processor`.

lib/runners/processor/detekt.rb
yykamei: Sorry, I read runner_ktlint again, `report_id` and `report_path` in `gradle_config`  are required because users can configure their `build.gradle` and maybe set the output to an arbitrary path. Such users have to set these parameters in `sider.yml`. So, `switch_constructor` is necessary :pray: 

However, `cli` option still doesn't need `report_id` and `report_path`.

lib/runners/processor/detekt.rb
yykamei: Like `runner_ktlint`, you could remove `#raise_err_unknown_report_id`.

https://github.com/sider/runners/blob/2497fe67704e0bac12873c55e3f32c0489f6b223/lib/runners/processor/ktlint.rb#L6

Also, it might be good to make it required because of the consistency with [runner_ktlint](https://help.sider.review/tools/kotlin/ktlint#ktlint).

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+        args.push("--auto-correct")
+      end
+
+      unless cli_config[:baseline_path].nil?
+        args = check_user_file_exists(
+          [cli_config[:baseline_path]],
+          "baseline",
+          args
+        )
+      end
```
yykamei: It would be better to define methods for each option. For example, runner_eslint provides the `ext` method for the `ext` option.

https://github.com/sider/runners/blob/2497fe67704e0bac12873c55e3f32c0489f6b223/lib/runners/processor/eslint.rb#L110-L113

images/detekt/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+      <version>1.5.1</version>
+    </dependency>
+  </dependencies>
+  <repositories>
+    <repository>
+      <id>jcenter</id>
+      <name>jcenter</name>
+      <url>https://jcenter.bintray.com</url>
+    </repository>
+  </repositories>
```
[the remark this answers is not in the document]
hiroko3: I was very annoyed by this. . .
In detekt, "kotlinx-html" is required for the html report, but it cannot be downloaded without specifying this repository.

lib/runners/processor/detekt.rb
hiroko3: detekt --help says,

>     --auto-correct, -ac
>       Allow rules to auto correct code if they support it. The default rule
>       sets do NOT support auto correcting and won't change any line in the
>       users code base. However custom rules can be written to support auto
>       correcting. The additional 'formatting' rule set, added with
>       '--plugins', does support it and needs this flag.

but we will not support the auto-correction, so this option is also useless. I'll delete it.

lib/runners/processor/detekt.rb
hiroko3: I remembered the reason for this implementation.
Passing the '--version' option to detekt did not produce any results, so it was getting the version specified during installation. (Same as the operation in Maven environment.)
Is this implementation not good?

lib/runners/processor/detekt.rb
hiroko3: Change 'Invalid configuration in `sider.yml`: unexpected value at config: ..' to raise.
By the way, in the case of Maven environment, I read `execution phase`,` report_id`, and `report_path` from 'pom.xml' instead of 'sider.yml'. But, when reading from 'pom.xml', it cannot be detected by the above method. So, also Maven environment, change the implementation that settings are read from 'sider.yml'. (Same as the gradle environment.)

Also change `report_id` and` report_path` to required.
As a result, there has no route to create a report from standard output, so I will remove the implementation that accompanies this route.

images/detekt/pom.xml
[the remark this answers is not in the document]
ybiquitous: I understand it (see https://github.com/arturbosch/detekt/issues/2146). 😓 
Could you please leave a comment like this?

```xml
<!-- NOTE: "kotlinx-html" is not hosted on the Maven central. See https://github.com/arturbosch/detekt/issues/2146 -->
<repositories>
  <repository>
    <id>jcenter</id>
```

lib/runners/processor/detekt.rb
ybiquitous: I think it's no problem. 👍 
`Processor::Ktlint` also do it.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,564 @@
...
+          else
+            type
+          end
+        })
+      end
+
+      let :cli_config, object(
+        auto_correct: boolean?,
+        baseline_path: string?,
+        classpath: array?(string),
```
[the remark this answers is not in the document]
ybiquitous: @hiroko3 ping?

lib/runners/processor/detekt.rb
ybiquitous: Users can specify `--plugins` regardless of `--auto-correct`, can't it?

lib/runners/processor/detekt.rb
yykamei: I got it. It's OK to keep it.

lib/runners/processor/detekt.rb
yykamei: I guess Runners will fail without specified plugins installed if users configure `cli` with `plugins`.

lib/runners/processor/detekt.rb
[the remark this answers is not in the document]
hiroko3: :pray: Sorry about that.
Is there any case to specify the classpath when compiling, such as when using an external jar?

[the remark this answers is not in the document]
yykamei: I'm not sure the behavior of `detekt`, but I think compiling is required for the `classpath` option. However, this `runner_detekt` does not try to compile when the `cli` is configured. Is it correct? If so, the `classpath` of the `cli` option seems unnecessary.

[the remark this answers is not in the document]
ybiquitous: `detekt --help` says:

```
--classpath, -cp
  EXPERIMENTAL: Paths where to find user class files and depending jar
  files. Used for type resolution.
```

This `--classpath` option is experimental yet, so I think we don't need to support it for now.
What do you think? @hiroko3

lib/runners/processor/detekt.rb
ybiquitous: When you will request us to support the `--plugins` option in the future, I think we will need to reconsider this option.

So, I think we don't need it for now. What do you think? @hiroko3

hiroko3: Oh... I understand the point now. Restore the `--plugins` option. (Sorry for my poor understanding of English.)

yykamei: My suggestions are:

* Remove `plugins` option (it's not necessary so far)
* Remove `auto_correct` option

Sorry, this review comment is outdated. I think @hiroko3 doesn't have to do when it comes to plugins or auto-correction.

hiroko3: (I didn't reload the page... :cyclone:)
Rewinds support for the `--plugins` option.

ybiquitous:
@hiroko3 This PR seems ready to review (not WIP) to me. Can we change the PR state from **draft** to **ready to review**?

Or, do you have any to do yet?

lib/runners/processor/detekt.rb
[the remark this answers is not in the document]
hiroko3: @yykamei exactly right...!
@ybiquitous I understood. No objection.

So remove classpath option.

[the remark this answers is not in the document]
ybiquitous: Thanks!

lib/runners/processor/detekt.rb
ybiquitous: Thanks!

hiroko3:
@ybiquitous I've just pushed the removal of the `classpath` and` plugins` options, so ready.

ybiquitous:
@hiroko3 OK, could you please update the PR title and state (Draft)?

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+
+      let :cli_config, object(
+        baseline: string?,
+        config: array?(string),
+        "config-resource": array?(string),
+        "disable-default-rulesets": boolean?,
+        excludes: array?(string),
+        includes: array?(string),
+        input: array?(string),
+        "language-version": number?
```
ybiquitous: [question] [`--language-version`](https://arturbosch.github.io/detekt/cli.html) option is marked as experimental, but is it enough safe to use?

I'm concerned the option will be unsupported... 🤔

ybiquitous: If we can add the option, it looks better to support string or number:

```suggestion
        "language-version": enum?(string, number),
```

```yaml
# OK
language-version: "1.4"
language-version: "1.4.3"

# OK
language-version: 1.4
```

hiroko3:
@ybiquitous 
I tried to add this option, but if the file is less than 200, detekt will not do the analysis. (Same as passing a baseline option in a non-existent file.)
If we support this option, we need to decide whether pass it from number of files to be analyzed.

ybiquitous:
Thanks for your reseach. It looks better not to support the `--parallel` option for now. 👌

lib/runners/processor/detekt.rb
hiroko3: > --language-version option is marked as experimental, but is it enough safe to use?

I read the documentation back, but it doesn't seem to be guaranteed to work.
They says `EXPERIMENTAL`, I think remove it is bring safe...

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+    end
+
+    def cli_input
+      unless cli_config[:input].empty?
+        return check_user_file_exists(
+          cli_config[:input],
+          "input"
+        )
+      end
+    end
```
ybiquitous: [suggestion] It seems unneeded to me to use `check_user_file_exists`, because:

- users may expect that the content of `sider.yml` is converted to the CLI options
  E.g. `input: ["src/", "lib/"]` -> `--input src/,lib/`
- detekt outputs the detail error message like `com.beust.jcommander.ParameterException: Provided path 'lib/' does not exist!`, which users can view the Sider log page

I suggest the simple version as follow:

```ruby
def cli_input
  cli_config[:input].join(",").then { |paths| ["--input", paths] }
end
```

lib/runners/processor/detekt.rb
ybiquitous: OK, let's remove it!

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+        cli_config_files,
+        cli_config_resource,
+        cli_disable_default_rulesets,
+        cli_excludes,
+        cli_input,
+        cli_language_version,
+        cli_report(report_id, report_path)
+      ].flatten.compact
+
+      capture3(analyzer_bin, *args)
```
yykamei: According to the [document](https://github.com/arturbosch/detekt/blob/efa3b4d949859de1d2170d179a8b928e162db0fd/docs/pages/gettingstarted/cli.md), detekt will exit with a specific status code. For example, when "An unexpected error occurred," detekt will exit with `1`. I think we should check whether the exit status code is valid or not here.

* exit `0` --> OK
* exit `1` --> just raise a specific error (this exception will be caught in `Harness`)
* exit `2` --> I don't know... IMO, the same as "exit `1`"
* exit `3` --> This is caused by users' configuration, so returning `Results::Failure` is good

hiroko3: exit `2` is warning for MaxIssues count was reached specified on detekt config `build.maxIssues` (https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/resources/default-detekt-config.yml).
This is a result of analyze runned normally, so treat same as exit `0`.

lib/runners/processor/detekt.rb
hiroko3: OK, raise error.

images/detekt/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>sider</groupId>
+  <artifactId>runners</artifactId>
+  <version>1.0.0</version>
+  <dependencies>
+    <dependency>
+      <groupId>io.gitlab.arturbosch.detekt</groupId>
+      <artifactId>detekt-cli</artifactId>
+      <version>1.5.1</version>
```
ybiquitous: Can you update it to the new version?

```suggestion
      <version>1.6.0</version>
```

- https://search.maven.org/artifact/io.gitlab.arturbosch.detekt/detekt-cli/1.6.0/jar
- https://github.com/arturbosch/detekt/releases/tag/1.6.0

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+      )
+    end
+
+    def cli_baseline
+      unless cli_config[:baseline].nil?
+        return check_user_file_exists(
+          [cli_config[:baseline]],
+          "baseline"
+        )
+      end
```
ybiquitous: [suggestion] We can simplify it like this: ⬇️ 

```suggestion
      cli_config[:baseline].then { |value| value ? ["--baseline", value] : [] }
```

Reason: same as https://github.com/sider/runners/pull/749/files#r384348305

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+      )
+
+      let :cli_config, object(
+        baseline: string?,
+        config: array?(string),
+        "config-resource": array?(string),
+        "disable-default-rulesets": boolean?,
+        excludes: array?(string),
+        includes: array?(string),
+        input: array?(string),
```
ybiquitous: [suggestion] Can we support a string or strings like this?

```suggestion
        config: enum?(string, array(string)),
        "config-resource": enum?(string, array(string)),
        "disable-default-rulesets": boolean?,
        excludes: enum?(string, array(string)),
        includes: enum?(string, array(string)),
        input: enum?(string, array(string)),
```

```yaml
# OK
config: some/config.yml

# OK
config:
  - config1.yml
  - config2.yml
```

To implement this in Ruby, `Array()` is useful:

```ruby
Array(nil)
#=> []
Array("a")
#=> ["a"]
Array(["a"])
#=> ["a"]
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+      end
+    end
+
+    def cli_config_files
+      unless cli_config[:config].empty?
+        return check_user_file_exists(
+          cli_config[:config],
+          "config"
+        )
+      end
```
ybiquitous: [suggestion] Same here: https://github.com/sider/runners/pull/749/files#r384348305

```suggestion
      cli_config[:config].then { |value| value.present? ? ["--config", value.join(",")] : [] }
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+    end
+
+    def cli_config_resource
+      unless cli_config[:"config-resource"].empty?
+        return check_user_file_exists(
+          cli_config[:"config-resource"],
+          "config-resource"
+        )
+      end
```
ybiquitous: [suggestion] Same here: https://github.com/sider/runners/pull/749/files#r384348305

```suggestion
      cli_config[:"config-resource"].then { |value| value.present? ? ["--config-resource", value.join(",")] : [] }
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+      detekt_config[:gradle]
+    end
+
+    def maven_config
+      detekt_config[:maven]
+    end
+
+    def run_cli
+      report_id = "xml"
+      report_path = "detekt-report.xml"
```
ybiquitous: [suggestion] What about using [`Tempfile.new`](https://ruby-doc.org/stdlib-2.7.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-new)? This does not affect on the current directory.

```suggestion
      report_path = Tempfile.new("detekt-report")
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+      
+      args = [
+        cli_baseline,
+        cli_config_files,
+        cli_config_resource,
+        cli_disable_default_rulesets,
+        cli_excludes,
+        cli_input,
+        cli_language_version,
+        cli_report(report_id, report_path)
```
ybiquitous:
```suggestion
        cli_report(report_id, report_path.path)
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,436 @@
...
+
+    def cli_language_version
+      unless cli_config[:"language-version"].nil?
+        return "--language-version", cli_config[:"language-version"].to_s
+      end
+    end
+
+    def cli_report(report_id, report_path)
+      output_file_path = current_dir / report_path
+      return "--report", "#{report_id}:#{output_file_path.to_path}"
```
ybiquitous:
```suggestion
      ["--report", "#{report_id}:#{report_path}"]
```

images/detekt/pom.xml
hiroko3: The format of the HTML report was slightly changed, so handled it.

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,407 @@
...
+      # 2: detekt ran normally and MaxIssues count was reached in BuildFailureReport.
+      # 3: Invalid detekt configuration file detected.
+      if status.exitstatus == 1
+        trace_writer.error stderr.strip
+        raise stderr.strip
+      elsif status.exitstatus == 3
+        return Results::Failure.new(guid: guid, message: stderr.strip, analyzer: analyzer)
+      else
+        return construct_result(report_id, report_path)
+      end
```
ybiquitous: [suggestion] What about this change?

- Handle successful exit codes for Runners (not for detekt).
- Set a Runner's custom message for an invalid config. Because users can confirm `stderr` on their build log page.
- I think we don't need to handle `2`. Because when other detekt exit codes will be added, we should fix this code.

```suggestion
      # detekt has some exit codes.
      # @see https://github.com/arturbosch/detekt/blob/1.6.0/docs/pages/gettingstarted/cli.md
      case status.exitstatus
      when 0, 2
        construct_result(report_id, report_path)
      when 3 # invalid configuration
        Results::Failure.new(guid: guid, message: "Your detekt configuration is invalid", analyzer: analyzer)
      else
        trace_writer.error stderr.strip
        raise stderr
      end      
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,407 @@
+module Runners
+  class Processor::Detekt < Processor
+    include Java
+
+    Schema = StrongJSON.new do
+
+      let :report_ids, enum(
+        literal("xml"),
+        literal("txt"),
+        literal("html")
```
ybiquitous: `html` format is easy to break, so let's remove the support of `html`!

ktlint also doesn't support it:

https://github.com/sider/runners/blob/76b2edee5f731980b4191e45bd286eefedf268f7/lib/runners/processor/ktlint.rb#L6

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,371 @@
...
+      unknown_version = "0.0.0"
+
+      @analyzer_version ||=
+        case
+        when gradle_config
+          extract_detekt_version_for_gradle || unknown_version
+        when maven_config
+          extract_detekt_version_for_maven || unknown_version
+        else
+          pom_dir = Pathname(ENV['RUNNER_USER_HOME']) / analyzer_name
```
ybiquitous:
```suggestion
          pom_dir = Pathname(Dir.home) / analyzer_name
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,371 @@
...
+    def cli_config_files
+      cli_config[:config].then { |value| value.present? ? ["--config", value.join(",")] : [] }      
+    end
+
+    def cli_config_resource
+      cli_config[:"config-resource"].then { |value| value.present? ? ["--config-resource", value.join(",")] : [] }
+    end
+
+    def cli_disable_default_rulesets
+      "--disable-default-rulesets" if cli_config[:"disable-default-rulesets"]
```
ybiquitous: [suggestion] A small refactoring for consistency.

```suggestion
      cli_config[:"disable-default-rulesets"] ? ["--disable-default-rulesets"] : []
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,371 @@
...
+        maven_config[:report_id],
+        maven_config[:report_path]
+      )
+    end
+
+    def construct_result(report_id, report_path)
+      output_file_path = current_dir / report_path
+      if output_file_path.exist?
+        trace_writer.message "Reading output from #{report_path}..."
+        output = output_file_path.read
```
ybiquitous: [suggestion] Can you add `output` to trace for debugging?

```suggestion
        output = output_file_path.read
        trace_writer.message output
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,371 @@
...
+          )
+        end
+      end
+
+      issues
+    end
+
+    def construct_issue(file:, line:, message:, rule:)
+      Issue.new(
+        path: relative_path(working_dir.realpath / file, from: working_dir.realpath),
```
ybiquitous:
```suggestion
        path: relative_path(file),
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,371 @@
...
+        location: Location.new(start_line: line),
+        id: rule,
+        message: message,
+      )
```
ybiquitous: [question] Can you add `severity` to an issue?

FYI.

https://github.com/sider/runners/blob/7318e82a3cab68de619c429caf3eec279d26423b/lib/runners/processor/rubocop.rb#L247

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,371 @@
...
+      end
+
+      issues = []
+
+      document.root.each_element("file") do |file|
+        file.each_element do |error|
+          case error.name
+          when "error"
+            regexp = /.*\.(.*)/
+            rule, = error[:source].match(regexp).captures
```
ybiquitous: [question] Does this code aim to remove the `detekt.` namespace from a rule ID?

detekt has a custom plugin system, so this removing may be unnecessary... 🤔 

For example, if a custom plugin would have a rule named `foo.EmptyClassBlock`, we could not distinguish between `detekt.EmptyClassBlock` and `foo.EmptyClassBlock`?

https://arturbosch.github.io/detekt/extensions.html

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,371 @@
...
+      end
+
+      issues
+    end
+
+    def construct_plain_result(output)
+      issues = []
+
+      regexp_a = /(.*) - .* - .* at (.*):(.*):(.*) - (.*)/
+      regexp_b = /(.*) - .* at (.*):(.*):(.*) - (.*)/
```
ybiquitous: [question] Is this pattern documented somewhere?

lib/runners/processor/detekt.rb
hiroko3: It's OK about case of using `xml` reporter. But case of `txt` reporter does not return 'severity'.

ybiquitous: So can you do the following?

- If `xml`, set `serverity`.
- If `txt`, does not set `severity`.

lib/runners/processor/detekt.rb
hiroko3: Case of using `txt` reporter, they has 2 pattern like this:
```
NestedBlockDepth - 4/4 - [complex] at /tmp/d20200228-6-1exdbtj/src/ComplexClass.kt:9:13 - Signature=ComplexClass.kt$ComplexClass.NestedClass$complex

EmptyClassBlock - [FilteredClass] at /tmp/d20200228-6-1exdbtj/src/FilteredClass.kt:2:21 - Signature=FilteredClass.kt$FilteredClass${ }
```

ybiquitous: Thanks, I suggest:

- leave these example above as a code comment, or
- add a link to the official document describing the pattern

What do you think?

lib/runners/processor/detekt.rb
hiroko3: Even if the same source code is analyzed, the format of the ID differs depending on the reporter.
* `xml`: detekt.EmptyClassBlock
* `txt`: EmptyClassBlock

At first, it stored on Issue as it was, but this measure was taken because there was a suggestion that the ID format should be unified.

Actually namespace is omitted case of using `txt`...  What should I do ...?

ybiquitous: OK, so what about this suggestion?

- If `xml`:
  - `detekt.EmptyClassBlock` is converted to `EmptyClassBlock`
  - `foo.EmptyClassBlock` is not converted, that is `foo.EmptyClassBlock`
- If `txt`:
  - nothing is converted

lib/runners/processor/detekt.rb
ybiquitous: Can you add also `schema: Schema.issue`?

https://github.com/sider/runners/blob/7318e82a3cab68de619c429caf3eec279d26423b/lib/runners/processor/rubocop.rb#L250

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,381 @@
...
+      unless severity.nil?
+        object = { severity: severity }
+      end
+
+      Issue.new(
+        path: relative_path(file),
+        location: Location.new(start_line: line),
+        id: rule,
+        message: message,
+        object: object
```
ybiquitous: The following looks good. `unless severity.nil?` looks not needed.

```suggestion
        object: {
          severity: severity
        },
```

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,381 @@
...
+      end
+
+      issues = []
+
+      document.root.each_element("file") do |file|
+        file.each_element do |error|
+          case error.name
+          when "error"
+            regexp = /detekt\.(.*)/
+            rule, = error[:source].match(regexp).captures
```
ybiquitous: [`delete_prefix`](https://ruby-doc.org/core-2.7.0/String.html#method-i-delete_prefix) looks safe

```suggestion
            rule = error[:source].delete_prefix("detekt.")
```

Regexp is hard to handle...

```ruby
/detekt\.(.*)/.match("foodetekt.abc").captures[0]
#=> "abc"
```

test/smokes/detekt/expectations.rb
```diff
@@ -0,0 +1,363 @@
...
+    guid: "test-guid",
+    timestamp: :_,
+    type: "success",
+    analyzer: { name: "detekt", version: "1.6.0" },
+    issues: [
+      {
+        id: "EmptyClassBlock",
+        path: "src/FilteredClass.kt",
+        location: { start_line: 2 },
+        message: "Signature=FilteredClass.kt$FilteredClass${ }",
```
ybiquitous: [question] Is this message correct? Is this `txt` format?

lib/runners/processor/detekt.rb
```diff
@@ -0,0 +1,381 @@
...
+      end
+
+      issues
+    end
+
+    def construct_plain_result(output)
+      issues = []
+
+      # Case of using txt reporter, they has 2 pattern like this:
+      # a) `NestedBlockDepth - 4/4 - [complex] at /path/to/ComplexClass.kt:9:13 - message...`
```
yykamei: Thank you for writing examples here. I discovered that this `4/4` seems necessary for the rule `NestedBlockDepth` because this rule output the message "Function complex is nested too deeply.". I think `4/4` specifies the level of block depth. As a user, I'm happy to see the information. How about adding it to the suffix of the message like this:

```ruby
# context_info is `"4/4"`
"#{message} [#{context_info}]"
```

I followed [RubCop](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Metrics/CyclomaticComplexity) about the format.

test/smokes/detekt/expectations.rb
hiroko3: That's right.

ybiquitous: Umm..., this message looks unhelpful for users, comparing to such as "The class or object FilteredClass is empty." 🤔

lib/runners/processor/detekt.rb
hiroko3: Sure, but case of using `xml`, has no info about this. So add the `context_info` only about the case of txt reporter using, is it OK?

yykamei: No :smile: 
OK, forget my words. I give up about the context information.

ybiquitous:
We've decided to provide the following constraints with our offline discussion:

1. Use the fixed version, e.g. `1.6.0`.
2. Support only CLI. No Maven or Gradle.
3. Support only XML report format.

Because we predict that we would have a lot of maintenance costs by supporting various formats or versions.

PR #756, opened 2020-02-19
Show stderr of git-blame(1)

yykamei:
We disabled `:trace_stdout` and `:trace_stderr` in case the bunch of traces would output. However, the lack of stderr makes it difficult to look into what is happening if git-blame(1) exits unsuccessfully.

This patch enables `:trace_stderr`, and I believe this does not cause performance issues.


yykamei:
@ybiquitous In case I miss any points, I want to hear your review :pray:

yykamei:
Thank you :running_man:

Issue #750, opened 2020-02-18
Whole `sider.yml` schema check


ybiquitous:
I believe that we need a schema check for the whole `sider.yml`.

Currently, each `linters.*` schema check is just supported. This cannot check the following `sider.yml`:

```yaml
linters:
  rubocop:
    gems:
      - rubocop-rails
  unknown: |
    What is this?
```

Next actions:

* [x] Pass `config` to `Processor#initialize` from `Harness`
* [x] Use `config` in brakeman
* [x] Use `config` in checkstyle
* [x] Use `config` in code_sniffer
* [x] Use `config` in coffeelint
* [x] Use `config` in cppcheck
* [x] Use `config` in cpplint
* [x] Use `config` in eslint
* [x] Use `config` in flake8
* [x] Use `config` in fxcop
* [x] Use `config` in go_vet
* [x] Use `config` in golangci_lint
* [x] Use `config` in golint
* [x] Use `config` in gometalinter
* [x] Use `config` in goodcheck
* [x] Use `config` in hadolint
* [x] Use `config` in haml_lint
* [x] Use `config` in javasee
* [x] Use `config` in jshint
* [x] Use `config` in ktlint
* [x] Use `config` in misspell
* [x] Use `config` in phinder
* [x] Use `config` in phpmd
* [x] Use `config` in pmd_java
* [x] Use `config` in querly
* [x] Use `config` in rails_best_practices
* [x] Use `config` in reek
* [x] Use `config` in rubocop
* [x] Use `config` in scss_lint
* [x] Use `config` in shellcheck
* [x] Use `config` in stylelint
* [x] Use `config` in swiftlint
* [x] Use `config` in tslint
* [x] Use `config` in tyscan

yykamei:
I need this issue to be resolved because of #760. What plan do you have?
If you don't get into this issue, I want to implement it.

My plan is to create `Config` class in `config.rb` and instantiate it in `Harness`.  Then, I want to pass the instance to a processor.

yykamei:
The reason I will instantiate `Config` in `Harness` is that `Harness` can raise `Results::Failure`. If `sider.yml` is broken, we should return `Failure`.

ybiquitous:
I've no plan now. Let's go as you want! 💪

yykamei:
I edited the description for the next steps.

yykamei:
I believe we can close this issue via #763, #777, #779, and #782.

PR #763, opened 2020-02-21
Add Runners::Config, which is responsible for sider.yml

yykamei:
This PR introduces `Runners::Config` which is responsible for our configuration file `sider.yml`. Currently, this file is parsed in each processor, but as it's told in #750, unknown keys or other top-level attributes are not validated. This commit just adds `Runners::Config`, so I will make Runners use the class in the subsequent commit.


lib/runners.rb
```diff
@@ -69,42 +107,8 @@
...
-require "runners/processor/scss_lint"
-require "runners/processor/shellcheck"
-require "runners/processor/stylelint"
-require "runners/processor/swiftlint"
-require "runners/processor/tslint"
-require "runners/processor/tyscan"
-
-module Runners
-end
+Runners::RUNNER_IDS.each { |id| require "runners/processor/#{id}" }
```
yykamei: I know this is a little strange, but it's required to use `RUNNER_IDS` [here](https://github.com/sider/runners/pull/763/files#diff-c8f2460235628acd1ec6059742358868R29).

It's not necessary if we set the `linter` field like this, but the purpose of #750 includes to check _unknown_ attributes.

```ruby
      let :payload, object(
        linter: any?,
        ignore: enum?(string, array(string)),
        branches: object?(exclude: enum?(string, array(string))),
      )
```

@ybiquitous What do you think? I want to hear your opinion :pray:

ybiquitous: Honestly, I feel it too tricky to introduce `RUNNER_IDS`... 😖 

I prefer to write it statically and explicitly. Of course, it's not DRY, but writing statically is clear and easier to debug.
If forgetting to add a new ID, a smoke test fails.

Why using `any`?

```diff
-linter: any?,
+linter: object?(rubocop, eslint, ...),
```

yykamei: Thank you. I actually wanted to write them statically :smile: I will follow your suggestion.

yykamei:
I will extract all schemas from each runner like this:

lib/runners/schema/config.rb:

```ruby
      let :brakeman, optional(ruby)
      let :checkstyle, optional(base.update_fields { |fields|
        fields[:config] = string?
        fields[:dir] = optional(enum(string, array(string)))
        fields[:exclude] = enum?(
          string,
          array(
            enum(
              string,
              object(pattern: string),
              object(string: string)
            )
          )
        )
        fields[:ignore] = array?(string)
        fields[:properties] = string?
      })
```

lib/runners/schema/config.rb
```diff
@@ -0,0 +1,431 @@
...
+              paths: array?(string),
+            )
+          }),
+        ),
+        ignore: enum?(string, array(string)),
+        branches: object?(exclude: enum?(string, array(string))),
+      )
+    end
+  end
+end
```
yykamei: From this patch, developers have to update this file to configure the schema of each runner. What do you think? @ybiquitous

ybiquitous: It seems a bit too long. 🤔  Can we import each schema from each processor class?

yykamei: OK, how about this?

lib/runners/processor/cppcheck.rb:
```ruby
let :runner_config, Schema::BaseConfig.base.update_fields { |fields|
  fields.merge!(
    target: enum?(string, array(string)),
    ignore: enum?(string, array(string)),
    enable: string?,
    std: string?,
    project: string?,
    language: string?,
  )
}
```


lib/runners/schema/config.rb:
```ruby
let :payload, object(
  linter: object?(
    ...,
    cppcheck: optional(Processor::Cppcheck::Schema.runner_config),
    ...,
  ),
)
```

ybiquitous: Looks good! 👍

yykamei: :thinking: In order to refer each processor in `lib/runners/schema/config.rb`, the file should be loaded after all processors in `lib/runners.rb`. However, I want to make each runner inherit the base configuration, which includes `root_dir`, so I have to separate the base configuration schema like `lib/runners/schema/base_config.rb`.  This top-level module is good? I think the code below lacks the code consistency.

`/lib/runners.rb`:
```ruby
require "runners/schema/base_config"
require "runners/schema/result"

require "runners/processor"
...
require "runners/processor/cppcheck"
...

require "runners/config"
require "runners/schema/config"
```

ybiquitous: Surely, it looks no good... 🤔 

To handle "too long" in one file, what about split to multiple files like this?


```ruby
require "runners/schema/config/cppcheck"
require "runners/schema/config/rubocop"
# ...
```

yykamei: In this case, where should I place the base configuration? I think `lib/runners/schema/config/cppcheck.rb` is necessary to split into discrete module.

ybiquitous: For example, what about like this? ⬇️ 

```ruby
require "runners/schema/config"
require "runners/schema/config/cppcheck"
require "runners/schema/config/rubocop"
# ...
```

```ruby
# runners/schema/config.rb
module Runners
  module Schema
    Config = StrongJSON.new do
      # top-level definitions...
    end
  end
end
```

```ruby
# runners/schema/config/cppcheck.rb
module Runners
  module Schema
    module Config
      Cppcheck = StrongJSON.new do
        # ...
      end
    end
  end
end
```

yykamei: I think it won't work :man_shrugging: because I want to refer `Cppcheck` from `Config`.

```ruby
# lib/runners/schema/config.rb
module Runners
  module Schema
    Config = StrongJSON.new do
      let :payload, object(
        cppcheck: optional(Runners::Schema::Config::Cppcheck.config),
      )
    end
  end
end
```

```
/home/kamei/git/Sider/runners/lib/runners/schema/config.rb:5:in `block in <module:Schema>': uninitialized constant Runners::Schema::Config (NameError)
Did you mean?  Runners::Config
               RbConfig
```

ybiquitous: Umm..., then what about this?

```ruby
# runners.rb
require "runners/schema/config"
```

```ruby
# runners/schema/config.rb
require_relatvie "./config/cppcheck"

module Runners
  module Schema
    Config = StrongJSON.new do
      # top-level definitions...
    end
  end
end
```

yykamei: It still doesn't seem to work because `Cppcheck` requires `Config.base`. Yes, they are referring to each other. That's why I said `base_config.rb` would be required.

```ruby
# runners/schema/config/cppcheck.rb
module Runners
  module Schema
    module Config
      Cppcheck = StrongJSON.new do
        let :config, Config.base.update_fields { |fields| fields.merge!(...) }
      end
    end
  end
end
```

yykamei: I'm considering another approach like this. This `Config` is declared, and then each processor registers its schema with `Config.register`. It might be tricky but allows us to separate schema declarations. What do you think?

```ruby
# lib/runners/schema/config.rb
    Config = StrongJSON.new do
      let :payload, object(
        linter: StrongJSON::Type::Optional.new(StrongJSON::Type::Object.new({}, on_unknown: :reject, exceptions: Set.new)),
        ignore: enum?(string, array(string)),
        branches: object?(exclude: enum?(string, array(string))),
      )

      def register(name, schema: nil, inherit: :base)
        runner_config =
          case inherit
          when :base
            BaseConfig.base.update_fields { |fields| fields.merge!(schema.fields) if schema }
          when :ruby
            BaseConfig.ruby.update_fields { |fields| fields.merge!(schema.fields) if schema }
          when :npm
            BaseConfig.npm.update_fields { |fields| fields.merge!(schema.fields) if schema }
          else
            raise ArgumentError, "#{inherit} is not supported"
          end
        linter_config = payload.fields[:linter].type
        payload.fields[:linter] = optional(linter_config.update_fields { |fields| fields[name] = optional(runner_config) })
      end
    end
```

```ruby
# lib/runners/processor/cpplint.rb
    Runners::Schema::Config.register(:cpplint, schema: Schema.runner_config)
```

ybiquitous: What about this my patch? It seems we need `merge` instead of `update_fiels` for `StrongJSON`.

```diff
diff --git a/lib/runners.rb b/lib/runners.rb
index d728e573..35418738 100644
--- a/lib/runners.rb
+++ b/lib/runners.rb
@@ -11,6 +11,7 @@ require "net/http"
 require "zlib"
 require "time"
 require "strong_json"
+require_relative "./strong_json/type/object"
 require "yaml"
 require "jsonseq"
 require "fileutils"
diff --git a/lib/runners/schema/config.rb b/lib/runners/schema/config.rb
index 794e4e14..dfdca046 100644
--- a/lib/runners/schema/config.rb
+++ b/lib/runners/schema/config.rb
@@ -1,3 +1,5 @@
+require_relative "./config_checkstyle"
+
 module Runners
   module Schema
     BaseConfig = _ = StrongJSON.new do
@@ -69,24 +71,7 @@ module Runners
       let :payload, object(
         linter: object?(
           brakeman: optional(BaseConfig.ruby),
-          checkstyle: optional(BaseConfig.base.update_fields { |fields|
-            fields.merge!(
-              config: string?,
-              dir: enum?(string, array(string)),
-              exclude: enum?(
-                string,
-                array(
-                  enum(
-                    string,
-                    object(pattern: string),
-                    object(string: string)
-                  )
-                )
-              ),
-              ignore: array?(string),
-              properties: string?,
-            )
-          }),
+          checkstyle: optional(BaseConfig.base.merge(ConfigCheckstyle.payload)),
           code_sniffer: optional(BaseConfig.base.update_fields { |fields|
             fields.merge!(
               version: enum?(string, numeric),
diff --git a/lib/runners/schema/config_checkstyle.rb b/lib/runners/schema/config_checkstyle.rb
new file mode 100644
index 00000000..c53c1d3b
--- /dev/null
+++ b/lib/runners/schema/config_checkstyle.rb
@@ -0,0 +1,22 @@
+module Runners
+  module Schema
+    ConfigCheckstyle = StrongJSON.new do
+      let :payload, object(
+        config: string?,
+        dir: enum?(string, array(string)),
+        exclude: enum?(
+          string,
+          array(
+            enum(
+              string,
+              object(pattern: string),
+              object(string: string),
+            )
+          )
+        ),
+        ignore: array?(string),
+        properties: string?,
+      )
+    end
+  end
+end
diff --git a/lib/strong_json/type/object.rb b/lib/strong_json/type/object.rb
new file mode 100644
index 00000000..35ebd2bc
--- /dev/null
+++ b/lib/strong_json/type/object.rb
@@ -0,0 +1,10 @@
+# NOTE: This is a monkey-patch!
+class StrongJSON
+  module Type
+    class Object
+      def merge(other_object_type)
+        self.class.new(fields.merge(other_object_type.fields), on_unknown: on_unknown, exceptions: exceptions)
+      end
+    end
+  end
+end
```

yykamei: It's good, but is it annoying that we have to update both of `lib/runners/schema/config.rb` and `lib/runners/schema/config_checkstyle.rb` when adding a new runner?

ybiquitous: Yes, I think so, but I got the following warnings: 😓 

```
$ bundle exec irb -Ilib -rrunners
/Users/koba/git/sider/runners/lib/runners/schema/config.rb:68: warning: already initialized constant Runners::Schema::Config
/Users/koba/git/sider/runners/lib/runners/schema/config/checkstyle.rb:3: warning: previous definition of Config was here
irb(main):001:0>
```

Do you have any good ideas?

yykamei: What about [this](https://github.com/sider/runners/pull/763#discussion_r384231494)? This does not declare the schema when the code is loaded, but it would be easy to add a new runner.

ybiquitous: Then, I suggest the following improvement, but the basic idea looks good!

I prefer an API like this: ⬇️ 

```ruby
module Runners
  class Processor
    def self.register_config_schema(**args)
      Schema::Config.register(**args)
    end
  end
end

module Runners
  class Processor::Checkstyle < Processor
    register_config_schema :checkstyle, Schema.config
    # or
    register_config_schema :checkstyle, Schema.config, inherit: :base
  end
end
```

yykamei: :clap: Thank you so much :pray: I will adopt your suggestion.

yykamei:
This PR only adds `Runners::Config` and `Runners::Schema::Config`. I want to use them in other PRs.

lib/runners/config.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def ignore
+      Array(content&.fetch(:ignore))
+    end
+
+    private
+
+    def path
+      return @path if defined?(@path)
+      @path = [working_dir / "sider.yml", working_dir / "sideci.yml"].find do |pathname|
```
ybiquitous: [suggestion] I feel it hard to maintain our cod if we write `sider.yml` or `sider.yml` here and there.

https://github.com/sider/runners/blob/acf237fe45d3cdf241c76c69ee36d1fd205ff0ec/lib/runners/processor.rb#L93-L108

What about passing the config file path instead of `working_dir` to `Runners::Config.new`?

For example:

```ruby
Runners::Config.new(ci_config_path)
```

lib/runners/config.rb
```diff
@@ -0,0 +1,58 @@
...
+        pathname.file?
+      end
+    end
+
+    def compact(val)
```
ybiquitous: [question] Could you tell me why deeply compact?

lib/runners/config.rb
yykamei: I'm planning to use `Runners::Config` in `Harness`, so I want `Runners::Config` to determine what configuration file should be used and would deprecate/remove `Processor#ci_config_path` later. It might be good to comment with `DEPRECATED` :smiley:

lib/runners/schema/config.rb
```diff
@@ -0,0 +1,50 @@
...
+      let :npm, (base.update_fields { |fields|
+        fields.merge!(npm_install: optional(npm_install))
+      })
+    end
+
+    Config = _ = StrongJSON.new do
+      # @type self: Types::Config
+
+      let :payload, object(
+        linter: StrongJSON::Type::Optional.new(StrongJSON::Type::Object.new({}, on_unknown: :reject, exceptions: Set.new)),
```
ybiquitous: [question] Could you tell me why using `StrongJSON::Type::Optional.new` instead of the native API `object`?

lib/runners/config.rb
ybiquitous: I see. Then, how will each processor access to `sider.yml` (`#ci_config_path`)?

lib/runners/config.rb
yykamei: Without `compact`, the object of `StrongJSON::Type::Object#coerce` would be like this. I think it is hard for users to read it on traces.

```json
{
  "linter": {
    "brakeman": null,
    "checkstyle": null,
    "code_sniffer": null,
    "coffeelint": null,
    "cppcheck": null,
    "cpplint": null,
    "eslint": {
      "root_dir": null,
      "npm_install": null,
      "dir": null,
      "ext": null,
      "config": "abc",
      "ignore-path": null,
      "ignore-pattern": null,
      "no-ignore": null,
      "global": null,
      "quiet": null,
      "options": {
        "npm_install": true,
        "dir": null,
        "ext": null,
        "config": null,
        "ignore-path": null,
        "no-ignore": null,
        "ignore-pattern": null,
        "global": null,
        "quiet": null
      }
    },
    "flake8": null,
    "fxcop": null,
    "go_vet": null,
    "golangci_lint": null,
    "golint": null,
    "gometalinter": null,
    "goodcheck": null,
    "hadolint": null,
    "rubocop": null,
    "haml_lint": null,
    "javasee": null,
    "jshint": null,
    "ktlint": null,
    "misspell": null,
    "phinder": null,
    "phpmd": null,
    "pmd_java": null,
    "querly": null,
    "rails_best_practices": null,
    "reek": null,
    "scss_lint": null,
    "shellcheck": null,
    "stylelint": null,
    "swiftlint": null,
    "tslint": null,
    "tyscan": null
  },
  "ignore": null,
  "branches": null
}
```

lib/runners/config.rb
yykamei: Yes, good point. I want to add a new parameter to `Processor#initialize` as `:config`.

ybiquitous: OK, make sense. 👍

lib/runners/schema/config.rb
yykamei: Nice catch :baseball: `object` without any `fields` assumes the parameter `on_unknown` as `:ignore`, but I want to make `on_unknown` as `:reject`.

https://github.com/soutaro/strong_json/blob/1439e6ad521c9d28dd1c06f5e11d68635b682dc0/lib/strong_json/types.rb#L5-L9

lib/runners/config.rb
ybiquitous: I feel it hard to maintain and read, because this `compact` depends on the StrongJSON definition for `Config.payload[:linters]`.

Is there any good way...? 🤔

lib/runners/schema/config.rb
ybiquitous: I see. 👍 

Ideally, do we want the following code? But it's out of the scope of this PR. 😅 

```ruby
def object(fields = {}, on_unknown: nil)
  if fields.empty?
    Type::Object.new(fields, on_unknown: on_unknown || :ignore, exceptions: Set.new)
  else
    Type::Object.new(fields, on_unknown: on_unknown || :reject, exceptions: Set.new)
  end
end
```

yykamei: Yes, I agree. :clap:

lib/runners/config.rb
yykamei: The value of `StrongJSON::Type::Object#coerce` is just a Hash instance :thinking: 
but I think just writing `coerce`d value seems OK.

lib/runners/schema/config.rb
ybiquitous: But sadly this code does not work... 😢 

```ruby
StrongJSON.new { object(id: number) }
#=> ArgumentError (unknown keyword: id)
```

ybiquitous: I find a refactoring idea. What about this?

```ruby
Config = _ = StrongJSON.new do
  # @type self: Types::Config

  @linter = {}

  let :payload, object(
    linter: optional(object(@linter)),
    ignore: enum?(string, array(string)),
    branches: object?(exclude: enum?(string, array(string))),
  )

  def register(name:, schema:)
    if @linter.include? name
      raise ArgumentError, "#{name.inspect} is already registered"
    end

    @linter[name] = optional(schema)
    payload.fields[:linter] = optional(object(@linter))
  end
end
```

yykamei: Pretty good. I want to try it.

lib/runners/config.rb
ybiquitous: > I think it is hard for users to read it on traces.

I think we need to show a `coerce`d value to users. Instead, we need to show a raw content string of `sider.yml` to users?

For example, I suggest a solution as follow:

```ruby
class InvalidConfiguration < UserError
  attr_reader :input_string

  def initialize(message, input_string)
    super(message)
    @input_string = input_string
  end
end

def initialize(working_dir)
  @working_dir = working_dir
  @input_string = path.read if path
  @content = check_schema(parse_yaml)
end

def parse_yaml
  YAML.safe_load(@input_string, symbolize_names: true)
rescue Psych::SyntaxError => exn
  message = "Your `#{path_name}` file may be broken (line #{exn.line}, column #{exn.column})."
  raise BrokenYAML.new(message)
end

def check_schema(object)
  Schema::Config.payload.coerce(object)
rescue StrongJSON::Type::UnexpectedAttributeError => exn
  message = "The attribute `#{exn.path}.#{exn.attribute}` of `#{path_name}` cannot exist."
  raise InvalidConfiguration.new(message, @input_string)
rescue StrongJSON::Type::TypeError => exn
  message = "The value of the attribute `#{exn.path}` is invalid."
  raise InvalidConfiguration.new(message, @input_string)
end
```

We can show `InvalidConfiguration#input_string` to users via traces.

lib/runners/schema/config.rb
yykamei: It looks good. Thank you.

f4266a1

lib/runners/config.rb
ybiquitous:
```ruby
@input_string = path.read if path
input_object = parse_yaml
@content = check_schema(input_object) if input_object

Comment (yykamei) [reply lib/runners/config.rb]: > I think we need to show a coerced value to users. 

Does it means "we don't need to show a `coerce`d value to users"?

Comment (yykamei) [reply lib/runners/config.rb]: Anyway, I will follow your implementation.

Comment (ybiquitous) [reply lib/runners/config.rb]: > Does it means "we don't need to show a `coerce`d value to users"?

Yes, sorry I missed "don't"... 😅

Comment (yykamei) [reply lib/runners/config.rb]: Done with fed2a17 and 1e2c142

Comment (ybiquitous) [reply lib/runners/config.rb]: Yep!

File: lib/runners/config.rb [hunk 6]
```diff
@@ -0,0 +1,74 @@
...
+      case val
+      when Hash
+        val.compact.each_with_object({}) do |(k, v), h|
+          h[k] = compact(v)
+        end
+      else
+        val
+      end
+    end
```
ybiquitous: `compact` is not used?

```suggestion
```

yykamei: Oh, sorry, I forgot to remove it.

yykamei: "Multi-line suggestions" is useful.

test/config_test.rb
```diff
@@ -0,0 +1,196 @@
...
+    mktmpdir do |path|
+      (path / "sider.yml").write(<<~YAML)
+        ---
+        linter:
+          unknown_linter:
+            config: abc
+      YAML
+      assert_raises Runners::Config::InvalidConfiguration do
+        Runners::Config.new(path)
+      end
```
ybiquitous: [suggestion] What about testing also an error message? Because the message depends on StrongJSON methods which may be changed in the future.

```suggestion
      exn = assert_raises Runners::Config::InvalidConfiguration do
        Runners::Config.new(path)
      end
      assert_equal "The attribute `$.linter.unknown_linter` of `sider.yml` cannot exist.", exn.message
```

test/config_test.rb
```diff
@@ -0,0 +1,196 @@
...
+    end
+
+    mktmpdir do |path|
+      (path / "sider.yml").write(<<~YAML)
+        ---
+        linter: []
+      YAML
+      assert_raises Runners::Config::InvalidConfiguration do
+        Runners::Config.new(path)
+      end
```
ybiquitous:
```suggestion
      exn = assert_raises Runners::Config::InvalidConfiguration do
        Runners::Config.new(path)
      end
      assert_equal "The value of the attribute `$.linter` is invalid.", exn.message
```

test/config_test.rb
```diff
@@ -0,0 +1,196 @@
...
+  end
+
+  def test_content_with_broken_yaml
+    mktmpdir do |path|
+      (path / "sider.yml").write(<<~YAML)
+        @
+      YAML
+      assert_raises Runners::Config::BrokenYAML do
+        Runners::Config.new(path)
+      end
```
ybiquitous:
```suggestion
      exn = assert_raises Runners::Config::BrokenYAML do
        Runners::Config.new(path)
      end
      assert_equal "Your `sider.yml` file may be broken (line 1, column 1).", exn.message
```

test/config_test.rb
yykamei: OK, looks good.

sig/runners/config.rbi
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader input_string: String?
+  attr_reader content: Hash<Symbol, any>
+
+  def initialize: (Pathname) -> any
+  def path_name: -> String
+  def ignore: -> Array<String>
+  def path: -> Pathname?
+  def parse_yaml: () -> String?
+  def check_schema: (String?) -> Hash<Symbol, any>
+  def compact: (any) -> any
```
ybiquitous: 😅 

```suggestion
```

ybiquitous:
@yykamei Can you use **Add suggestion to batch**?

image

test/config_test.rb
yykamei: I added the check of `#input_string` with 6c2fc7f.

yykamei:
> Can you use **Add suggestion to batch**?

I checked your suggestions one by one, so I didn't use the batch.

lib/runners/config.rb
```diff
@@ -0,0 +1,64 @@
...
+      @content = check_schema(parse_yaml)
+    end
+
+    def path_name
+      f = path # XXX: This assignment is for typecheck
+      if f
+        f.basename.to_s
+      else
+        "sider.yml"
+      end
```
ybiquitous: What about this?

```suggestion
      path&.basename&.to_s || "sider.yml"
```

lib/runners/config.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    def ignore
+      Array(content&.fetch(:ignore))
+    end
+
+    private
+
+    def path
+      return @path if defined?(@path)
+      @path = [working_dir / "sider.yml", working_dir / "sideci.yml"].find do |pathname|
```
ybiquitous: [suggestion] What about making constants like this to clarify the intention?

```rubn
CONFIG_FILENAME = "sider.yml"
CONFIG_FILENAME_OLD = "sideci.yml"
```

ybiquitous:
> I checked your suggestions one by one, so I didn't use the batch.

The multiple email notification is noisy to me... 😢

yykamei:
In the first place, this button is disabled for some reason.

ybiquitous:
This reason?

image

yykamei:
Actually, I don't know the feature.

sig/runners/config.rbi
yykamei: Done with a572497

yykamei:
Thank you for your time

ybiquitous:
BTW, I do want to change `ci_config_section_name` method name because "Why ci?". 😓 

Do you have any good idea?

yykamei:
> BTW, I do want to change `ci_config_section_name` method name because "Why ci?". 
> 
> Do you have any good idea?

I agree, `ci` should not be used.
I think the method `ci_config_section_name` itself is not necessary because it's enough to refer a runner name. If I must say, `runner_config_name`?

ybiquitous:
> I think the method `ci_config_section_name` itself is not necessary because it's enough to refer a runner name.

Agreed 👍

yykamei:
I have a plan to rename `ci_config_section_name` with `analyzer_id`.

The current code does not have analyzer IDs, such as `golangci_lint`, so renaming seems better.
Also, I can use `analyzer_id` in the file below:

https://github.com/sider/runners/blob/09f9e08516d053b006eff0cb6e3581810a1696eb/lib/runners/cli.rb#L34-L38

This is the newer version that uses `analyzer_id`.

```ruby
def processor_class
  @processor_class ||= (Processor.subclasses.detect do |cls|
    analyzer == cls.analyzer_id
  end or raise "Not found processor class with '#{analyzer}'")
end
```

ybiquitous:
The idea sounds good!

PR #769, opened 2020-02-26
[RuboCop] Use `--out=FILE` option

ybiquitous:
Handling STDOUT is unsafe due to bugs of RuboCop itself or its plugins.

The [RuboCop document](https://docs.rubocop.org/en/latest/basic_usage/#command-line-flags) says:

> `-o/--out`    Write output to a file instead of STDOUT.


ybiquitous:
NOTE: This change is for internal, so I don't add it to changelog.

ybiquitous:
@yykamei Good catch! I will leave it as a code comment. 👍

PR #780, opened 2020-02-27
[hadolint] Exclude template files by default

ybiquitous:
By default, the hadolint runner tries to analyze files to match the `**/Dockerfile{,.*}` glob pattern.
For example, the pattern matches `Dockerfile.erb` which is used as a template file such as in this *sider/runners* repository.

This change excludes `*.erb` and `*.txt` files by default.
Because such files are a template file and are not a real Dockerfile.

*NOTE: We may add some file extensions to be excluded in the future.*


ybiquitous:
@yykamei What do you think about this change? Are there some pitfalls?

ybiquitous:
Thank you for the review! 👏

PR #783, opened 2020-02-28
Tool deprecation warnings

### LINKED ISSUE #709: Deprecate TSLint
TSLint has been marked as deprecated since the version 6.0.0:

https://github.com/palantir/tslint/releases/tag/6.0.0

See also <https://github.com/palantir/tslint/issues/4534>.

I don't think we should drop support of TSLint soon, but we may need to mark it as deprecated (including the document) at least for the future smooth migration.

ybiquitous:
- Add `:ref` option (an useful URL) to `#add_warning_for_deprecated_linter` method.
- Add a deprecation warning for TSLint.

Fix #709

### DIFF
--- a/lib/runners/processor/gometalinter.rb
+++ b/lib/runners/processor/gometalinter.rb
@@ -44,6 +44,10 @@ module Runners
     end
 
     def setup
+      add_warning_for_deprecated_linter(alternative: "GolangCi-Lint",
+                                        ref: "https://github.com/alecthomas/gometalinter/issues/590",
+                                        deadline: Time.new(2020, 3, 31))
+
       with_import_path do |path|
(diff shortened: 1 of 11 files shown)


lib/runners/processor/gometalinter.rb
```diff
@@ -44,6 +44,10 @@ def analyzer_name
     end
 
     def setup
+      add_warning_for_deprecated_linter(alternative: "GolangCi-Lint",
+                                        ref: "https://github.com/alecthomas/gometalinter/issues/590",
+                                        deadline: Time.new(2020, 3, 31))
+
```
ybiquitous: Move first than other warnings.

PR #785, opened 2020-02-28
Refacotr Processor

yykamei:
This is just a refactoring. These changes are included.

* Remove `ci_section_root_dir`
* Rename `build_field_reference_from_path` to `config_field_path`
* Always run `push_dir` in `push_root_dir`
* Always check the directory existence in `check_root_dir_exist`
  even though the `root_dir` is not specified

### DIFF
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -85,38 +85,20 @@ module Runners
       config.content.dig(:linter, self.class.ci_config_section_name.to_sym) || {}
     end
 
-    def ci_section_root_dir
-      ci_section[:root_dir]
-    end
-
     def check_root_dir_exist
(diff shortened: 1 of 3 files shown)


lib/runners/processor.rb
```diff
@@ -85,38 +85,20 @@ def ci_section
...
     end
 
     def push_root_dir(&block)
-      root_dir = ci_section_root_dir
-      if root_dir
-        push_dir(working_dir + root_dir, &block)
-      else
-        yield
-      end
+      push_dir(root_dir, &block)
```
yykamei: I believe it's safe to always run `push_dir` even though the `root_dir` is the same as `working_dir`.

PR #787, opened 2020-02-28
Support LanguageTool

### LINKED ISSUE #725: Support LanguageTool
## Information

- Website: https://languagetool.org/
- Source code: https://github.com/languagetool-org/languagetool
- Language: Natural languages. E.g. English, French, German, Polish, Russian, etc.

## Why do you need the tool?

I think that checking style and grammar is useful for most users.

## Additional context

Written in Java.

## See also

None.

mpls104:
Fix #725

- [x] Write a [document](https://github.com/sider/sider-docs) for the new runner. PR sider/sider-docs#357


ybiquitous:
@mpls104 We've migrated CircleCI to GitHub Actions and introduced [`analyzers.yml`](https://github.com/sider/runners/blob/master/analyzers.yml).

Please rebase the master branch and fix them (I updated this PR description). 🙏

mpls104:
# Language Tool

## Done
- Created Dockerfile to run smoke test so far.
- Set default language to `en-US` since without specifying country, detection for misspelling will not work.

## Remains problems
- When specifying multiple files, language tool returns invalid json.
- Somehow, target file is emiited on stderr, so it's needed to extracted.
- Recursive option can only be specified when targe is directory.
- 

## wiki
[Tips and Tricks - LanguageTool Wiki](http://wiki.languagetool.org/tips-and-tricks)

## Not yet
- [ ] Tackle remains problems
- [ ] Refactor Dockerfile
- [ ] Add required option
- [ ] Add test pattern
- [ ] 


## Options
```
    " Available options:\n" +
    "  -r, --recursive          work recursively on directory, not on a single file\n" +
    "  -c, --encoding ENC       character set of the input text, e.g. utf-8 or latin1\n" +
    "  -b                       assume that a single line break marks the end of a paragraph\n" +
    "  -l, --language LANG      the language code of the text, e.g. en for English, en-GB for British English\n" +
    "  --list                   print all available languages and exit\n" +
    "  -adl, --autoDetect       auto-detect the language of the input text - note this will not detect\n" +
    "                           variants like 'English (US)', so you will not get spell checking for\n" +
    "                           languages with variants\n" +
    "  -m, --mothertongue LANG  the language code of your first language, used to activate false-friend checking\n" +
    "  -d, --disable RULES      a comma-separated list of rule ids to be disabled (use no spaces between ids)\n" +
    "  -e, --enable RULES       a comma-separated list of rule ids to be enabled (use no spaces between ids)\n" +
    "  -eo, --enabledonly       disable all rules except those enabled explicitly in -e\n" +
    "  --enablecategories CAT   a comma-separated list of category ids to be enabled (use no spaces between ids)\n" +
    "  --disablecategories CAT  a comma-separated list of category ids to be disabled (use no spaces between ids)\n" +
    "  -t, --taggeronly         don't check, but only print text analysis (sentences, part-of-speech tags)\n" +
    "  -u, --list-unknown       also print a summary of words from the input that LanguageTool doesn't know\n" +
    "  -b2, --bitext            check bilingual texts with a tab-separated input file,\n" +
    "                           see http://languagetool.wikidot.com/checking-translations-bilingual-texts\n" +
    "  --api                    print results as XML (deprecated, please use --json or the JSON API in server mode)\n" +
    "  --json                   print results as JSON\n" +
    "  -p, --profile            print performance measurements\n" +
    "  -v, --verbose            print text analysis (sentences, part-of-speech tags) to STDERR\n" +
    "  --version                print LanguageTool version number and exit\n" +
    "  -a, --apply              automatically apply suggestions if available, printing result to STDOUT\n" +
    "                           NOTE: only use with very robust rules, as this will otherwise introduce new errors\n" +
    "  --rulefile FILE          use an additional grammar file; if the filename contains a known language code,\n" +
    "                           it is used in addition of standard rules\n" +
    "  --falsefriends FILE      use external false friend file to be used along with the built-in rules\n" +
    "  --bitextrules  FILE      use external bitext XML rule file (useful only in bitext mode)\n" +
    "  --languagemodel DIR      a directory with e.g. 'en' sub directory (i.e. a language code) that contains\n" +
    "                           '1grams'...'3grams' sub directories with Lucene indexes with\n" +
    "                           ngram occurrence counts; activates the confusion rule if supported\n" +
    "  --word2vecmodel DIR      a directory with e.g. 'en' sub directory (i.e. a language code) that contains\n" +
    "                           final_embeddings.txt and dictionary.txt; activates neural network based rules\n" +
    "  --neuralnetworkmodel DIR a base directory for various saved neural network models\n" +
    "  --fasttextmodel FILE     fasttext language detection model (optional), see https://fasttext.cc/docs/en/language-identification.html\n" +
    "  --fasttextbinary FILE    fasttext executable (optional), see https://fasttext.cc/docs/en/support.html\n" +
    "  --xmlfilter              remove XML/HTML elements from input before checking (deprecated)\n" +
    "  --line-by-line           work on file line by line (for development, e.g. inside an IDE)\n" +
    "'>"},

```

ybiquitous:
LanguageTool doesn't support any markup languages e.g. Markdown, HTML, etc.:
https://github.com/languagetool-org/languagetool/issues/445

So, we decide to support only `.txt` files except for `requirements.txt` and `robots.txt` because analyzing markup files can produce too many false positives.

But we also provide the options to change `.txt` to `.md` or `.html`, so users can analyze their Markdown or HTML files as they want.

This PR is ready to review. But I'm writing a document on sider/sider-docs to understand this runner well.

lib/runners/processor/languagetool.rb
```diff
@@ -0,0 +1,183 @@
...
+          next if matches.empty?
+
+          # NOTE: Line numbers are not output to JSON, so we need to calculate them by ourselves.
+          offset_ranges = offset_ranges_per_line(File.readlines(filepath, encoding: config_encoding))
+          relative_file = relative_path(filepath)
+
+          matches.each do |match|
+            offset = Integer(match.fetch(:offset))
+            line_index = offset_ranges.find_index { |range| range.include?(offset) } or
+              raise "Not found line number from the offset #{offset}: ranges=#{offset_ranges}, file=#{relative_file}"
```
yykamei, line 144: `offset_ranges` is a little tricky for me. I wonder if we could define a class which is responsible for handling the offset like this:

```ruby
    class OffsetHandler
      def initialize(filepath, encoding)
        @filepath = filepath
        @encoding = encoding
      end

      def line_by(offset)
        offset_ranges.find_index { |range| range.include?(offset) }&.then { |v| v + 1 } or
          raise "Not found line number from the offset #{offset}: ranges=#{offset_ranges}, file=#{@filepath}"
      end

      private

      def offset_ranges
        @offset_ranges ||=
          begin
            start = 0
            File.readlines(@filepath, encoding: @encoding).map do |line|
              last = start + line.size
              (start...last).tap { start = last }
            end
          end
      end
    end
```

lib/runners/processor/languagetool.rb
```diff
@@ -0,0 +1,183 @@
...
+          target: string?,
+          ext: array?(string),
+          exclude: array?(string),
+          language: string?,
+          encoding: string?,
+          disable: array?(string),
+          enable: array?(string),
+          enabled_only: boolean?,
+          disable_categories: array?(string),
+          enable_categories: array?(string),
```
yykamei, line 17: I realize that containing underscore is better in terms of name conventions, but it would be good to follow the original option name `--enabledonly` because of the consistency. Other analyzers, such as ESLint, is following the original option name like `--ignore-path`.

ybiquitous:
> I'm curious about what happens in a smoke test when the option `disable` and `enabled_only` are set at the same time.

`disable` is a typo of `enalbe`? Because `enable` and `enable_categories` does not work as we expect unless `enabled_only`, right?

lib/runners/processor/languagetool.rb
ybiquitous: Yes, we could do it of course, but honestly, it looks the solution is excessive for this case... 🤔

lib/runners/processor/languagetool.rb
ybiquitous: I think that using a pattern like `enabled-only` instead of `enabled_only` may rather make users confusing. Because `--enabled-only` is incorrect and `--enabledonly` is correct.

We have already used the pattern like `root_dir` or `npm_install` for our own options, so I think `enabled_only` does not break the consistency.

ybiquitous: Here is another solution: supporting both patterns like `enabled-only:` and `enabled_only:`

ybiquitous:
@yykamei Thanks for your review! 😄 

I've written the document that may help you review:

https://github.com/sider/sider-docs/pull/357
https://deploy-preview-357--sider-docs.netlify.app/tools/others/languagetool

lib/runners/processor/languagetool.rb
yykamei: The "consistency" is something among analyzers. 

|              |Original Option | Runners Option  | Following original? |
| -------------|----------------|-----------------|---------------------|
| ESLint       | --ignore-path  | "ignore-path"   | Yes                 |
| LanguageTool | --enabledonly  | enabled_only    | No                  |

ESLint follows the original option, but LanguageTool does not. Isn't it consistent?

lib/runners/processor/languagetool.rb
yykamei: OK, I understood. Keep as it is.

yykamei:
> `disable` is a typo of `enalbe`? Because `enable` and `enable_categories` does not work as we expect unless `enabled_only`, right?

No, I read [this](https://github.com/languagetool-org/languagetool/blob/8aadacce1163083d5ef894b07978599ee27b35c6/languagetool-commandline/src/main/java/org/languagetool/commandline/CommandLineParser.java#L65) code, and I wonder something will happen when both options are set. In this case, we should return `Failure` instead of `Error`.

lib/runners/processor/languagetool.rb
ybiquitous: OK, what about an `enabledonly` pattern from the original?

I think the original pattern is easy to make typos, but it can keep the consistency.

ybiquitous:
Make sense. I will change the implementation and add a smoke test case. 👍

lib/runners/processor/languagetool.rb
ybiquitous: Users may look the following error message:

> You cannot specify both disabled rules and enabledonly

So, I think better to use `enabledonly` instead of `enabled_only` or `enabled-only`. What do you think?

yykamei: > I think the original pattern is easy to make typos

I agree with it. If the LanguageTool users aren't confused, `enabled_only` is better. Do they get confused? If not, I like `enabled_only`.

yykamei: > > You cannot specify both disabled rules and enabledonly
> 
> So, I think better to use `enabledonly` instead of `enabled_only` or `enabled-only`. What do you think?

Oh, I see. The error message contains `enabledonly`, right? If so, `enabledonly` is good.

ybiquitous: Yes, you teached it to me:

https://github.com/languagetool-org/languagetool/blob/8aadacce1163083d5ef894b07978599ee27b35c6/languagetool-commandline/src/main/java/org/languagetool/commandline/CommandLineParser.java#L65

```java
throw new IllegalArgumentException("You cannot specify both disabled rules and enabledonly");
```

I will fix the option names. Thanks!

ybiquitous:
I've addressed all the reviews. Could you please re-review? @yykamei

ybiquitous:
@yykamei Thanks for your review!

@mpls104 Thanks for your contribution!

It's a hard task. 😅 🎉

PR #798, opened 2020-03-02
Remove specified files via the `ignore` option

### LINKED ISSUE #760: Port the "ignore" feature to Runners
Currently, Sider removes the specified files/directories if users set the ["ignore" option](https://help.sider.review/getting-started/custom-configuration#ignore-option) in `sider.yml`.
Now, we are internally migrating to a new analysis system, and it expects Runners to remove those files.

I'm going to remove files here:

https://github.com/sider/runners/blob/f043b545358d710209f3eed3d2dc508825d813d9/lib/runners/workspace.rb#L48

yykamei:
Resolve #760


lib/runners/harness.rb
```diff
@@ -91,5 +92,15 @@ def handle_error(exn)
         trace_writer.error "#{exn.message} (#{exn.class})"
       end
     end
+
+    def remove_ignored_files(config)
+      globs = Array(config.content[:ignore])
```
ybiquitous: [question] Can we use `Config#ignore` here?

https://github.com/sider/runners/blob/4df6d6165e94be005a6a229c2659843487d1323f/lib/runners/config.rb#L32-L34

yykamei: Oh, I forgot the method (This is the one I created for the patch 👻).

test/harness_test.rb
```diff
@@ -203,6 +203,69 @@ def analyze(changes)
...
+    end
+
+    with_working_dir do |working_dir|
+      (working_dir / "sider.yml").write(<<~YAML)
+        ignore:
+          - examples/**/out
+          - test/**/out*
+          - "**/.idea"
+          - "*.log"
+      YAML
```
ybiquitous: [question] For example, Is `.pdf` value supported? Is this document incorrect?

https://help.sider.review/getting-started/custom-configuration#ignore-option

lib/runners/harness.rb
```diff
@@ -91,5 +92,15 @@ def handle_error(exn)
...
+      return if globs.empty?
+
+      trace_writer.header("Deleting specified files via the `ignore` option: #{globs.join(', ')}")
+      list = globs.flat_map { |g| Dir[working_dir / g] }
+      message = list.map { |l| Pathname(l).relative_path_from(working_dir) }.join("\n")
+      trace_writer.message(message) { FileUtils.rm_rf(list) }
```
ybiquitous: What about the following simple solution?

```suggestion
      trace_writer.message "Deleting files via the `ignore` option..." do
        FileUtils.rm_rf globs
        trace_writer.message globs.join("\n")
      end
```

- `trace_writer.header` looks better to me if under "Setting up source code".

image

ybiquitous: Sorry, the following is correct 🙏 

```suggestion
      trace_writer.message "Deleting files via the `ignore` option..." do
        files = Pathname(working_dir).glob(globs)
        FileUtils.rm_rf files
        trace_writer.message files.map { |f| f.relative_path_from(working_dir) }.join("\n")
      end
```

test/harness_test.rb
yykamei: Yes, the document is wrong. I updated with sider/sider-docs#313.

ybiquitous:
We are currently using **gitignore(5)**, but this change changes it to **glob** format.
So the following will be breaking:

```yaml
ignore:
  - "*.pdf"
```

* **gitignore(5)** - ignores **recursively** files with the `.pdf` extension
* **glob** - ignores files **only** direct under the root directory

yykamei:
Yes, are there any good alternatives?

ybiquitous:
I'm not sure about alternatives...

ybiquitous:
Can we use this? https://git-scm.com/docs/git-check-ignore

lib/runners/harness.rb
yykamei: Fixed with 944b7df

sig/polyfill.rbi
```diff
@@ -82,11 +82,13 @@ class FileUtils
   def self.install: (String, String, ?any) -> void
   def self.copy_entry: (any, any) -> void
   def self.remove_entry: (String|Pathname, ?bool) -> void
+  def self.rm_rf: (Array<String>) -> void
 end
 
 class Dir
   def self.mktmpdir: <'a> { (String) -> 'a } -> 'a
                    | () -> String
+  def self.[]: (*(String | Pathname)) -> Array<String>
```
ybiquitous: [question] Is this unneeded?

```suggestion
```

yykamei: Yes, it's not necessary, but do I have to remove it?

ybiquitous: As you want. 😉

yykamei: OK, I now remember the migration to Steep 0.14 is WIP. I should remove it for the task.

yykamei:
To use git-check-ignore(1), Runners must configure Git in the working directory.

```ruby
capture3!("git", "init")
(working_dir / ".gitignore").write(ignore)
stdout, = capture3!("git", "check-ignore", *Dir["**/*"])
FileUtils.rm_rf(stdout.lines(chomp: true))
```

This overwrites the file `.gitignore`, but it looks good.
In the first place, users don't have to write `sider.yml` if they have `.gitignore`...

ybiquitous:
> In the first place, users don't have to write `sider.yml` if they have `.gitignore`...

What do you mean? For example, if a user has committed `foo.pdf`, the user should write `sider.yml` with `ignore: [".pdf"]`?

yykamei:
OK, I was wrong. Forget about that.

What do you think about [this](https://github.com/sider/runners/pull/798#issuecomment-593223724) approach?

ybiquitous:
Looks good! 👍

yykamei:
@ybiquitous I updated this PR to use git-check-ignore(1), would you please review it when you have time?

lib/runners/harness.rb
```diff
@@ -91,5 +94,33 @@ def handle_error(exn)
         trace_writer.error "#{exn.message} (#{exn.class})"
       end
     end
+
+    def remove_ignored_files(config)
+      ignores = config.ignore
+      return if ignores.empty?
+
+      trace_writer.message("Deleting specified files via the `ignore` option...") do
```
ybiquitous: [IMO] This message implicitly depends on `ignore:` option of `sider.yml`, and we may forget to update if the name will change in the future.

So, what about changing the message like this?

```suggestion
      trace_writer.message("Deleting ignored files...") do
```

lib/runners/harness.rb
```diff
@@ -91,5 +94,33 @@ def handle_error(exn)
...
+    end
+
+    def with_gitignore(ignores)
+      gitignore = (working_dir / ".gitignore")
+
+      backup = gitignore.file? ? gitignore.read : nil
+      gitignore.write(ignores.join("\n"))
+
+      shell = Shell.new(current_dir: working_dir, trace_writer: trace_writer, env_hash: {})
+      shell.capture3!("git", "init")
```
ybiquitous: [suggestion] I'm concerned that `git init` output may be unuseful and make confused users, so what about suppressing as follow?

```suggestion
      shell.capture3!("git", "init", trace_command_line: false, trace_stdout: false)
```

```diff
diff --git a/lib/runners/shell.rb b/lib/runners/shell.rb
index 47db1296..376446b7 100644
--- a/lib/runners/shell.rb
+++ b/lib/runners/shell.rb
@@ -92,13 +92,14 @@ module Runners
 
     def capture3_trace(command, *args, **options)
       # @type var options: any
+      trace_command_line = options.fetch(:trace_command_line, true)
       trace_stdout = options.fetch(:trace_stdout, true)
       trace_stderr = options.fetch(:trace_stderr, true)
       raise_on_failure = options.fetch(:raise_on_failure, false)
       is_success = options.fetch(:is_success) { ->(status) { status.success? } }
 
       command_line = [command] + args
-      trace_writer.command_line(command_line)
+      trace_writer.command_line(command_line) if trace_command_line
 
       Open3.capture3(env_hash, command, *args, { chdir: current_dir.to_s }).tap do |stdout_str, stderr_str, status|
         trace_writer.stdout stdout_str if trace_stdout
```

lib/runners/harness.rb
```diff
@@ -91,5 +94,33 @@ def handle_error(exn)
...
+      stdout, = shell.capture3("git", "check-ignore", *working_dir.glob("**/*", File::FNM_DOTMATCH).map(&:to_s))
```
ybiquitous: [suggestion] What about checking the exit status for safety like this?

```suggestion
      # @see https://git-scm.com/docs/git-check-ignore#_exit_status
      stdout, = shell.capture3!("git", "check-ignore", *working_dir.glob("**/*", File::FNM_DOTMATCH).map(&:to_s), is_success: ->(s) { [0, 1].include?(s.exitstatus) })
```

See https://git-scm.com/docs/git-check-ignore#_exit_status

lib/runners/harness.rb
yykamei: Looks good. I updated with 740f7e7

lib/runners/harness.rb
yykamei: Good. I updated with 9e49f19.

lib/runners/harness.rb
ybiquitous: [suggestion] What about adding `trace_command_line: false, trace_stdout: false` also here?

yykamei: OK, I do it. I think to check whether the exit status is `128` or not seems better.

yykamei: 3ea37cd

ybiquitous: What do you think about https://github.com/sider/runners/pull/798#discussion_r386255007?

yykamei: 66accd5

lib/runners/harness.rb
```diff
@@ -91,5 +94,35 @@ def handle_error(exn)
...
+      gitignore = (working_dir / ".gitignore")
+
+      backup = gitignore.file? ? gitignore.read : nil
+      gitignore.write(ignores.join("\n"))
+
+      shell = Shell.new(current_dir: working_dir, trace_writer: trace_writer, env_hash: {})
+      shell.capture3!("git", "init", trace_command_line: false, trace_stdout: false)
+
+      # @see https://git-scm.com/docs/git-check-ignore#_exit_status
+      stdout, = shell.capture3!("git", "check-ignore", *working_dir.glob("**/*", File::FNM_DOTMATCH).map(&:to_s), trace_command_line: false, trace_stdout: false, is_success: -> (s) { s.exitstatus != 128 })
```
ybiquitous: [suggestion] Can you format the code layout like this? I feel too long line hard to read... 😢 

```suggestion
      stdout, = shell.capture3!(
        "git", "check-ignore", *working_dir.glob("**/*", File::FNM_DOTMATCH).map(&:to_s),
        trace_command_line: false, trace_stdout: false, is_success: -> (s) { s.exitstatus != 128 })
```

lib/runners/harness.rb
```diff
@@ -91,5 +94,35 @@ def handle_error(exn)
...
+    end
+
+    def with_gitignore(ignores)
+      gitignore = (working_dir / ".gitignore")
+
+      backup = gitignore.file? ? gitignore.read : nil
+      gitignore.write(ignores.join("\n"))
+
+      shell = Shell.new(current_dir: working_dir, trace_writer: trace_writer, env_hash: {})
+      shell.capture3!("git", "init", trace_command_line: false, trace_stdout: false)
```
ybiquitous: [question] Will `Workspace::Git` class affect this `git init`?

yykamei: No, because `Workspace::Git` owns its `git_directory` and manages Git information in the directory.

ybiquitous: I see, thanks 👍

lib/runners/shell.rb
```diff
@@ -94,11 +94,12 @@ def capture3_trace(command, *args, **options)
       # @type var options: any
       trace_stdout = options.fetch(:trace_stdout, true)
       trace_stderr = options.fetch(:trace_stderr, true)
+      trace_command_line = options.fetch(:trace_command_line, true)
       raise_on_failure = options.fetch(:raise_on_failure, false)
       is_success = options.fetch(:is_success) { ->(status) { status.success? } }
 
       command_line = [command] + args
-      trace_writer.command_line(command_line)
+      trace_writer.command_line(command_line) if trace_command_line
```
ybiquitous: [suggestion] Sorry, it seems that we need to suppress also `trace_writer.status`. 🤔 

image

What about this? 👇 

```diff
diff --git a/lib/runners/shell.rb b/lib/runners/shell.rb
index 8a4eb7c3..10a00b58 100644
--- a/lib/runners/shell.rb
+++ b/lib/runners/shell.rb
@@ -106,7 +106,7 @@ module Runners
         trace_writer.stderr stderr_str if trace_stderr
 
         if status.exited?
-          trace_writer.status status
+          trace_writer.status status if trace_command_line
         else
           trace_writer.error "Process aborted or coredumped: #{status.inspect}"
         end
```

yykamei: OK

yykamei:
Thank you :pray:

PR #800, opened 2020-03-02
[Refactor] More strict type-check

ybiquitous:
For safety and maintenancibility.

### DIFF
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -23,7 +23,7 @@ module Runners
 
       hash = {
         "RUBYOPT" => nil,
-        "GIT_SSH" => git_ssh_path,
+        "GIT_SSH" => git_ssh_path&.to_path,
       }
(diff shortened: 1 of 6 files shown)


lib/runners/processor.rb
```diff
@@ -23,7 +23,7 @@ def initialize(guid:, workspace:, config:, git_ssh_path:, trace_writer:)
 
       hash = {
         "RUBYOPT" => nil,
-        "GIT_SSH" => git_ssh_path,
+        "GIT_SSH" => git_ssh_path&.to_path,
```
ybiquitous: [note] I prefer `#to_path` over `#to_s` because this can clarify the difference with `NilClass#to_s`.

See https://ruby-doc.org/stdlib-2.7.0/libdoc/pathname/rdoc/Pathname.html#method-i-to_path

ybiquitous:
CI passed. It looks no problem.

PR #805, opened 2020-03-03
Aggregate analyzers metadata

ybiquitous:
This aggregates all analyzers' metadata to a YAML file.
By this metadata file, we will be able to make some automation (e.g. an auto-generated README section).

### DIFF
--- a/lib/tasks/bump/analyzers.rb
+++ b/lib/tasks/bump/analyzers.rb
@@ -49,6 +49,7 @@ BumpAnalyzers = Struct.new(
 
   def self.each
     analyzers = YAML.safe_load(Pathname(__dir__).join("../../../analyzers.yml").read, symbolize_names: true).fetch(:analyzers)
+    analyzers.keep_if { |_, meta| meta[:dependabot] == false && !meta[:deprecated]  }
     analyzers.each do |analyzer, meta|
       yield new(
         analyzer: analyzer,
--- a/lib/tasks/bump/analyzers.rb
+++ b/lib/tasks/bump/analyzers.rb
@@ -1,5 +1,6 @@
 require "net/http"
 require "json"
+require "yaml"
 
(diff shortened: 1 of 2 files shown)


ybiquitous:
@yykamei I want to listen to your opinion. 🙏

lib/tasks/bump/analyzers.rb
```diff
@@ -46,20 +47,9 @@ def initialize(response)
...
-    javasee:  { name: "JavaSee", github: "sider/JavaSee" },
-    misspell: { name: "Misspell", github: "client9/misspell" },
-    shellcheck: { name: "ShellCheck", github: "koalaman/shellcheck" },
-    swiftlint: { name: "SwiftLint", github: "realm/SwiftLint" },
-  }.freeze
-
   def self.each
-    ANALYZERS_UNSUPPORTED_BY_DEPENDABOT.each do |analyzer, meta|
+    analyzers = YAML.safe_load(Pathname(__dir__).join("../../../analyzers.yml").read, symbolize_names: true).fetch(:analyzers)
+    analyzers.each do |analyzer, meta|
```
yykamei: Is it OK to walk all analyzers here? Before this patch, only `ANALYZERS_UNSUPPORTED_BY_DEPENDABOT` seemed iterated.

ybiquitous: Oh, my bad. 😖 
I'll fix it soon, thank you!

ybiquitous: Fixed via 98c555f.

ybiquitous:
I think this metadata file can declaratively provide data such as the following `analyzer_name`:

https://github.com/sider/runners/blob/12c1b856673ca560e7ed1ab0c837c925a18977e8/lib/runners/processor.rb#L51-L53

PR #807, opened 2020-03-03
Auto-generated README content

ybiquitous:
This adds a new rake task `readme:generate` via the metadata in `analyzers.yml`.
The new task is executed on CI, so we will not need to maintain the analyzers list on `README.md`.

See also #805


ybiquitous:
Thanks!

PR #808, opened 2020-03-03
[Refactor] Make it easier to implement a `Processor` subclass

### LINKED ISSUE #786: Rename `ci_config_section_name` to `analyzer_id`.
The name `ci_config_section_name` is not appropriate because Sider is not a CI.
I want to change the method with `analyzer_id`.
Also, I can use the new method `analyzer_id` in the file below:

https://github.com/sider/runners/blob/09f9e08516d053b006eff0cb6e3581810a1696eb/lib/runners/cli.rb#L34-L38

This is the newer version that uses `analyzer_id`.

```ruby
def processor_class
  @processor_class ||= (Processor.subclasses.detect do |cls|
    analyzer == cls.analyzer_id
  end or raise "Not found processor class with '#{analyzer}'")
end
```

_Originally posted by @yykamei in https://github.com/sider/runners/pull/763#issuecomment-592442055_

ybiquitous:
This change aims to make it easier to implement a processor subclass.

- Read metadata from `analyzers.yml`.
- Define dynamically a `#analyzer_id` method.

Fix #786


test/processor_test.rb
```diff
@@ -127,61 +127,6 @@ def test_relative_path_from
...
-            quiet: nil,
-          },
-        },
-        processor.ci_section,
-      )
-
-      assert(trace_writer.writer.find { |hash| hash[:trace] == :ci_config && hash[:content] })
-    end
-  end
-
```
ybiquitous: [note] This logic is tested by `AnalyzersTest`, so I think it unneeded.

test/smokes/brakeman/expectations.rb
```diff
@@ -45,7 +45,7 @@
         git_blame_info: nil
       }
     ],
-    analyzer: { name: "brakeman", version: "4.3.1" }
+    analyzer: { name: "Brakeman", version: "4.3.1" }
```
ybiquitous: [note] I think this change has no problems because `analyzer.name` is not actually used anywhere.

ybiquitous:
@yykamei I'm sorry for large diffs, but could you please review this PR? (Tomorrow is OK)

PR #819, opened 2020-03-05
Add raw content of config file to trace

ybiquitous:
This change aims to make it easier for users to tackle a configuration error by themselves.

See also #818

### DIFF
--- a/lib/runners/harness.rb
+++ b/lib/runners/harness.rb
@@ -78,7 +78,11 @@ module Runners
           end
         end
       rescue Config::Error => exn
-        trace_writer.error exn.message
+        trace_writer.error <<~MSG.strip
+          #{exn.message}
+          ---
+          #{exn.raw_content}
+        MSG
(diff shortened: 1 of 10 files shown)


lib/runners/harness.rb
```diff
@@ -78,7 +78,11 @@ def ensure_result
           end
         end
       rescue Config::Error => exn
-        trace_writer.error exn.message
+        trace_writer.error <<~MSG.strip
+          #{exn.message}
+          ---
+          #{exn.raw_content}
+        MSG
```
ybiquitous: For example:

```ruby
{:trace=>"error",
 :message=>
  "The value of the attribute `$.linter.goodcheck.config` in your `sideci.yml` is invalid. Please fix and retry.\n" +
  "---\n" +
  "linter:\n" +
  "  goodcheck:\n" +
  "    config: true",
 :recorded_at=>"2020-03-05T03:10:28.684Z",
 :truncated=>false}
```

PR #825, opened 2020-03-06
Push Docker images with `master` tag

ybiquitous:
It assumes that Docker images with a `master` tag will be used only for development.


ybiquitous:
I will release the next version 0.21.3 for the test after merging this PR.

PR #830, opened 2020-03-06
Do not use base for git-blame(1)

### LINKED ISSUE #828: #range_git_blame_info fails with "fatal: More than one commit to dig from %s and %s?"
This might be a bug of Runners, but I couldn't confirm why it happens and how should I reproduce this problem.

yykamei:
Fix #828

### DIFF
--- a/lib/runners/git_blame_info.rb
+++ b/lib/runners/git_blame_info.rb
@@ -1,10 +1,10 @@
 module Runners
   GitBlameInfo = Struct.new(:commit, :original_line, :final_line, :line_hash, keyword_init: true) do
-    # @param porelain_output [String] The output of git-blame(1) with -p option
+    # @param porcelain_output [String] The output of git-blame(1) with -p option
     # @see https://www.git-scm.com/docs/git-blame#_the_porcelain_format
-    def self.parse(porelain_output)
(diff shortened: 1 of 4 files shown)


lib/runners/git_blame_info.rb
```diff
@@ -1,10 +1,10 @@
...
   GitBlameInfo = Struct.new(:commit, :original_line, :final_line, :line_hash, keyword_init: true) do
-    # @param porelain_output [String] The output of git-blame(1) with -p option
+    # @param porcelain_output [String] The output of git-blame(1) with -p option
     # @see https://www.git-scm.com/docs/git-blame#_the_porcelain_format
-    def self.parse(porelain_output)
-      headers = porelain_output.scan(/^([a-z0-9]{40}) (\d+) (\d+)/)
-      contents = porelain_output.scan(/^\t(.*)$/).flatten
+    def self.parse(porcelain_output)
+      headers = porcelain_output.scan(/^([a-z0-9]{40}) (\d+) (\d+)/)
+      contents = porcelain_output.scan(/^\t(.*)$/).flatten
```
yykamei: Typo :keyboard:

PR #837, opened 2020-03-10
Do not write the output of git-diff(1)

yykamei:
Sometimes, `Shell#capture3!` fails with the error message "source sequence is illegal/malformed utf-8."

In the first place, the output of git-diff(1) is unnecessary, so I make the caller use `:trace_stdout`.

### DIFF
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -58,7 +58,7 @@ module Runners
       if base && head
         git_directory.yield_self do |path|
           shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
-          stdout, _ = shell.capture3!("git", "diff", base, head)
+          stdout, _ = shell.capture3!("git", "diff", base, head, trace_stdout: false)
           GitDiffParser.parse(stdout)
         end
       else


lib/runners/workspace/git.rb
```diff
@@ -58,7 +58,7 @@ def patches
       if base && head
         git_directory.yield_self do |path|
           shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
-          stdout, _ = shell.capture3!("git", "diff", base, head)
+          stdout, _ = shell.capture3!("git", "diff", base, head, trace_stdout: false)
```
yykamei: Should `:trace_command_line` be `false`, too?

ybiquitous: Will not this disabling be useful for debugging? Can `git diff --output` option help us?

yykamei: I agree with you. For debugging, the output would be useful.

ybiquitous: I'm not sure what is the file encoding when using the `--output=<file>` option. 🤔 
`man git-diff` does not seem to describe it...?

yykamei: :thinking: git-diff(1) seems to dump the output with the raw encoding. So, before reading the output, Runners ensure what the output uses actual encoding. I couldn't find such a feature in Ruby, but file(1) could work. However, I feel cumbersome to deal with the issue... I want to just use `trace_stdout: false`.

ybiquitous: Thanks for the research. I agree with using `trace_stdout: false`. 👍

ybiquitous: For debugging, I don't think that `trace_command_line: false` is necessary, doesn't it?

yykamei: Agree. I have the same opinion.

PR #844, opened 2020-03-11
Improve warnings for Node.js runners

ybiquitous:
- Add warning if `package.json` is not found and `npm_install` option is specified.
- Refactor unit test for `Runners::Nodejs` module.

### DIFF
--- a/lib/runners/nodejs.rb
+++ b/lib/runners/nodejs.rb
@@ -50,25 +50,33 @@ module Runners
 
     # Install Node.js dependencies by using given parameters.
     def install_nodejs_deps(defaults, constraints:, install_option:)
-      install_option = INSTALL_OPTION_ALL if install_option.nil?
-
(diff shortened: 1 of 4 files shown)


lib/runners/nodejs.rb
```diff
@@ -50,25 +50,33 @@ def yarn_lock_path
...
-        else
-          npm_install(install_option)
-        end
+      if install_option && !package_json_path.exist?
+        file = package_json_path.basename.to_path
+        add_warning <<~MSG, file: file
+          The `npm_install` option is specified, but a `#{file}` file is not found. In this case, Sider does not install any npm packages.
+        MSG
+        return
+      end
```
ybiquitous: This PR's core is here. What do you think about this? @yykamei

ybiquitous:
Thank you!

PR #850, opened 2020-03-11
Improve warning in `Nodejs#install_nodejs_deps` method

ybiquitous:
- Do not run any extra `npm install`.
- Improve the warning message.

Follow up #844

### DIFF
--- a/lib/runners/nodejs.rb
+++ b/lib/runners/nodejs.rb
@@ -54,12 +54,16 @@ module Runners
 
       return if install_option == INSTALL_OPTION_NONE
 
-      if install_option && !package_json_path.exist?
-        file = package_json_path.basename.to_path
-        add_warning <<~MSG, file: file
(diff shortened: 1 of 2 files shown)


lib/runners/nodejs.rb
```diff
@@ -54,12 +54,16 @@ def install_nodejs_deps(defaults, constraints:, install_option:)
...
+      unless package_json_path.exist?
+        if install_option
+          file = package_json_path.basename.to_path
+          add_warning <<~MSG, file: file
+            The `npm_install` option is specified in your `#{config.path_name}`, but a `#{file}` file is not found in the repository.
+            In this case, any npm packages are not installed.
+          MSG
+        end
+
+        return # not install
```
ybiquitous: [note] Before this change, in the case of `install_option != nil` and `package_json_path.exist? == false`, an extra `npm install` command was run.

PR #852, opened 2020-03-12
Refactor `Processor` and `Harness` classes

ybiquitous:
This change is a simple refactoring.

### DIFF
--- a/lib/runners/harness.rb
+++ b/lib/runners/harness.rb
@@ -36,17 +36,17 @@ module Runners
           remove_ignored_files(conf)
 
           begin
-            instance = processor_class.new(guid: guid, workspace: workspace, config: conf, git_ssh_path: git_ssh_path, trace_writer: trace_writer)
+            processor = processor_class.new(guid: guid, workspace: workspace, config: conf, git_ssh_path: git_ssh_path, trace_writer: trace_writer)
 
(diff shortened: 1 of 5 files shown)


lib/runners/harness.rb
```diff
@@ -36,17 +36,17 @@ def run
           remove_ignored_files(conf)
 
           begin
-            instance = processor_class.new(guid: guid, workspace: workspace, config: conf, git_ssh_path: git_ssh_path, trace_writer: trace_writer)
+            processor = processor_class.new(guid: guid, workspace: workspace, config: conf, git_ssh_path: git_ssh_path, trace_writer: trace_writer)
```
ybiquitous: The name `instance` is ambiguous.

sig/polyfill.rbi
```diff
@@ -27,8 +27,6 @@ end
 extension Object (Polyfill)
   def instance_of?: (any) -> bool
   def then: <'a> () { (self) -> 'a } -> 'a
-  def Array: (any) -> Array<any>
-  def Pathname: (String) -> Pathname
```
ybiquitous: They should be in `Kernel`:

```console
$ ruby -r pathname -e 'p method(:Pathname).owner ; p method(:Array).owner'
Kernel
Kernel
```

PR #853, opened 2020-03-13
Refactor `Processor` and `Harness` classes

ybiquitous:
## Goal

- Make the `Harness` class more simple and readable.
- Make it easier to implement a `Processor` subclass.

## How

- Extract a new `Ignoring` class from the `Harness` class in order to split the responsibility. By this, the code is readable and testable.
- Move the location loading a `sider.yml`: `Harness` to `Processor`. By this, the `Harness`'s dependency on `Config` decreases.
- In `Processor`, `#register_config_schema` uses internally `#analyzer_id`. By this, the code is more DRY, see below:

```diff
-register_config_schema(name: :checkstyle, schema: Schema.runner_config)
+def config_schema
+  Schema.runner_config
+end
```

Please see each commit for details.


ybiquitous:
This challenge (especially a58f194) does not work, so I closed.

We should register all schemas for `Processor` subclasses.

PR #864, opened 2020-03-17
Improve HTTP download

ybiquitous:
We are currently using [`Net::HTTP`](https://ruby-doc.org/stdlib-2.6.5/libdoc/net/http/rdoc/Net/HTTP.html) to download a user's source code.

Because the `Net::HTTP` has own retry mechanism, this change aims to make our retrying more simple.

### DIFF
--- a/lib/runners/workspace/http.rb
+++ b/lib/runners/workspace/http.rb
@@ -73,7 +73,7 @@ module Runners
               max_redirects: max_redirects - 1,
             )
           else
-            raise DownloadError, "Download failed: #{response.inspect}"
+            raise DownloadError, "Download failed: #{response.code} #{response.message}"
           end
         end
       end
--- a/lib/runners/workspace/http.rb
+++ b/lib/runners/workspace/http.rb
@@ -26,7 +26,7 @@ module Runners
       end
     end
 
-    # NOTE: Some exceptions are handled by ``Net::HTTP` class.
+    # NOTE: Some exceptions are handled by `Net::HTTP` class.
     #
     # @see https://github.com/ruby/ruby/blob/v2_6_5/lib/net/http.rb#L1520-L1536
     def download_with_retry(uri, &block)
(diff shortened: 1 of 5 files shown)


lib/runners/workspace/http.rb
```diff
@@ -19,40 +19,58 @@ def prepare_head_source(dest)
...
+        use_ssl: true,
+        max_retries: max_retries,
+      }
+
+      Net::HTTP.start(uri.hostname, uri.port, http_options) do |http|
+        http.request_get(uri) do |response|
+          case response
+          when Net::HTTPSuccess
+            response.read_body(dest)
+          when Net::HTTPRedirection
```
ybiquitous: [note] See also the link below about redirection:

https://ruby-doc.org/stdlib-2.6.5/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-Following+Redirection

ybiquitous:
@yykamei This PR is in progress yet, but can you tell me your idea? Can I continue?

lib/runners/workspace/http.rb
```diff
@@ -19,40 +19,62 @@ def prepare_head_source(dest)
     # @param dest [Pathname]
     # @param key [String, nil]
     def provision(uri, dest, key)
-      # When OpenSSL::Cipher::CipherError, probably the downloading is failed.
```
ybiquitous: [question] Can `#decrypt` method raise `OpenSSL::Cipher::CipherError`?

https://github.com/sider/runners/blob/c8b77180b66344e868da3868efa20f05daca522c/test/workspace/http_test.rb#L19

yykamei: I want to expect `Net::HTTP` to handle such an error with `max_retries`. Let's forget `CipherError`

yykamei:
I think this is a nice refactoring.

PR #866, opened 2020-03-17
Migration to new Steep 0.14.0 gradually

ybiquitous:
The migration to Steep 0.14.0 is a big change because of the changes for the `.rbs` format and the syntax.
So, I think it better that we will migrate our code gradually and keep both for a while.

This change does mainly the two things:

1. Install the new version of Steep via another Gemfile, that is `Gemfile.steep`.
2. Run both type-checking (via the old and new version) on CI.

When we will complete the migration, we will be able to remove the old code.

Close #719

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,6 +76,8 @@ jobs:
   typecheck_new:
     runs-on: ubuntu-latest
     needs: check_skip
+    env:
+      BUNDLE_GEMFILE: Gemfile.steep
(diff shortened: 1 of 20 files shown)


ybiquitous:
Note: If you try the new version, run

```
$ bundle install --gemfile Gemfile.steep
```

then,

```
$ bundle exec steep check
```

ybiquitous:
Thanks. Let's typing!

PR #870, opened 2020-03-18
Fix `rake bump:analyzers` task

ybiquitous:
- The regex to detect a version does not work. So this fixes the script and Dockerfiles.
- Add `DRY_RUN` option to help us test the task locally.
  (usage: `rake bump:analyzers DRY_RUN=1`)
- Manage cpplint via Dependabot. (it has `Pipfile`)
  See #869


ybiquitous:
I will try on master. 💪

PR #875, opened 2020-03-18
Add smoke test cases about `*.gemspec` files

### LINKED ISSUE #652: The strict gemspec file causes an error of analyses
If the target repository contains a gemspec file which requires Bundler dependency, some analyses will fail because of its strict dependency.

```ruby
spec.add_development_dependency "bundler", "~> 2.0.0"
```

In this gemspec file, we should use Bundler 2.0.x, but our current Runners use Bundler 2.1.x, so analyses will fail.

> Maybe. We might have to install Bundler `2.1.x`, `2.0.x`, and `1.x` in case we can handle the various gemspec files.

_Originally posted by @yykamei in https://github.com/sider/runners/pull/646/review_comment/create_

ybiquitous:
The new `old_bundler_via_gemspec` case outputs an error message as follows, but the analysis continues.

```
{:trace=>"stderr",
 :string=>
  "Bundler could not find compatible versions for gem \"bundler\":\n" +
  "  In Gemfile:\n" +
  "    bundler (= 2.0.0)\n" +
  "\n" +
  "  Current Bundler version:\n" +
  "    bundler (2.1.4)\n" +
  "This Gemfile requires a different version of Bundler.\n" +
  "Perhaps you need to update Bundler by running `gem install bundler`?\n" +
  "\n" +
  "Could not find gem 'bundler (= 2.0.0)' in any of the relevant sources:\n" +
  "  the local ruby installation",
 :recorded_at=>"2020-03-18T09:21:41.047Z",
 :truncated=>false}
{:trace=>"status", :status=>6, :recorded_at=>"2020-03-18T09:21:41.047Z"}
{:trace=>"error",
 :message=>
  "An error was found in your Gemfile but Sider continues the analysis by ignoring your dependencies.\n" +
  "If you want to make Sider aware of your dependencies, please fix the Gemfile according to the above error message, or commit your Gemfile.lock.",
 :recorded_at=>"2020-03-18T09:21:41.047Z",
 :truncated=>false}
 ```

Close #652

### DIFF
--- a/test/smokes/rubocop/expectations.rb
+++ b/test/smokes/rubocop/expectations.rb
@@ -476,3 +476,13 @@ Smoke.add_test(
     }
   ]
 )
+
+Smoke.add_test(
+  "old_bundler_via_gemspec",
+  { guid: "test-guid", timestamp: :_, type: "success", issues: [], analyzer: { name: "RuboCop", version: "0.80.1" } }
+)
+
+Smoke.add_test(
+  "latest_bundler_via_gemspec",
+  { guid: "test-guid", timestamp: :_, type: "success", issues: [], analyzer: { name: "RuboCop", version: "0.80.1" } }
+)
--- a/test/smokes/rubocop/latest_bundler_via_gemspec/Gemfile
+++ b/test/smokes/rubocop/latest_bundler_via_gemspec/Gemfile
@@ -1,0 +1,3 @@
+source "https://rubygems.org"
+
+gemspec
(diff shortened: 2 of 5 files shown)


ybiquitous:
The related code:

https://github.com/sider/runners/blob/1f46c8ee2971151ca85199176e404e57610ba259/lib/runners/ruby/lockfile_loader.rb#L55-L58

ybiquitous:
I'm not sure. At least, all the cases by this PR succeed.

ybiquitous:
Thanks. Can we close #652?

yykamei:
Yes, please close it.

ybiquitous:
Close #652

PR #883, opened 2020-03-19
Fix the version of steep for Gemfile

yykamei:
We are currently maintaining Steep v0.11.1 and v0.14.0, and `Gemfile`
should include Steep v0.11.1. v0.14.0 is declared in `Gemfile.steep`.

This is a revert of #882


ybiquitous:
👍👍👍

PR #889, opened 2020-03-23
[PHPMD] Some improvements

ybiquitous:
This change includes some improvements but they are not breaking:

- Refactor the default ruleset file.
- Change error type for an invalid rule error: `error` -> `failure`
- Refactor the application code.
- Add some smoke test cases.

### DIFF
--- a/lib/runners/processor/phpmd.rb
+++ b/lib/runners/processor/phpmd.rb
@@ -79,22 +79,34 @@ module Runners
     end
 
     def run_analyzer(changes, targets, rule, options)
-      report_file = working_dir / "phpmd-report-#{Time.now.to_i}.xml"
+      report_file = Pathname(Tempfile.new("phpmd-report-").path)
 
       # PHPMD exits 1 when some violations are found.
       # The `--ignore-violation-on-exit` will exit with a zero code, even if any violations are found.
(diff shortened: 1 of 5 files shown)


lib/runners/processor/phpmd.rb
```diff
@@ -79,22 +79,34 @@ def strict
     end
 
     def run_analyzer(changes, targets, rule, options)
-      report_file = working_dir / "phpmd-report-#{Time.now.to_i}.xml"
+      report_file = Pathname(Tempfile.new("phpmd-report-").path)
```
ybiquitous: [note] Not to pollute the working directory.

lib/runners/processor/phpmd.rb
```diff
@@ -79,22 +79,34 @@ def strict
...
+          raise stderr
+        end
+      end
 
       output_xml =
         if report_file.exist?
           report_file.read
         else
-          raise "Failed to output a report file via the command: #{command_line.join(' ')}"
+          raise "Failed to output a report file"
```
ybiquitous: [note] The command line is output to the trace log, so this removes the redundant part of the message.

ybiquitous:
@yykamei Could you please review just in case? It's OK when you have time. 👍

PR #896, opened 2020-03-26
Add PHPStan

### LINKED ISSUE #746: Support PHPStan
## Information

- Website: https://github.com/phpstan/phpstan
- Language: PHP

## Why do you need the tool?

PHPStan is more popular than PHPMD.

## Additional context

None.

## See also

- https://php.libhunt.com/compare-phpstan-vs-php-mess-detector

yykamei:
We are planning to support PHPStan on Sider. This PR will add "runner_phpstan."

Fix #746

---

- [ ] Provide option(s) that users can customize.
- [ ] Write smoke test cases with enough coverage. E.g. all options, successful end, warnings, errors, etc.
- [ ] Add a [CI setting](https://github.com/sider/runners/blob/master/.github/workflows/build.yml).
- [ ] Add a new [changelog](https://github.com/sider/runners/blob/master/CHANGELOG.md) entry.
- [ ] Write a [document](https://github.com/sider/sider-docs) for the new runner. PR sider/sider-docs#<PULL_NUMBER>

---

**WARNINGS**

* PHPStan does not have rule IDs so far, but they're considering about it [here](https://github.com/phpstan/phpstan/issues/3065)
    - In this PR, I just set `message` to `id`...
* PHPStan supports extensions, but I don't add this feature
    - Before doing it, some refactoring on the `PHP` module is necessary


images/phpstan/phpstan-sider-formatter/src/SiderErrorFormatter.php
```diff
@@ -0,0 +1,25 @@
...
+        foreach ($analysisResult->getFileSpecificErrors() as $fileSpecificError) {
+            $errors[] = $fileSpecificError->jsonSerialize();
+        }
+
+        $json = Json::encode($errors);
+        $output->writeRaw($json);
+
+        return 0;
+    }
+}
```
yykamei: I added a custom error formatter class like this because builtin error formatters do not include all information.
The function `jsonSerialize` is defined here:

https://github.com/phpstan/phpstan-src/blob/f4a9d5f2f71e4615f03d2f6ca8f8fe184a936846/src/Analyser/Error.php#L170-L185

I implemented this class following the builtin `JsonErrorFormatter`.

https://github.com/phpstan/phpstan-src/blob/0.12.18/src/Command/ErrorFormatter/JsonErrorFormatter.php

images/phpstan/composer.json
```diff
@@ -1,5 +1,13 @@
...
+    {
+      "type": "path",
+      "url": "./phpstan-sider-formatter"
+    }
+  ],
   "require": {
-    "phpstan/phpstan": "^0.12.18"
+    "phpstan/phpstan": "0.12.18",
+    "sider/phpstan-sider-formatter": "@dev",
+    "phpstan/extension-installer": "^1.0"
```
yykamei: phpstan/extension-installer automatically includes extension configuration defined in `extra.phpstan` field in the library's `composer.json`. Using this, Our runner_phpstan can use the `sider` error formatter.

https://github.com/sider/runners/pull/896/commits/946f993dbde741f8115deed9d43db80f6fcb2e4d#diff-8f745073fcc2562c5bc1f1b74c5649bcR5-R11

Also, I confirmed this feature does not overwrite users' configuration.

yykamei:
Before analyzing with PHPStan, `composer install` seems necessary. It's hard for Runners to perform with PHPStan because the images of Runners may not contain system-level dependencies required by PHP packages...

https://github.com/phpstan/phpstan/blob/1ce27fe29c8660a27926127d350d53d80c4d4286/.github/workflows/build.yml#L116-L121

yykamei:
Unfortunately, it's hard for us to support PHPStan for now because we lack human resources...

PR #901, opened 2020-03-27
[detekt] Try fixing that report file is not found

ybiquitous:
The smoke test on CI sometimes fails, for example:

https://github.com/sider/runners/runs/535753379?check_suite_focus=true#step:8:79

### DIFF
--- a/lib/runners/processor/detekt.rb
+++ b/lib/runners/processor/detekt.rb
@@ -32,7 +32,7 @@ module Runners
     private
 
     def run_analyzer
-      report_path = Pathname(Tempfile.new("detekt-report-").path)
+      report_path = Pathname(Tempfile.new(["detekt-report-", ".xml"]).path)
 
       _stdout, stderr, status = capture3(
         analyzer_bin,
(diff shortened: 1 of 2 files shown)


lib/runners/processor/detekt.rb
```diff
@@ -32,7 +32,7 @@ def analyze(changes)
     private
 
     def run_analyzer
-      report_path = Pathname(Tempfile.new("detekt-report-").path)
+      report_path = Pathname(Tempfile.new(["detekt-report-", ".xml"]).path)
```
ybiquitous: See also: https://ruby-doc.org/stdlib-2.7.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-new

lib/runners/processor/detekt.rb
```diff
@@ -97,7 +97,7 @@ def cli_report(report_path)
     end
 
     def construct_result(report_path)
-      if report_path.exist?
+      if report_path.file?
```
ybiquitous: See also https://ruby-doc.org/core-2.7.0/FileTest.html#method-i-file-3F

PR #902, opened 2020-03-27
Add /workdir for the task "docker:shell"

yykamei:
I want to use files in `test/smokes` with `docker:shell`, so I added the `--volume` option for the task `docker:shell`.


ybiquitous:
Nice! 👏

PR #906, opened 2020-03-27
[Refactor] Handle output file from each analyzer

ybiquitous:
This change extracts the following common methods from each analyzer implementation:

- `Processor#read_output_file`
- `Processor#read_output_xml`
- `Processor#read_output_json`

Also, since recent smoke tests on CI are unstable due to the usage of `Tempfile.new`,
this fixes it to `Tempfile.create` and adds a Querly rule about the usage.

See also <https://ruby-doc.org/stdlib-2.7.0/libdoc/tempfile/rdoc/Tempfile.html>

### DIFF
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -187,5 +187,39 @@ module Runners
     def show_runtime_versions
       # noop by default
     end
+
+    def read_output_file(file_path)
+      file_path_s = file_path.to_s
+      trace_writer.message "Reading output from #{file_path_s}..."
+      File.read(file_path_s).tap do |output|
+        if output.empty?
+          trace_writer.message "No output"
+        else
+          trace_writer.message output, limit: 24_000 # Prevent timeout
+        end
+      end
+    end
+
+    class InvalidXML < SystemError; end
+
+    def read_output_xml(file_path)
+      output = read_output_file(file_path)
+      REXML::Document.new(output).tap do |document|
+        unless document.root
+          message = "Output XML is invalid from #{file_path}"
+          trace_writer.error message
+          raise InvalidXML, message
+        end
+      end
+    end
+
+    def read_output_json(file_path, &block)
+      output = read_output_file(file_path)
+      if output.empty? && block_given?
+        block.call
+      else
+        JSON.parse(output, symbolize_names: true)
+      end
+    end
   end
 end
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -213,10 +213,10 @@ module Runners
       end
     end
 
-    def read_output_json(file_path, &block)
+    def read_output_json(file_path)
       output = read_output_file(file_path)
       if output.empty? && block_given?
-        block.call
+        yield
       else
         JSON.parse(output, symbolize_names: true)
       end
--- a/lib/runners/processor/eslint.rb
+++ b/lib/runners/processor/eslint.rb
@@ -138,8 +138,8 @@ module Runners
     end
 
     # @see https://eslint.org/docs/developer-guide/working-with-custom-formatters#the-results-object
-    def parse_result(stdout)
-      JSON.parse(stdout, symbolize_names: true).each do |issue|
+    def parse_result(result)
+      result.each do |issue|
         path = relative_path(issue[:filePath])
         # ESLint informs errors as an array if ESLint detects errors in a file.
         issue[:messages].each do |details|
@@ -180,7 +180,7 @@ module Runners
       # NOTE: We must use the `--output-file` option because some plugins may output a non-JSON text to STDOUT.
       #
       # @see https://github.com/typescript-eslint/typescript-eslint/blob/v2.6.0/packages/typescript-estree/src/parser.ts#L237-L247
-      output_file = working_dir / "eslint-output-#{Time.now.to_i}.json"
+      output_file = Tempfile.create(["eslint-", ".json"]).path
 
       _stdout, stderr, status = capture3(
         nodejs_analyzer_bin,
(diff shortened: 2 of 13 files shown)


sig/runners.rbi
```diff
@@ -149,7 +149,9 @@ class Runners::Processor
   def root_dir: -> Pathname
   def directory_traversal_attack?: (String) -> bool
   def show_runtime_versions: -> void
-  def git_blame_info: (String, Integer, Integer) -> Array<GitBlameInfo>
```
ybiquitous: Not exist.

lib/runners/processor.rb
```diff
@@ -187,5 +187,39 @@ def directory_traversal_attack?(path)
...
+          trace_writer.error message
+          raise InvalidXML, message
+        end
+      end
+    end
+
+    def read_output_json(file_path, &block)
+      output = read_output_file(file_path)
+      if output.empty? && block_given?
+        block.call
```
ybiquitous: > __[RuboCop]__ Use `yield` instead of `block.call`. ([view](https://sider.review/gh/repos/202486306/pulls/906#issue-5362611))
>
> <details>
>
> | Rule | Severity |
> | - | - |
> | `Performance/RedundantBlockCall` | convention |
>
> References:
> - <https://github.com/JuanitoFatas/fast-ruby#proccall-and-block-arguments-vs-yieldcode>
> - <https://github.com/rubocop-hq/rubocop-performance/blob/v1.5.2/manual/cops_performance.md#performanceredundantblockcall>
>
> *You can [mark this issue as **closed**](https://sider.review/gh/repos/202486306/pulls/906#issue-5362611) if you should not tackle it. (see our [document](https://help.sider.review/getting-started/working-with-issues) about issues)*
>
> </details>

I will fix.

ybiquitous:
```suggestion
    def read_output_json(file_path)
      output = read_output_file(file_path)
      if output.empty? && block_given?
        yield
```

lib/runners/processor/phpmd.rb
```diff
@@ -79,7 +79,7 @@ def strict
     end
 
     def run_analyzer(changes, targets, rule, options)
-      report_file = Pathname(Tempfile.new("phpmd-report-").path)
+      report_file = Tempfile.new(["phpmd-", ".xml"]).path
```
ybiquitous: > __[Querly]__ Consider using `Tempfile.create` instead of `Tempfile.new`. ([view](https://sider.review/gh/repos/202486306/pulls/906#issue-5362612))
>
> <details>
>
> `Tempfile.create` does **not** automatically remove the created file.
> On the other hand, since `Tempfile.new` removes automatically the file, smoke tests can sometimes fail.
>
> See https://ruby-doc.org/stdlib-2.7.0/libdoc/tempfile/rdoc/Tempfile.html
>
> | Rule |
> | - |
> | `sider.runners.tempfile_new` |
>
> *You can [mark this issue as **closed**](https://sider.review/gh/repos/202486306/pulls/906#issue-5362612) if you should not tackle it. (see our [document](https://help.sider.review/getting-started/working-with-issues) about issues)*
>
> </details>

I will fix.

ybiquitous:
```suggestion
      report_file = Tempfile.create(["phpmd-", ".xml"]).path
```

lib/runners/processor/ktlint.rb
```diff
@@ -87,10 +87,7 @@ def run_gradle
 
       output_file = gradle_config[:output]
       output = if output_file
-                 trace_writer.message "Reading output from #{output_file}..."
-                 (current_dir + output_file).read.tap do |out|
-                   trace_writer.message out
-                 end
+                 read_output_file(output_file)
```
ybiquitous:
```suggestion
                 read_output_file(current_dir / output_file)
```

lib/runners/processor/ktlint.rb
```diff
@@ -105,16 +102,7 @@ def run_maven
...
-      if output_file_path.exist?
-        trace_writer.message "Reading output from #{output_file}..."
-        output = output_file_path.read
-        trace_writer.message output
-      else
-        msg = "#{output_file} does not exist because an unexpected error occurred!"
-        trace_writer.error msg
-        raise msg
-      end
+      read_output_file(current_dir / output_file)
```
ybiquitous:
```suggestion
      output = read_output_file(current_dir / output_file)
```

lib/runners/processor/eslint.rb
```diff
@@ -191,12 +191,7 @@ def run_analyzer(target_dir, additional_options)
         *target_dir
       )
 
-      output_json =
-        if output_file.exist?
-          output_file.read.tap { |json| trace_writer.message json }
-        else
-          nil
-        end
+      output_json = File.file?(output_file) ? read_output_json(output_file) { nil } : nil
```
yykamei, line 194: It would be useful `File.file?` is checked in `read_output_json` or `read_output_file`.

ybiquitous: Good question. At first, I thought calling `File.file?` in `read_output_file`. But I noticed that it is not very meaningful. See the example below:

```ruby
def read_output_file(file)
  if File.file?(file)
    # reading file...
  else
    raise CustomFileNotFound # Should we add the custom error class instead of Errno::ENOENT?
    # Or should we return something?
  end
end
```

As you see the example above, a caller of `read_output_file` should raise an exception or receive a return value which expresses "File not found".

So, to keep it clear the implementation of `read_output_file`, I don't do `File.file?` in the method. That is, if needed, a caller of the method needs to call `File.file?` outside of the method.

What do you think about this decision?

yykamei: It makes sense :+1: Yes, the complexity of the method grows with `File.file?`. I thought it's OK to return `nil`, but as you said, it might be meaningless.

PR #912, opened 2020-03-30
Add the signature for Runners::Shell

yykamei:


sig_new/runners/shell.rbs
```diff
@@ -0,0 +1,45 @@
+type capture3_options = {
+  trace_stdout: bool?,
+  trace_stderr: bool?,
+  trace_command_line: bool?,
+  raise_on_failure: bool?,
+  chdir: Pathname?,
+  is_success: (^(Process::Status) -> bool)?
+}
```
yykamei: Actually, this type seems not to be used because `kwsplat` is not supported so far, but I defined it in order to make it explicit that the methods expect such fields.

```
[Steep 0.14.0] [target=lib] [synthesize:(1:1)] [synthesize:(2:3)] [synthesize:(3:5)] [synthesize:(75:5)] [synthesize:(76:7)] [(::String, *::String, **::capture3_options) -> [::String, ::String, ::Process::Status]] Keyword arg with kwsplat(**) node are not supported.
[Steep 0.14.0] [target=lib] [synthesize:(1:1)] [synthesize:(2:3)] [synthesize:(3:5)] [synthesize:(79:5)] [synthesize:(80:7)] [synthesize:(80:7)] [synthesize:(80:25)] [(::String, *::String, **::capture3_options) -> [::String, ::String, ::Process::Status]] Keyword arg with kwsplat(**) node are not supported.
```

PR #931, opened 2020-04-01
Required constraints for Node.js tools

ybiquitous:
This change requires the default version constraints for Node.js tools.
Actually, all tools have a default version constraints, so the change should have no effects (internal changes).

Also, this change removes the `Runners::Nodejs::Constraint#unsatisfied_by?` method, which can be replaced with more simple `satisfied_by?`.


ybiquitous:
Oh, I missed that this change causes "all default dependencies must have a version constraint". This is not a very good design.

PR #936, opened 2020-04-03
Fix Ignoring to support many files in the `ignore` configuration

### LINKED ISSUE #935: Fix "Argument list too long" error generated with git(1)
I guess this error is attributed to this line.

https://github.com/sider/runners/blob/1ff6d9d89c49539f3547eb94fe114475d984484d/lib/runners/ignoring.rb#L42

I think xargs(1) helps us.

yykamei:
Fix #935


yykamei:
Thank you for your review.

Issue #764, opened 2020-02-25
Drop support of Go Meta Linter, golint, and govet


ybiquitous:
Follow-up #483 and #485.

ybiquitous:
Deadline: **2020-04-30**

sumyapp:
- We will announce this information to our blogs before the end of March
  - also mails

ybiquitous:
Research of the usage:

```
govet and golint: 325 repos (97%)
govet or  golint:   9 repos ( 3%)
```

PR #937, opened 2020-04-03
Use an ssh_key of Runners and delete a smoke test on gometalinter

yykamei:
We deleted the repository sider/go_private_library.
It's not necessary to use a private repository in this test, so, I set a deploy key on sider/runners.
Also, I deleted the smoke test `private_dependency` from `gometalinter`.


ybiquitous:
@yykamei Thanks a lot! Can we fix also the failed smoke test in the same way?

https://github.com/sider/runners/pull/937/checks?check_run_id=557865602#step:8:33

yykamei:
> @yykamei Thanks a lot! Can we fix also the failed smoke test in the same way?

Oh, I just deleted it. What do you think?

ybiquitous:
No problem. 👍

ybiquitous:
It seems better if we could change the PR title. 😏

yykamei:
Thank you. To supplement the reason I deleted the test, we are going to drop the support of Go Meta Linter on #764, so I didn't care about the deprecated smoke test.

PR #954, opened 2020-04-07
Fix Changes#include? to return false for unchanged files

yykamei:
The current implementation always return `true` when the `issue.location` is `nil`. This is wrong, and I fixed the method.

### DIFF
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -32,10 +32,11 @@ module Runners
       gdp = patches # NOTE: This assignment is required for typecheck by Steep
       if gdp
         location = issue.location
-        return true unless location
         patch = gdp.find_patch_by_file(issue.path.to_s)
         if patch && location
           patch.changed_lines.one? { |line| location.start_line == line.number }
+        elsif patch
+          true
         else
           false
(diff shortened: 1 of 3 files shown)


yykamei:
@ybiquitous What do you think about this PR? I think it's a bug.

lib/runners/changes.rb
```diff
@@ -32,10 +32,11 @@ def include?(issue)
       gdp = patches # NOTE: This assignment is required for typecheck by Steep
       if gdp
         location = issue.location
-        return true unless location
         patch = gdp.find_patch_by_file(issue.path.to_s)
         if patch && location
           patch.changed_lines.one? { |line| location.start_line == line.number }
+        elsif patch
```
yykamei, line 38: This means an issue is not for the specific lines but is included in the changed files. The previous implementation always returned `true` without `location` even though an issue indicates other unchanged files.

yykamei:
Thank you. I will add the changelog and merge it.

yykamei:
Oh, `ci skip` didn't get caught. I should have set `skip ci` in the commit :disappointed:

PR #955, opened 2020-04-07
Fix #patches and use ... for git-diff(1)

yykamei:
https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-Comparingbranches

According to this document, comparing branches with `...` means like
this:

> Changes that occurred on the master branch since when the topic branch
> was started off it.

I think using `...` is appropriate for our use case.

### DIFF
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -58,7 +58,7 @@ module Runners
       if base && head
         git_directory.yield_self do |path|
           shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
-          stdout, _ = shell.capture3!("git", "diff", base, head, trace_stdout: false)
+          stdout, _ = shell.capture3!("git", "diff", "#{base}...#{head}", trace_stdout: false)
           GitDiffParser.parse(stdout)
         end
       else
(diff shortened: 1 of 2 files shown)


ybiquitous:
I found another point:

> #### git diff [\<options>] \<commit>...\<commit> [--] [\<path>…​]
> This form is to view the changes on the branch containing and up to the second \<commit>, starting at a common ancestor of both \<commit>. "git diff A...B" is equivalent to "git diff $(git merge-base A B) B". You can omit any one of \<commit>, which has the same effect as using HEAD instead.

-- https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgtltcommitgtltcommitgt--ltpathgt82308203

ybiquitous:
See also https://stackoverflow.com/a/7256391

yykamei:
> See also https://stackoverflow.com/a/7256391

Yes, that's what we wanted!

lib/runners/workspace/git.rb
```diff
@@ -58,7 +58,7 @@ def patches
       if base && head
         git_directory.yield_self do |path|
           shell = Shell.new(current_dir: path, trace_writer: trace_writer, env_hash: {})
-          stdout, _ = shell.capture3!("git", "diff", base, head, trace_stdout: false)
+          stdout, _ = shell.capture3!("git", "diff", "#{base}...#{head}", trace_stdout: false)
```
ybiquitous, line 61: [suggestion] What about adding the following comment to avoid our future misunderstandings?

```suggestion
          # NOTE: We should use `...` (triple-dot) instead of `..` (double-dot). See https://git-scm.com/docs/git-diff
          stdout, _ = shell.capture3!("git", "diff", "#{base}...#{head}", trace_stdout: false)
```

PR #962, opened 2020-04-09
Add `:merge_output` option to `Shell#capture3_trace`

ybiquitous:
Usage:

```ruby
stdout_and_stderr, _, status = shell.capture3_trace(some_command, merge_output: true)
```

I will use this new option for #787.

Note: [`Open3.capture2e`](https://ruby-doc.org/stdlib-2.7.1/libdoc/open3/rdoc/Open3.html#method-c-capture2e) is used internally.

### DIFF
--- a/lib/runners/shell.rb
+++ b/lib/runners/shell.rb
@@ -102,11 +102,18 @@ module Runners
       chdir = options[:chdir] || current_dir
       is_success = options.fetch(:is_success) { ->(status) { status.success? } }
       stdin_data = options.fetch(:stdin_data, nil)
+      merge_output = options.fetch(:merge_output, false)
 
       command_line = [command] + args
       trace_writer.command_line(command_line) if trace_command_line
 
-      Open3.capture3(env_hash, command, *args, chdir: chdir.to_s, stdin_data: stdin_data).tap do |stdout_str, stderr_str, status|
+      method = merge_output ? :capture2e : :capture3
+      Open3.public_send(method, env_hash, command, *args, chdir: chdir.to_s, stdin_data: stdin_data).then do |stdout_str, stderr_str, status|
+        if merge_output
+          status = stderr_str
+          stderr_str = nil
+        end
+
         trace_writer.stdout stdout_str if trace_stdout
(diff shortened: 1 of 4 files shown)


ybiquitous:
@yykamei Could you please give me feedback?

ybiquitous:
> I think updating the signature of `Runners::Shell#capture3_trace` is necessary because `stderr` might be `nil`.

Yes, but I have no idea to realize with fewer duplications. Do you have somethin?

yykamei:
:grey_question: My point is changing the signature like this. Is this unnecessary?

```diff
diff --git a/sig/runners.rbi b/sig/runners.rbi
index 813aabd..347f05b 100644
--- a/sig/runners.rbi
+++ b/sig/runners.rbi
@@ -163,10 +163,10 @@ class Runners::Shell

   def initialize: (current_dir: Pathname, env_hash: Hash<String, String?>, trace_writer: TraceWriter) -> any

-  def capture3: (String, *String, **capture3_options) -> [String, String, Process::Status]
-  def capture3!: (String, *String, **capture3_options) -> [String, String]
-  def capture3_with_retry!: (String, *String, ?tries: Integer) -> [String, String]
-  def capture3_trace: (String, *String, **capture3_options) -> [String, String, Process::Status]
+  def capture3: (String, *String, **capture3_options) -> [String, String?, Process::Status]
+  def capture3!: (String, *String, **capture3_options) -> [String, String?]
+  def capture3_with_retry!: (String, *String, ?tries: Integer) -> [String, String?]
+  def capture3_trace: (String, *String, **capture3_options) -> [String, String?, Process::Status]

   def push_dir: <'x> (Pathname) { -> 'x } -> 'x
   def current_dir: () -> Pathname
```

ybiquitous:
Oh, I misunderstood. 😓 Thanks, I will fix it.

ybiquitous:
Or, it seems that we can return `""` (empty string) instead of `nil`. This can make type-checking easy. What do you think?

ybiquitous:
Fixed via 4333caa

PR #963, opened 2020-04-09
[Refactor] Use `File::FNM_EXTGLOB` for `Changes#deletable?`

ybiquitous:
- https://ruby-doc.org/core-2.7.1/File.html#method-c-fnmatch-3F
- https://ruby-doc.org/core-2.7.1/File/Constants.html


ybiquitous:
@yykamei Could you please review if you have time, just in case?

ybiquitous:
Thanks! 😺

PR #965, opened 2020-04-13
[Test] Use `amazing_print` instead of `rainbow` for smoke test's colored output

ybiquitous:
[amazing_print](https://github.com/amazing-print/amazing_print) is a fork of [awesome_print](https://github.com/awesome-print/awesome_print).
awesome_print is no longer maintained and does not support Ruby 2.7.

See also #908

image


ybiquitous:
The CI passed! I'll merge.

PR #966, opened 2020-04-13
Do not run releasing action on pull requests

ybiquitous:
This action is meaningless on pull requests.

See also <softprops/action-gh-release#-usage>


ybiquitous:
Let's try!

PR #971, opened 2020-04-15
[FxCop] roslyn-analyzers-runner provides a static code analysis without running build

suetsuguk:
- [x] Write a [document](https://github.com/sider/sider-docs) for the new runner. PR sider/sider-docs#359


ybiquitous:
@suetsuguk Thank you for the new pull request! Could you please update this PR description as follows?

- Remove `Fix #<ISSUE_ID>` if a related issue is not found.
- Fill out all the checkboxes.

ybiquitous:
Also, if you are working in progress, please add `WIP` to the PR title or convert it to draft.

lib/runners/processor/fxcop.rb
```diff
@@ -20,68 +16,46 @@ def analyzer_version
...
+          .changed_paths
+          .select{|p| p.extname.eql?(".cs")}
+          .map{|p| relative_path(working_dir / p, from: current_dir)}
+          .map(&:to_s))
 
-      # generate a result instance
-      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
-        parse_result(output_json) do |issue|
-          result.add_issue(issue)
+      if status.exited? && status.exitstatus == 0
```
suetsuguk, line 29: > __[Querly]__ Process#exitstatus may be nil ([view](https://sider.review/gh/repos/202486306/pulls/971#issue-5460754))
>
> <details>
>
>
>
> | Rule |
> | - |
>
> *You can [mark this issue as **closed**](https://sider.review/gh/repos/202486306/pulls/971#issue-5460754) if you should not tackle it. (see our [document](https://help.sider.review/getting-started/working-with-issues) about issues)*
>
> </details>

Already checked the exited? flag as suggested in the detailed message.

ybiquitous: This rule tries to find such an error like `status.exitstatus.to_i`. If you think no problem, please close the issue on Sider.

Also, in this case, `if status.success?` may be better:

https://ruby-doc.org/core-2.6.6/Process/Status.html#method-i-success-3F

suetsuguk: GOTCHA
As seen from the source code of `success?`, it is appropriate in this case.

suetsuguk: Fixed.
619bd95545d9b944d5562bf10e1789f6d024806c

test/smokes/fxcop/expectations.rb
```diff
@@ -159,43 +129,37 @@
...
+  --outputfile         (Default: -)
+
+  --help               Display this help screen.
+
+  --version            Display version information.
+
+  file ... (pos. 0)    
+
+EOS
+    analyzer: ANALYZER
```
yykamei, line 163: This is an interesting smoke test. Of course, a file like `--dummy.cs` is rarely used, and there is something wrong in a project which contains such a file. However, I think Runners should analyze the code `--dummy.cs`. Maybe, some projects really use the `--dummy.cs`.

lib/runners/processor/fxcop.rb
```diff
@@ -20,68 +16,46 @@ def analyzer_version
...
-      # generate a result instance
-      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
-        parse_result(output_json) do |issue|
-          result.add_issue(issue)
+      if status.success?
+        Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+          construct_result(result, read_output_json(output_file))
         end
+      else
+        Results::Failure.new(guid: guid, analyzer: analyzer, message: err)
```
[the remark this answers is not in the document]
suetsuguk: Currently `Sider.RoslynAnalyzersRunner` returns the code 1 when runtime error occurred (e.g. a dependent dll was not found), and the code 2 when an illegal command-line argument passed. Can I treat these cases as the `Failure`? Be noted that no configuration file nor user preference is available so far. Any error by incompleteness in user configuration shall not be occurred.

lib/runners/processor/fxcop.rb
```diff
@@ -1,16 +1,12 @@
...
-    RULE_ID_PATTERN = /CA[0-9]+/.freeze
-
     Schema = StrongJSON.new do
       let :runner_config, Schema::BaseConfig.base
 
       let :issue, object(
-        severity: string?,
+        category: string,
+        description: string,
+        severity: integer,
```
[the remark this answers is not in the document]
suetsuguk: Agreed. I will update `Sider.RoslynAnalyzersRunner` so that it returns text representations rather than numbers.

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
yykamei: `Error` is proper for the case of runtime error. On the other hand, it depends on the situation regarding an illegal command-line argument. I think the `exitstatus 2` should also be handled as `Error` because the command line argument is passed by Runners.

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
yykamei: Awesome :+1:

test/smokes/fxcop/expectations.rb
suetsuguk: Can you tell me how to pass a command-line argument to a test case? I want to test an error case. Runtime error is another candidate, but it would be more difficult to cause such an error, though.

images/fxcop/Dockerfile.erb
```diff
@@ -1,14 +1,27 @@
...
 # cache target analyzer
 USER $RUNNER_USER
 RUN cd && \
     mkdir dummy_project && \
     cd dummy_project && \
     dotnet new console && \
+    dotnet add package Microsoft.CodeAnalysis.Analyzers --version ${FXCOP_VERSION} && \
     dotnet add package Microsoft.CodeAnalysis.FxCopAnalyzers --version ${FXCOP_VERSION} && \
     cd && \
     rm -rf dummy_project
```
ybiquitous, line 27: [suggestion] What about the following improvements?

- Use always the non-root user.
- Reduce `RUN`.

```dockerfile
FROM sider/devon_rex_dotnet:2.16.1

<%= render_erb 'images/Dockerfile.base.erb' %>

ENV FXCOP_VERSION=2.9.8

ARG ROSLYN_ANALYZERS_RUNNER_VERSION=0.1.2
ARG ROSLYN_ANALYZERS_RUNNER_HOME=${RUNNER_USER_HOME}/roslyn-analyzers-runner
ENV PATH ${ROSLYN_ANALYZERS_RUNNER_HOME}:$PATH

COPY --chown=<%= chown %> images/fxcop/analyzers.json ${RUNNER_USER_HOME}/

RUN mkdir -p ${ROSLYN_ANALYZERS_RUNNER_HOME} && \
    curl -SL https://github.com/sider/roslyn-analyzers-runner/releases/download/v${ROSLYN_ANALYZERS_RUNNER_VERSION}/Sider.RoslynAnalyzersRunner.${ROSLYN_ANALYZERS_RUNNER_VERSION}.linux-x64.tar.gz \
      | tar -zxC ${ROSLYN_ANALYZERS_RUNNER_HOME} && \
    mv ${RUNNER_USER_HOME}/analyzers.json ${ROSLYN_ANALYZERS_RUNNER_HOME}/ && \
    sed -i -e"s/\${FXCOP_VERSION}/${FXCOP_VERSION}/" ${ROSLYN_ANALYZERS_RUNNER_HOME}/analyzers.json && \
    cd ${RUNNER_USER_HOME} && \
    mkdir dummy_project && \
    cd dummy_project && \
    dotnet new console && \
    dotnet add package Microsoft.CodeAnalysis.Analyzers --version ${FXCOP_VERSION} && \
    dotnet add package Microsoft.CodeAnalysis.FxCopAnalyzers --version ${FXCOP_VERSION} && \
    cd .. && \
    rm -rf dummy_project

<%= render_erb 'images/Dockerfile.end.erb' %>
```

test/smokes/fxcop/expectations.rb
yykamei: In my opinion, this test is not required for Runners because users cannot control how to pass arguments to `Sider.RoslynAnalyzersRunner`. Instead, all users can do is pass parameters, not command-line arguments. If the command-line argument error happens, this is a bug of Runners.

images/fxcop/Dockerfile.erb
suetsuguk: Agreed. That could be better.

test/smokes/fxcop/expectations.rb
suetsuguk: Let me back to square one. I just wanted to make an error-case test for the code block below:
```
      else
        Results::Failure.new(guid: guid, analyzer: analyzer, message: err)
      end
````
As discussed before, this `else` block runs only when runtime error occurred or illegal argument passed. But it is hard to cause runtime error. Instead, I came up with passing an illegal argument to run the 'else' block. My intention was to test an error case, not a test for arguments.

Anyway, I will omit this test case in order to avoid confusion.

Comment (yykamei) [reply test/smokes/fxcop/expectations.rb:163]: As I said [here](https://github.com/sider/runners/pull/971#discussion_r410007771), the `exitstatus` of `1` and `2` should be handled as `Error`. Honestly, I'm not sure about `Sider.RoslynAnalyzersRunner`, but if all exit codes are not caused by users, Runners should return `Error`. In this case, using `capture3!` is better here, and you don't have to care about `status`.

https://github.com/sider/runners/pull/971/files/e46b8f9996c4a380ec8e0005b1e7c39ca1fd9e9c#diff-f581c0ae2875a90c7ed089156e3e91cdR21

File: test/smokes/fxcop/expectations.rb [hunk 6]
```diff
@@ -1,37 +1,63 @@
...
-        message: 'Parameter args of method Main is never used. Remove the parameter or use it in the method body.',
-        object: {
-          severity: 'warning'
-        },
-        git_blame_info: nil,
-        links: ["https://docs.microsoft.com/visualstudio/code-quality/ca1801-review-unused-parameters"]
-      }
-    ],
-    analyzer: { name: 'FxCop', version: '2.9.8' }
+    issues: ISSUES_PROGRAM_CS,
```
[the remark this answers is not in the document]
suetsuguk: When I think carefully I found that only two test case are necessary and needed; the normal case 'success' and the 'success_csproj_in_non_root_dir' in which a sub-directory is located. I will omit others.

lib/runners/processor/fxcop.rb
```diff
@@ -20,68 +16,46 @@ def analyzer_version
...
-      # generate a result instance
-      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
-        parse_result(output_json) do |issue|
-          result.add_issue(issue)
+      if status.success?
+        Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+          construct_result(result, read_output_json(output_file))
         end
+      else
+        raise err
```
yykamei, line 34: The related comment is [here](https://github.com/sider/runners/pull/971#discussion_r410059303), but you don't have to care about `status`. Instead, it's better fo use `capture3!` [here](https://github.com/sider/runners/pull/971/files#diff-f581c0ae2875a90c7ed089156e3e91cdR21).

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
suetsuguk: Fixed.
commit 8f5eb013df27cc593dbf16ed1311f2bbefaaa152

lib/runners/processor/fxcop.rb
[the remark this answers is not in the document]
suetsuguk: Fixed.
commit d535928f420032c349dce8164488d1bf06a88113

test/smokes/fxcop/expectations.rb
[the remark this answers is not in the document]
suetsuguk: Fixed.
commit b0a9c3658e9c5e89389638173c8f9ac29149d0c4

test/smokes/fxcop/expectations.rb
suetsuguk: Fixed.
commit 8f5eb013df27cc593dbf16ed1311f2bbefaaa152 and commit b0a9c3658e9c5e89389638173c8f9ac29149d0c4

images/fxcop/Dockerfile.erb
suetsuguk: Fixed.
commit a7e066e98d9443cba2e826640e343274a9454489

lib/runners/processor/fxcop.rb
```diff
@@ -20,68 +16,46 @@ def analyzer_version
...
-      # run 'dotnet build' with static analysis module
       output_file = Tempfile.create(["fxcop-", ""]).path
-      capture3!('dotnet', 'build', '--no-incremental', "-property:errorlog=#{output_file}")
-      output_json = read_output_json(output_file)
+      _, err, status = capture3('Sider.RoslynAnalyzersRunner',
+        '--outputfile', output_file,
+        *changes
+          .changed_paths
+          .select{|p| p.extname.eql?(".cs")}
+          .map{|p| relative_path(working_dir / p, from: current_dir)}
```
yykamei, line 26: Currently,  I cannot run smoke tests on my machine, but I think this `working_dir` seems unnecessary. What do you think?

```suggestion
          .map{ |path| relative_path(path, from: current_dir) }
```

lib/runners/processor/fxcop.rb
suetsuguk: It seems that the `capture3!` method traces the stdout and stderr text, and raises an exception if needed.
So, we do not need to handle the stderr text when we use the `capture3!`. The code is going to be as follows;
```
capture3('Sider.RoslynAnalyzersRunner', ...)

  construct_result(result, read_output_json(output_file))
end
```
There is no need to write `if status.success?` before `Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|`.

Does that make sense?

lib/runners/processor/fxcop.rb
suetsuguk: Let me take an example.
source file: /a/b/c/project/src/hello.cs
the project file directory which contains sider.yml: /a/b/c/project

If the `root_dir` is set to `src` in the sider.yml located in `/a/b/c/project/`, the `current_dir` is set to `/a/b/c/project/src`.
On the other hand, the `changes.changed_paths` file is still `src/hello.cs` which is relative to the `working_dir` (e.g. /a/b/c/project).
I need to pass "hello.cs" to Sider.RoslynAnalyzersRunner because it runs on the `current_dir` (/a/b/c/project/src).
So I normalized the path as follows:
relative_path(working_dir / p, from: current_dir)
->
relative_path("/a/b/c/project" / "src/hello.cs", from: "/a/b/c/project/src") leads to "hello.cs".

lib/runners/processor/fxcop.rb
yykamei: Yes, I think so. Neither `stderr` nor `status` could not have to be considered in this case.

lib/runners/processor/fxcop.rb
yykamei: Yes, I know the parameter of `from` is necessary, but passing the path prefixed `working_dir` seems unnecessary because `p` is already an instance of `Pathname`.

lib/runners/processor/fxcop.rb
suetsuguk: Good suggestion. Thank you.
Fixed. commit 75ce9a5c08821b46a7248764fd1af3005a1ccca3

lib/runners/processor/fxcop.rb
suetsuguk: Although `p` is an instance of the Pathname class, it has a relative path to the `working_dir`. I just wanted to obtain a relative path to the `current_dir`, so I needed to expand `p` into the absolute path by adding the prefix `working_dir` before splitting into `current_dir` and its relative path. For example, `relative_path("/a/b/c/project" / "src/bar/hello.cs", from: "/a/b/c/project/src")` leads to "bar/hello.cs".

```
changes.rb
def self.calculate(base_dir:, head_dir:, working_dir:, patches:)
  relative_path = working_path.relative_path_from(working_dir)
  changed_paths << relative_path

processor.rb
def relative_path(original, from: working_dir)
  path = Pathname(original)
  if path.relative?
    path = (current_dir / path).cleanpath
  path.relative_path_from(from)
```

yykamei: :thinking: Whichever the `p` is a relative path or not, using `from` of `Processor#relative_path` seems to resolve your demand.

I created a test `test_relative_path` and get the succeeded result.... Do I misunderstand?

```diff
diff --git a/test/processor_test.rb b/test/processor_test.rb
index 3d3ba11..5cf7f70 100644
--- a/test/processor_test.rb
+++ b/test/processor_test.rb
@@ -31,6 +31,20 @@ class ProcessorTest < Minitest::Test
     )
   end

+  def test_relative_path
+    with_workspace do |workspace|
+      working_dir = workspace.working_dir
+      (working_dir / "src").mkdir
+      (working_dir / "src/abc.cs").write("abc")
+      (working_dir / "src/foo.cs").write("foo")
+
+      current_dir = (working_dir / "src").realpath
+      processor = new_processor(workspace: workspace)
+      assert_equal "src/abc.cs", processor.relative_path(Pathname("src/abc.cs"), from: working_dir).to_path # <-- This line is the case `root_dir` is set as `"src"`
+      assert_equal "abc.cs", processor.relative_path(Pathname("src/abc.cs"), from: current_dir).to_path
+    end
+  end
+
   def test_capture3_env_setup
     with_workspace do |workspace|
       mock(Open3).capture3({"RUBYOPT" => nil, "GIT_SSH" => (workspace.working_dir / "id_rsa").to_s},
```

yykamei: :thinking: Whichever the `p` is a relative path or not, using `from` of `Processor#relative_path` seems to resolve your demand.

I created a test `test_relative_path` and get the succeeded result.... Do I misunderstand?

```diff
diff --git a/test/processor_test.rb b/test/processor_test.rb
index 3d3ba11..5cf7f70 100644
--- a/test/processor_test.rb
+++ b/test/processor_test.rb
@@ -31,6 +31,20 @@ class ProcessorTest < Minitest::Test
     )
   end

+  def test_relative_path
+    with_workspace do |workspace|
+      working_dir = workspace.working_dir
+      (working_dir / "src").mkdir
+      (working_dir / "src/abc.cs").write("abc")
+      (working_dir / "src/foo.cs").write("foo")
+
+      current_dir = (working_dir / "src").realpath
+      processor = new_processor(workspace: workspace)
+      assert_equal "src/abc.cs", processor.relative_path(Pathname("src/abc.cs"), from: working_dir).to_path # <-- This line is the case `root_dir` is set as `"src"`
+      assert_equal "abc.cs", processor.relative_path(Pathname("src/abc.cs"), from: current_dir).to_path
+    end
+  end
+
   def test_capture3_env_setup
     with_workspace do |workspace|
       mock(Open3).capture3({"RUBYOPT" => nil, "GIT_SSH" => (workspace.working_dir / "id_rsa").to_s},
```

test/smokes/fxcop/expectations.rb
```diff
@@ -10,96 +10,43 @@
     issues: [
```
ybiquitous, line 10: @suetsuguk I've refactored `Smoke.add_test` via PR #989, so could you please merge the `master` branch and fix this `test/smokes/fxcop/expectations.rb` file following by the latest `Smoke.add_test`?

lib/runners/processor/fxcop.rb
suetsuguk: To update the `current_dir` value in the `processor` instance, you need to call the `push_dir` method. Try the code below, you will get assertion errors.
```
      processor = new_processor(workspace: workspace)
      processor.push_dir(current_dir) do
        assert_equal "src/abc.cs", ...
        assert_equal "abc.cs", ...
      end
```

yykamei: OK. I understood. You're right. Sorry to bother you.

test/smokes/fxcop/expectations.rb
suetsuguk: Merged from master and fixed tests accordingly.
commit 84b9bf01848f430ec528ed0a9efde1041515341c
commit b730ff3d16d86a515f6393ec0dbfb9a38fdf84f4
commit a502fea46110771d156524ed7a0858fe218939e3

lib/runners/processor/fxcop.rb
```diff
@@ -20,68 +16,42 @@ def analyzer_version
...
       output_file = Tempfile.create(["fxcop-", ""]).path
-      capture3!('dotnet', 'build', '--no-incremental', "-property:errorlog=#{output_file}")
-      output_json = read_output_json(output_file)
+      capture3!('Sider.RoslynAnalyzersRunner',
+        '--outputfile', output_file,
+        *changes
+          .changed_paths
+          .select{|p| p.extname.eql?(".cs")}
+          .map{|p| relative_path(working_dir / p, from: current_dir)}
+          .map(&:to_s))
```
ybiquitous, line 27: [suggestion] What about the following refactoring idea?

```suggestion
          .filter_map{|f| relative_path(working_dir / f, from: current_dir).to_s if f.extname == ".cs"})
```

- [`filter_map`](https://ruby-doc.org/core-2.7.1/Enumerable.html#method-i-filter_map) is useful in this case.
- Use `f` instead of the confusing [`p`](https://ruby-doc.org/core-2.7.1/Kernel.html#method-i-p) method.
- Use `==` instead of [`eql?`](https://ruby-doc.org/core-2.7.1/Object.html#method-i-eql-3F), because the former is used usually (both is different).

ybiquitous:
@suetsuguk Can I merge this? Are you working on this yet?

suetsuguk:
There is nothing left for me. You can merge it to master when it is convenient for you.

ybiquitous:
Thank you so much! 😄

PR #975, opened 2020-04-17
Improve performance for `Changes` class

ybiquitous:
`Changes#include?` is called in the following loop:

https://github.com/sider/runners/blob/ba501d8c1423cdec467df1ac3b5572e508d28a15/lib/runners/results.rb#L64-L68

This avoids the creation of a `Set` instance on each loop iteration.

### DIFF
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -6,9 +6,9 @@ module Runners
     attr_reader :patches
 
     def initialize(changed_paths:, unchanged_paths:, untracked_paths:, patches:)
-      @changed_paths = changed_paths
-      @unchanged_paths = unchanged_paths
-      @untracked_paths = untracked_paths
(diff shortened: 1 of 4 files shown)


lib/runners/changes.rb
```diff
@@ -6,16 +6,17 @@ class Changes
...
+      @unchanged_paths = Set.new(unchanged_paths).freeze
+      @untracked_paths = Set.new(untracked_paths).freeze
       @patches = patches
     end
 
     def delete_unchanged(dir:, except: [], only: [])
       unchanged_paths
         .filter { |file| deletable?(dir, file, except, only) }
         .each { |file| dir.join(file).delete }
+        .to_a
```
ybiquitous, line 19: [note] `delete_unchanged` is usually called once. 👍

ybiquitous:
Thanks! 🥰

PR #997, opened 2020-04-23
Apply Prettier for FxCop smoke test code

ybiquitous:
To reduce diffs on other PRs, see #922 for example.


ybiquitous:
The CI passed!

PR #998, opened 2020-04-23
[SwiftLint] Improve Docker image

ybiquitous:
- Do not use the `root` user.
- Use `curl` instead of `git clone` to download the source code.

### DIFF
--- a/images/swiftlint/Dockerfile
+++ b/images/swiftlint/Dockerfile
@@ -6,15 +6,16 @@ FROM sider/devon_rex_swift:master
 
 ARG SWIFTLINT_VERSION=0.39.2
 
+ENV PATH ${RUNNER_USER_HOME}/bin:${PATH}
+
 # Build SwiftLint
-USER root
-RUN SWIFTLINT_BUILD_DIR=/tmp/swiftlint_build && \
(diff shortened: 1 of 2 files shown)


images/swiftlint/Dockerfile
```diff
@@ -6,15 +6,16 @@ FROM sider/devon_rex_swift:master
...
-    git clone --branch ${SWIFTLINT_VERSION} --single-branch https://github.com/realm/SwiftLint.git ${SWIFTLINT_BUILD_DIR} && \
-    cd ${SWIFTLINT_BUILD_DIR} && \
-    make install && \
-    swiftlint version && \
-    rm -rf ${SWIFTLINT_BUILD_DIR}
-USER $RUNNER_USER
+RUN cd /tmp && \
+    curl -sSL --compressed https://github.com/realm/SwiftLint/archive/${SWIFTLINT_VERSION}.tar.gz | tar -xz && \
+    cd SwiftLint-${SWIFTLINT_VERSION} && \
+    make install BINARIES_FOLDER=${RUNNER_USER_HOME}/bin && \
```
ybiquitous, line 15: [note] `BINARIES_FOLDER` is used in `Makefile`:

https://github.com/realm/SwiftLint/blob/b1c72069cafa6cdebfb9274806cd976c54d8dde5/Makefile#L89-L91

PR #661, opened 2020-01-11
Support GolangCI-Lint

mpls104:
## Issue

Fix #484 

## TODOs

- [x]  Need  to handle some validation pattern.
https://github.com/golangci/golangci-lint/blob/master/pkg/lint/lintersdb/validator.go
- [x] nolint
- [x] Add warning pattern
- [x] Add no config file detected pattern
- [x] Add pattern to detect error
- ~[ ] Add links~
- [x] Add tests option
- [x] no-config  option
- [x] skip-dirs
- [x] skip-dirs-use-default 
- [x] skip-files
- [x] presets
- [x] target 
- [x] fix expectation.rb
- [x] hundle prerequisite error
- [x] list not supported options
- [x] List linters convenient and disable in default 
- [x] refactor duplicated code

## Omitted option
- `--enable-all`  because it's deprecated
-  `--modules-download-mode` no need for customer since linter will be executed on Docker
- `--build-tags` no meaning
- `--timeout` Sider already provides this function
- `--print-resources-usage` no meaning
- `--max-same-issues`, `--max-issues-per-linter` didn't come up the usecase
- `--new`, `--new-from-patch`, `--new-from-rev` I suppose Sider provides this funciton
- `--fix` can not support

## Docs
https://github.com/sider/sider-docs/pull/290

PR #1001, opened 2020-04-24
[GolangCI-Lint] Change default linters

ybiquitous:
This change aims to make it easy to migrate from Golint to GolangCI-Lint.

Before: [stylecheck](https://github.com/dominikh/go-tools/tree/master/stylecheck)
After: [golint](https://github.com/golang/lint)

Surely, stylecheck is a replacement for golint and adds some extra rules.
But stylecheck is a 3rd-party tool, on the other hand, golint is a standard tool.
I judge that the standard tool is more suitable as a default.

See also https://github.com/sider/runners/pull/661#issuecomment-583803958

### DIFF
--- a/images/golangci_lint/sider_golangci.yml
+++ b/images/golangci_lint/sider_golangci.yml
@@ -4,9 +4,10 @@ linters-settings:
       - diagnostic
       - performance
       - style
+
 linters:
   enable:
     - bodyclose
-    - gocyclo
-    - stylecheck
     - gocritic
+    - gocyclo
+    - golint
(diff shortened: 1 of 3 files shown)


images/golangci_lint/sider_golangci.yml
```diff
@@ -4,9 +4,10 @@ linters-settings:
...
       - style
+
 linters:
   enable:
     - bodyclose
-    - gocyclo
-    - stylecheck
     - gocritic
+    - gocyclo
+    - golint
```
ybiquitous, line 13: [note] Sort alphabetically.

PR #1007, opened 2020-04-27
Run smoke tests in processors instead of threads

ybiquitous:
SwiftLint's smoke test on CI is unstable, so this change tries to reduce the failures by running the test in 1 processor.

### DIFF
--- a/lib/runners/testing/smoke.rb
+++ b/lib/runners/testing/smoke.rb
@@ -34,8 +34,7 @@ module Runners
       def run
         load expectations.to_s
 
-        threads = ENV["JOBS"]&.to_i || Parallel.processor_count * 2
-        results = Parallel.map(self.class.tests, in_threads: threads) do |name, pattern|
+        results = Parallel.map(self.class.tests, in_processes: ENV["JOBS"]&.to_i) do |name, pattern|
           out = StringIO.new(''.dup)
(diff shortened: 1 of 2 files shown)


lib/runners/testing/smoke.rb
```diff
@@ -34,8 +34,7 @@ def initialize(argv)
       def run
         load expectations.to_s
 
-        threads = ENV["JOBS"]&.to_i || Parallel.processor_count * 2
-        results = Parallel.map(self.class.tests, in_threads: threads) do |name, pattern|
+        results = Parallel.map(self.class.tests, in_processes: ENV["JOBS"]&.to_i) do |name, pattern|
```
ybiquitous, line 37: [note] If `JOBS` is unset, `Parallel.processor_count` is used instead.

See also: https://github.com/grosser/parallel/blob/34cb10e875ff6f6c7069dd5a900a02ee89b9f968/lib/parallel.rb#L259

ybiquitous:
It seems that this change does not raise the CI delay. 👍

ybiquitous:
I guess the cause of the instability: Open3 is using Thread internally.

https://github.com/ruby/ruby/blob/aca935a67042dd7b82e34844ae6139cedd66339d/lib/open3.rb#L287-L288

PR #1009, opened 2020-04-27
Refactor Dockerfile: ShellCheck, SwiftLint, and Misspell

ybiquitous:
`~/bin` is set to PATH already. See https://github.com/sider/devon_rex/pull/196


images/shellcheck/Dockerfile
```diff
@@ -9,8 +9,8 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends file && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
+USER $RUNNER_USER
```
ybiquitous, line 12: [note] Restore the user from root.

Issue #1000, opened 2020-04-24
Support third-party dependencies for JVM tools


yykamei:
According to the [document](https://checkstyle.sourceforge.io/config.html), Checkstyle can get written arbitrary modules in its configuration.

Some of our customers want its feature, but I'm not sure how to implement it.

ybiquitous:
Perhaps, I think we can do it by setting the `CLASSPATH` environment variable.

ybiquitous:
E.g.

```yaml
linter:
  checkstyle:
    classpath:
      - custom-rules/*
      - https://foo/custom-rules.jar # ???
```

yykamei:
Also, it might be better to support build tools, such as Gradle.

ybiquitous:
Oh, I don't want to support Maven or Gradle until we will really need them...
The implementations and test cases would be too complex.

ybiquitous:
I found the document for specifying classpath:
https://checkstyle.sourceforge.io/cmdline.html#Usage_by_Classpath_update

<img width="1157" alt="image" src="https://user-images.githubusercontent.com/473530/80178504-55440380-8639-11ea-8005-cd2898ceb685.png">

ybiquitous:
We need to consider:

- a new option that can support all Java and JVM tools (also Kotlin)
- a new option that can download JAR files from the network.
- what the standard format is, e.g. Maven? (`linter.*.maven` like `linter.*.gems`)

ybiquitous:
One suggestion:

```yaml
linter:
  checkstyle:
    # Download from the maven repository
    maven:
      # short version (like Gradle format) - {groupId}:{artifactId}:{version}
      - org.wildfly.checkstyle:wildfly-checkstyle-config:1.0.8.Final
      # long version
      - groupId: org.wildfly.checkstyle
        artifactId: wildfly-checkstyle-config
        version: 1.0.8.Final

    # Set `CLASSPATH`
    classpath:
      # relative path
      - custom-rules/foo.jar
      # glob
      - "custom-rules/**/*.jar"
      # remote (e.g. via `curl`)
      - https://foo.com/custom-rules.jar
```

ybiquitous:
See also https://en.wikipedia.org/wiki/Classpath

yykamei:
As long as `classpath` supports the remote URL, it meets the requirement, I guess.

ybiquitous:
If we need not only `jvm_deps` but also `classpath`, let's open a new issue.

PR #1035, opened 2020-04-30
[Checkstyle] Shorten official issue ID

ybiquitous:
I think we no longer need to distinguish issue IDs in detail.
Too long IDs make it hard for users to read issues.

- Remove the message hash.
- Remove the rule namespace.

For example:

```diff
-com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck#a134cc
+LineLengthCheck
```

Note that this change limits to the **official** rules.
We will support for users to specify third-party rules, via the issue #1000.


ybiquitous:

ybiquitous:
Thank you for your review! 😄

PR #1039, opened 2020-04-30
[remark-lint] Use `remark-cli` as main package

ybiquitous:
This change aims to make it easier to use this runner by making the `remark-lint` package optional.

The `remark-lint` package is actually just to provide the `<!--lint ignore-->` feature,
and it is just a plugin of `remark`.

https://github.com/remarkjs/remark-lint/blob/316c8c698227746787481cd12bd8cb487f15fae3/packages/remark-lint/index.js

If the `remark-cli` package is installed, it's enough for analysis.
Note that, as a side-effect of this change, the remark-lint runner's version will be changed significantly.
(that is, we need to fix also [our document](https://help.sider.review/tools/markdown/remark-lint))


ybiquitous:
@yykamei Thank you for your review!

PR #1040, opened 2020-04-30
[Refactor] Simplify Node.js runners

ybiquitous:
This change aims to reduce the code of Node.js runners and to make our maintenance easier.

The current check for default dependencies is too strict and not useful.
Because the same thing is ensured by smoke tests.

I think that "dependency constraints" and smoke tests is enough to ensure the runner behavior.


ybiquitous:
@yykamei What do you think?

PR #1045, opened 2020-05-07
[Misspell] Remove excluding files safely

ybiquitous:
Currently when trying to delete a file or directory that does not exist, `Errno::ENOENT` error occurs.
For example:

```yaml
linters:
  misspell:
    exclude: ["not_exist/file.rb"]
```

I think this behavior is too strict, so this change is to use "force deleting".

See also:
- https://ruby-doc.org/stdlib-2.7.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-remove_entry_secure
- https://ruby-doc.org/stdlib-2.7.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-rm_rf


ybiquitous:
@yykamei Could you please review this?

ybiquitous:
Make sense. 👍 
It seems there are few users who are confused by this error, so I'll close this PR. No problem?

yykamei:
No problem :+1:

PR #1050, opened 2020-05-07
[Misspell] Some improvements

ybiquitous:
I believe that the following changes can improve the usability:

- Use a correct word as an issue ID instead of an issue message.
- Save an issue's metadata: a correct word and an incorrect word.

Also, this change makes refactoring to make the code readable a bit.


ybiquitous:

ybiquitous:
Thank you!

yykamei:
By the way, `Tempfile.create` typically seems the same pattern among analyzers.

ybiquitous:
Yes, do you have any ideas?

ybiquitous:
For example, how about this? ⬇️ 

```ruby
class Processor
  def report_file(ext: ".txt")
    Tempfile.create(["#{analyzer_id}_", ext]).path
  end
end
```

yykamei:
:clap: Great, I like it.

ybiquitous:
Thanks, I'll try it. 💪

PR #1053, opened 2020-05-07
[Refactor] Replace ActiveSupport's `delegate` with Forwardable

ybiquitous:
`Forwardable` is a standard library. See the API doc:
https://ruby-doc.org/stdlib-2.7.1/libdoc/forwardable/rdoc/Forwardable.html


sig/polyfills/forwardable.rbi
```diff
@@ -0,0 +1,4 @@
+module Forwardable
+  def def_instance_delegators: (Symbol, *Symbol) -> void
+  alias def_delegators def_instance_delegators
+end
```
ybiquitous, line 4: https://github.com/ruby/ruby/blob/f38c641994265a87a0380752dbc670937094be42/lib/forwardable.rb

ybiquitous:
CI passed! 🎉

PR #1054, opened 2020-05-07
Remove ActiveSupport

ybiquitous:
Purpose:

- Make our maintenance easier.
- Extensions (monkey-patch) of built-in methods are often confusing.
- Reduce polyfills for type signatures.


lib/runners/cli.rb
```diff
@@ -33,7 +33,7 @@ def with_working_dir(&block)
     end
 
     def processor_class
-      @processor_class ||= (Processor.subclasses.detect do |cls|
+      @processor_class ||= (ObjectSpace.each_object(Class).filter { _1 < Processor }.detect do |cls|
```
ybiquitous, line 36: https://github.com/rails/rails/blob/b738f1930f3c82f51741ef7241c1fee691d7deb2/activesupport/lib/active_support/core_ext/class/subclasses.rb#L47
https://github.com/rails/rails/blob/b738f1930f3c82f51741ef7241c1fee691d7deb2/activesupport/lib/active_support/core_ext/class/subclasses.rb#L23

test/cli_test.rb
```diff
@@ -165,9 +165,12 @@ def test_format_duration
       assert_equal "0.0s", cli.format_duration(0.0)
       assert_equal "0.0s", cli.format_duration(0.000123)
       assert_equal "0.123s", cli.format_duration(0.1234567)
-      assert_equal "10.123s", cli.format_duration(10.1234567)
-      assert_equal "16m 40.123s", cli.format_duration(1000.1234567)
-      assert_equal "3h 46m 40.123s", cli.format_duration(100000.1234567)
```
ybiquitous, line 170: Incorrect expectations... 😓

ybiquitous:
@yykamei What do you think about this?

ybiquitous:
Thank you for your review! 😄

PR #1063, opened 2020-05-11
[SwiftLint] Improve error handling

ybiquitous:
Handling `JSON::ParserError` as a failure makes it hard for us to investigate a failure.
So, this change stops handling `JSON::ParserError`.
After this change, `JSON::ParserError` should be reported as a `Results::Error`.


ybiquitous:
CI passed! 🎉

PR #1069, opened 2020-05-12
[SwiftLint] Delete unchanged files

ybiquitous:
Since Runners 0.23.0, SwiftLint errors have increased.
This change tries to reduce them.


ybiquitous:
I will try this with Canary. 💪

PR #1070, opened 2020-05-12
[SwiftLint] Add smoke test for `unused_import` rule

ybiquitous:
See https://realm.github.io/SwiftLint/unused_import.html
Related to #1069


ybiquitous:
CI passed!

PR #1075, opened 2020-05-13
Add comment about #find_patch_by_file

yykamei:
Our implementation depends on [git_diff_parser](https://github.com/packsaddle/ruby-git_diff_parser), so I left comments about the behavior.

### DIFF
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -35,6 +35,7 @@ module Runners
       if gdp
         location = issue.location
         # NOTE: #find_patch_by_file omits just renamed files.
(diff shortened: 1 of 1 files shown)


lib/runners/changes.rb
```diff
@@ -34,6 +34,7 @@ def include?(issue)
       gdp = patches # NOTE: This assignment is required for typecheck by Steep
       if gdp
         location = issue.location
+        # NOTE: #find_patch_by_file omits just renamed files.
```
ybiquitous, line 37: Is it reasonable to add the related issue's link of GitDiffParser?

yykamei: Yes, I think so. Thank you :+1:

PR #1078, opened 2020-05-14
Add :use_local not to attempt to fetch gems remotely

### LINKED ISSUE #1074: Skip install_gems if the repository does not include Gem related information
If the repository does not include `Gemfile`, `Gemfile.lock`, `*.gemspec`, and the `gems` section in `sider.yml`, Runners do not have to run `install_gems`. This change improves performance a bit. Also, it would be possible to run Ruby analyses in some environments that have no access to the internet as long as Gem information does not exist in the repository.

yykamei:
If the repository does not include `Gemfile`, `Gemfile.lock`,
`*.gemspec`, and the `gems` section in `sider.yml`, Runners do not have
to access the internet. So, I made Runners not to fetch gems remotely
with the `--local` option of `bundle install`.

Close #1074

### DIFF
--- a/lib/runners/ruby.rb
+++ b/lib/runners/ruby.rb
@@ -3,6 +3,7 @@ module Runners
     class InstallGemsFailure < UserError; end
 
     def install_gems(default_specs, optionals: [], constraints:, &block)
+      original_default_specs = default_specs.dup
       user_specs = GemInstaller::Spec.from_gems(config_linter[:gems] || [])
 
       LockfileLoader.new(root_dir: root_dir, shell: shell).ensure_lockfile do |lockfile|
@@ -10,7 +11,7 @@ module Runners
         optionals = optional_specs(optionals, lockfile)
         user_specs = user_specs(user_specs, lockfile)
       end
-
+      use_local = (user_specs + optionals).empty? && original_default_specs == default_specs
       specs = GemInstaller::Spec.merge(default_specs, user_specs.size > 0 ? user_specs : optionals)
 
       mktmpdir do |path|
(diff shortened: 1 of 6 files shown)


test/ruby_test.rb
```diff
@@ -617,6 +623,9 @@ def test_install_gems_with_gemfile_lock
 
   def test_install_gems_with_gemfile_lock_which_does_not_satisfy_constraints
     with_workspace do |workspace|
+      # Install the default_specs beforehand.
+      shell.capture3!("gem", "install", "multi_json", "--version", "1.14.0")
```
yykamei, line 627: [Help Wanted] I couldn't find any solutions to install gems without affecting the test environment. Ideally, the test should not mutate the test environment, but `multi_json@1.14.0` will be always installed when running `bundle exec rake test`.

lib/runners/ruby.rb
```diff
@@ -3,14 +3,15 @@ module Ruby
...
+      original_default_specs = default_specs.dup
       user_specs = GemInstaller::Spec.from_gems(config_linter[:gems] || [])
 
       LockfileLoader.new(root_dir: root_dir, shell: shell).ensure_lockfile do |lockfile|
         default_specs = default_specs(default_specs, constraints, lockfile)
         optionals = optional_specs(optionals, lockfile)
         user_specs = user_specs(user_specs, lockfile)
       end
-
+      use_local = (user_specs + optionals).empty? && original_default_specs == default_specs
```
yykamei, line 14: In addition to check the absence of `user_specs + optionals`, we have to check whether the `default_specs` is overridden or not. If the `default_specs` is overridden, it means Runners must install the users-specified gems.

yykamei:
@ybiquitous What do you think about this direction?

ybiquitous:
I think it's correct. 👍

yykamei:
I will add a smoke test for goodcheck in the offline environment.

yykamei:
I changed some smoke tests of Goodcheck to use `add_offline_test`, and the changed tests have succeeded on CI.

PR #1080, opened 2020-05-14
Update Smoke.add_test to receive :options

yykamei:
I want to change the test behavior via `Smoke.add_test` and want to make
smoke tests without the internet.

This will be used on #1078
https://github.com/sider/runners/pull/1078#pullrequestreview-411525454

### DIFF
--- a/lib/runners/testing/smoke.rb
+++ b/lib/runners/testing/smoke.rb
@@ -36,11 +36,11 @@ module Runners
       def run
         load expectations.to_s
 
-        results = Parallel.map(self.class.tests, in_processes: ENV["JOBS"]&.to_i) do |test_set|
+        results = Parallel.map(self.class.tests, in_processes: ENV["JOBS"]&.to_i) do |params|
           out = StringIO.new(''.dup)
-          result = run_test(test_set, out)
+          result = run_test(params, out)
           print out.string
-          [result, test_set.name]
+          [result, params.name]
         end
 
         abort "❌ No smoke tests!" if results.empty?
@@ -61,8 +61,8 @@ module Runners
         end
       end
 
-      def run_test(test_set, out)
-        commandline = command_line(test_set)
+      def run_test(params, out)
+        commandline = command_line(params)
         out.puts "$ #{commandline}"
         reader = JSONSEQ::Reader.new(io: StringIO.new(`#{commandline}`), decoder: -> (json) { JSON.parse(json, symbolize_names: true) })
         traces = reader.each_object.to_a
@@ -75,7 +75,7 @@ module Runners
           Schema::Result.envelope =~ object
         }
 
-        unify_result(result, test_set.pattern, out) ? :passed : :failed
+        unify_result(result, params.pattern, out) ? :passed : :failed
       end
 
       def unify_result(result, pattern, out)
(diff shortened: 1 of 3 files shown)


test/smokes/misspell/expectations.rb
```diff
@@ -210,15 +214,17 @@
       git_blame_info: nil
     }
   ],
-  analyzer: { name: "Misspell", version: "0.3.4" }
+  analyzer: { name: "Misspell", version: "0.3.4" },
+  options: { offline: true }
```
yykamei, line 218: :thinking: The mixture of the pattern and the options might be confusing. What do you think? @ybiquitous

ybiquitous: Yes, it's confusing because `add_test()` arguments are expected results.

What about adding a new method `add_test_offline`?

yykamei: Thank you. It makes sense.

lib/runners/testing/smoke.rb
```diff
@@ -99,10 +101,13 @@ def unify_result(result, pattern, out)
...
       end
 
-      def command_line(name)
-        smoke_target = (expectations.parent / name).realpath
+      def command_line(test_set)
+        smoke_target = (expectations.parent / test_set.name).realpath
         runners_options = JSON.dump(source: { head: PROJECT_PATH })
-        commands = %W[docker run --rm --mount type=bind,source=#{smoke_target},target=#{PROJECT_PATH} --env RUNNERS_OPTIONS='#{runners_options}' #{docker_image} test-guid]
+        commands = %W[docker run --rm --mount type=bind,source=#{smoke_target},target=#{PROJECT_PATH} --env RUNNERS_OPTIONS='#{runners_options}']
+        commands << "--network=none" if test_set.options[:offline]
```
ybiquitous, line 108: 👍

lib/runners/testing/smoke.rb
```diff
@@ -13,6 +13,8 @@ class Smoke
 
       PROJECT_PATH = "/project".freeze
 
+      TestSet = Struct.new(:name, :pattern, :options, keyword_init: true)
```
ybiquitous, line 16: [suggestion] What about doing it simply like this?

```suggestion
      TestSet = Struct.new(:name, :pattern, :offline, keyword_init: true)
```

yykamei:
> [IMO] I think the name `TestSet` is a bit confusing, so what about `TestParams` instead?

LGTM. I will rename soon.

lib/runners/testing/smoke.rb
```diff
@@ -34,11 +36,11 @@ def initialize(argv)
       def run
         load expectations.to_s
 
-        results = Parallel.map(self.class.tests, in_processes: ENV["JOBS"]&.to_i) do |name, pattern|
+        results = Parallel.map(self.class.tests, in_processes: ENV["JOBS"]&.to_i) do |test_set|
```
ybiquitous, line 39: What about renaming also `test_set` to `params`?

yykamei:
Thank you!

yykamei:
I ignore "Codacy/PR Quality Review."

PR #1084, opened 2020-05-15
[Refactor] `Workspace` provide `Shell` helper

ybiquitous:
This change aims to reduce code duplication in `Workspace` and its subclasses
and to provide the common way for executing an external command via `Shell`.

- Replace the direct use of `Open3` with `Shell`.
- One call `Shell.new` in constructor.

### DIFF
--- a/lib/runners/shell.rb
+++ b/lib/runners/shell.rb
@@ -57,7 +57,6 @@ module Runners
     ensure
       dir_stack.pop
     end
-    alias chdir push_dir
 
     def push_env_hash(env)
       env_hash_stack.push(env)
--- a/lib/runners/shell.rb
+++ b/lib/runners/shell.rb
@@ -57,6 +57,7 @@ module Runners
     ensure
       dir_stack.pop
     end
+    alias chdir push_dir
 
     def push_env_hash(env)
       env_hash_stack.push(env)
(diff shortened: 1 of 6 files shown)


lib/runners/shell.rb
```diff
@@ -57,6 +57,7 @@ def push_dir(dir)
     ensure
       dir_stack.pop
     end
+    alias chdir push_dir
```
ybiquitous, line 60: [note] I think the name `push_dir` is confusing because it seems that the name `push` has a side effect. I will address it via another PR.

ybiquitous: Open #1085

yykamei: [IMO] Although I don't disagree with adding `alias`, the name of `push_dir` is appropriate for the current implementation since it has a "stack." So, it might be good to delete `dir_stack` if you try to implement it.

ybiquitous: OK, let's discuss in the PR #1085. I'll revert the alias in this PR. 👍

ybiquitous: Reverted via 29a9335.

ybiquitous:
Thank you for your review! 😸

PR #1086, opened 2020-05-15
Add top-level `Bugsnag.notify`

ybiquitous:
Bugsnag automatically registers the `at_exit` hook:

- https://github.com/bugsnag/bugsnag-ruby/blob/0466dd2a654cc1adbd1e1bb2fb7ba128fae3ba85/lib/bugsnag.rb#L129
- https://docs.bugsnag.com/platforms/ruby/other/#reporting-unhandled-exceptions

```ruby
at_exit do
  $! #=> SystemExit is ignored by default.
end
```

But an exception captured in `at_exit` is a `SystemExit` class,
and Bugsnag [ignores it by default](https://docs.bugsnag.com/platforms/ruby/other/configuration-options/#ignore_classes).

So, this change aims to report an expected error to add the top-level `rescue` block.

### Note

Since `Runners::Harness` has already called `Bugsnag.notify`, this addition will not have a big impact:

https://github.com/sider/runners/blob/c6c97a293f72a7c79deb07c78d66f36efa5940de/lib/runners/harness.rb#L92

### DIFF
--- a/bin/runners
+++ b/bin/runners
@@ -11,7 +11,7 @@ class RunnersTimeoutError < StandardError; end
 
 EXIT_STATUS_FOR_SIGTERM = 126
 
-# @see https://docs.bugsnag.com/platforms/ruby/rails/configuration-options
+# @see https://docs.bugsnag.com/platforms/ruby/configuration-options
 Bugsnag.configure do |config|
   config.app_version = ENV["RUNNERS_VERSION"]
   config.release_stage = ENV["BUGSNAG_RELEASE_STAGE"]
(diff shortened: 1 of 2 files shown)


bin/runners
```diff
@@ -32,4 +32,9 @@ trap('SIGUSR2') do
   Bugsnag.notify(RunnersTimeoutError.new)
 end
 
-Runners::CLI.new(argv: ARGV.dup, stdout: STDOUT, stderr: STDERR).run
+begin
+  Runners::CLI.new(argv: ARGV.dup, stdout: STDOUT, stderr: STDERR).run
+rescue => exn
```
yykamei, line 37: My understanding is that this change tries to handle `SystemExit`, isn't it? If so, I think `rescue` block should explicitly `SystemExit` because `rescue` block handles `StandardError` by default. Also, should we change `ignore_classes` to exclude `SystemExit` from it?

```suggestion
rescue StandardError, SystemExit => exn
```

ybiquitous: > My understanding is that this change tries to handle `SystemExit`, isn't it?

No. This PR aims to report `StandardError`. Unless this added `rescue`, any `StandardError` errors are not reported. 😢

yykamei: OK, I'm not sure. Let's do it :muscle:

ybiquitous:
Thanks 😉

Issue #951, opened 2020-04-07
"No space left on device" error on Git workspace


ybiquitous:
>  Errno::ENOSPC · No space left on device - copy_file_range 

https://github.com/sider/runners/blob/eb710073dbb23a51c29200e76fc96274a8cf3741/lib/runners/workspace/git.rb#L24-L29

ybiquitous:
I will try to use the git checkout directory as the working directory.

yykamei:
It must be a good idea.

ybiquitous:
Should `Workspace::Git` have two working directory (base and head)...? 🤔

ybiquitous:
It seems that we can simplify `Changes.calculate()` for Git workspace:

https://github.com/sider/runners/blob/fdaf9b95af8d3b2eee0a5e60e2e5064a14b9e40c/lib/runners/changes.rb#L49

For example, `git diff --stat 953c5e7a...HEAD` can get changed files. 🤔

yykamei:
For `Workspace::HTTP`, the self-implemented `calculate` seems necessary. Of course, only with `Workspace::Git`, git-diff(1) completely satisfies our needs.

PR #1104, opened 2020-05-18
[Refactor] Remove `working_dir` argument from `Changes.calculate`

ybiquitous:
This change aims to simplify the `Changes.calculate` method signature.
The removed `working_dir` argument is actually needless.
Because `FileUtils.copy_entry(head_path, working_dir)` is always called before calling `Changes.calculate`.

https://github.com/sider/runners/blob/288b1736a873325994c2da494e7c269e97dfe6e3/lib/runners/workspace.rb#L50-L56

Note: This PR is preparation for #951.

### DIFF
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -47,36 +47,26 @@ module Runners
       end
     end
 
-    def self.calculate(base_dir:, head_dir:, working_dir:, patches:)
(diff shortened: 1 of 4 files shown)


ybiquitous:
@yykamei Don't I miss something?

ybiquitous:
Thank you so much. 🥰 👍

PR #1105, opened 2020-05-18
[Refactor] Use head source directory as `Workspace#working_dir`

ybiquitous:
I don't think we need to create another working directory for a head source code.
This change aims to simplify the code in the `Workspace#open` method.

This also removes extra copying from `head_path` to `working_dir`,
so it probably can improve the performance a bit.

Related to #1104

Note: This PR is preparation for #951.

### DIFF
--- a/lib/runners/workspace.rb
+++ b/lib/runners/workspace.rb
@@ -37,26 +37,23 @@ module Runners
     # @yieldparam changes [Runners::Changes]
     def open
       prepare_ssh do |git_ssh_path|
+        trace_writer.header "Setting up source code"
+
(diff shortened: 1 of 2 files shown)


ybiquitous:
@yykamei Could you please review if you have time?

ybiquitous:
Good question. 👍 
I think there is no problem because this change does not change the behavior of `Workspace::Git`:

https://github.com/sider/runners/blob/286cf3b9551e680e864a37a190b3975dbd17f16c/lib/runners/workspace/git.rb#L27

Right?

ybiquitous:
Thank you for your review. 😸

PR #1111, opened 2020-05-19
Optimize `Workspace::Git#patches`

ybiquitous:
This change aims to prevent that `git diff` is accidentally called twice.


ybiquitous:
`#patches` is called once here:

https://github.com/sider/runners/blob/1872fe0658bb5dcf2211be25103c1879d552e0bf/lib/runners/workspace.rb#L53

PR #1117, opened 2020-05-21
[GolangCI-Lint] Explicitly set `--timeout`

ybiquitous:
If the `--timeout` option is not set on the command-line, users can change the value they want via `.golangci.yml`.
This change aims to prevent the change by users.

Also, we need to consider the top-level timeout value:

https://github.com/sider/runners/blob/1231de0ca047c7a23449ab1a7bb0751f39f16643/images/Dockerfile.end.erb#L11

Note: The GolangCI-Lint's default value is `1m`. It's too little.
https://golangci-lint.run/usage/configuration/#command-line-options

### DIFF
--- a/lib/runners/processor/golangci_lint.rb
+++ b/lib/runners/processor/golangci_lint.rb
@@ -75,12 +75,20 @@ module Runners
     def analyzer_options
       [].tap do |opts|
         analysis_targets.each { |target| opts << target }
+
+        # NOTE: Always override the value in `.golangci.yml`
         opts << "--out-format=json"
         opts << "--print-issued-lines"
         opts << "--print-linter-name"
         opts << "--issues-exit-code=0"
         opts << "--color=never"
         opts << "--concurrency=2"
+
+        # NOTE: The value should be less than the top-level timeout value.
+        #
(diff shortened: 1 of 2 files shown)


lib/runners/processor/golangci_lint.rb
```diff
@@ -75,12 +75,20 @@ def run_analyzer
...
         opts << "--print-issued-lines"
         opts << "--print-linter-name"
         opts << "--issues-exit-code=0"
         opts << "--color=never"
         opts << "--concurrency=2"
+
+        # NOTE: The value should be less than the top-level timeout value.
+        #
+        # @see https://github.com/sider/runners/blob/1231de0ca047c7a23449ab1a7bb0751f39f16643/images/Dockerfile.end.erb#L11
+        opts << "--timeout=15m"
```
ybiquitous, line 90: @yykamei What do you think about the value `15m`?

yykamei: It makes sense. According to our data within the last 24 hours, the max analysis time of GolangCI-Lint is 47 seconds. 15 minutes is sufficient and reasonable.

ybiquitous: Thank you! 👍

PR #1122, opened 2020-05-22
Support PMD CPD

### LINKED ISSUE #812: Support PMD CPD
## Information

- Website: https://pmd.github.io/latest/pmd_userdocs_cpd.html
- Source code: https://github.com/pmd/pmd
- Language: Many. See the [supported languages](https://pmd.github.io/latest/pmd_userdocs_cpd.html#supported-languages)

## Why do you need the tool?

Any CPD (Copy Paste Detector) tools have not been supported.

## Additional context

None.

## See also

None.

suetsuguk:
Fix #812

- [x] Add smoke tests for `--encoding`, `--non-recursive` and `--skip-lexical-errors` with C# files
- [x] Consider using `--failOnViolation`

It says:
By default CPD exits with status 4 if code duplications are found. Disable this option with --failOnViolation false to exit with 0 instead and just write the report.

In the current implementation, the runner will report `Results::Failure()` even if no code duplications are found. ->It was my misunderstanding. The smoke test 'no_issues' works fine.

Doc: https://github.com/sider/sider-docs/pull/431


images/pmd_cpd/Dockerfile.erb
```diff
@@ -0,0 +1,13 @@
+<%= render_erb 'images/Dockerfile.java.erb' %>
+<%= render_erb 'images/Dockerfile.base.erb' %>
+
+COPY --chown=<%= chown %> images/pmd_cpd/cpd ${RUNNER_USER_HOME}/bin/
+
+# NOTE: PMD does not have a CLI option to show its version.
+COPY --chown=<%= chown %> images/pmd_cpd/ShowPMDVersion.java ${RUNNER_USER_HOME}/bin/
+COPY --chown=<%= chown %> images/pmd_cpd/show_pmd_version ${RUNNER_USER_HOME}/bin/
```
ybiquitous, line 8: [suggestion] How about borrowing the files in `images/pmd_java` as follows?

```suggestion
COPY --chown=<%= chown %> images/pmd_java/ShowPMDVersion.java ${RUNNER_USER_HOME}/bin/
COPY --chown=<%= chown %> images/pmd_java/show_pmd_version ${RUNNER_USER_HOME}/bin/
```

If doing so, `images/pmd_cpd/{show_pmd_version,ShowPMDVersion.java}` can be removed.

images/pmd_cpd/pom.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>sider</groupId>
+  <artifactId>runners</artifactId>
+  <version>1.0.0</version>
+  <dependencies>
+    <dependency>
+      <groupId>net.sourceforge.pmd</groupId>
+      <artifactId>pmd-java</artifactId>
```
ybiquitous, line 10: [question] If we want to support all the languages, don't we need to change it as follows?

```suggestion
      <artifactId>pmd-dist</artifactId>
```

https://mvnrepository.com/artifact/net.sourceforge.pmd/pmd-dist

lib/runners/processor/pmd_cpd.rb
```diff
@@ -0,0 +1,199 @@
+module Runners
+  class Processor::PmdCpd < Processor
+    include Java
+
+    Schema = StrongJSON.new do
+      let :runner_config, Schema::BaseConfig.base.update_fields { |fields|
+        fields.merge!({
+                    minimum_tokens: numeric?,
+                    files: string?,
+                    filelist: string?,
```
ybiquitous, line 10: [suggestion] How about specifying target files or directories more easily?

```suggestion
                    files: enum(string, array(string)),
```

E.g. Support only `files` (required):

```yaml
linter:
  pmd_cpd:
    files: src #=> --files src
    # or
    files: [src, lib] #=> --files src --files lib
    # or
    files: [path/to/Some.java, src] #=> --files path/to/Some.java --files src
```

Ruby code:

```ruby
Array(config_linter[:files] || DEFAULT_FILES).map { |v| ["--files", v] }
```

See https://pmd.github.io/latest/pmd_userdocs_cpd.html#cli-usage

lib/runners/processor/pmd_cpd.rb
```diff
@@ -0,0 +1,199 @@
+module Runners
+  class Processor::PmdCpd < Processor
+    include Java
+
+    Schema = StrongJSON.new do
+      let :runner_config, Schema::BaseConfig.base.update_fields { |fields|
+        fields.merge!({
+                    minimum_tokens: numeric?,
```
ybiquitous, line 8: [suggestion] How about using `-` instead of `_`? Original CLI flags use `-`, so I think users are less confused.

```suggestion
                    minimum-tokens: numeric?,
```

Ruby code:

```diff
-config_linter[:minimum_tokens]
+config_linter[:"minimum-tokens"]
```

Other flags are the same.

test/smokes/pmd_cpd/expectations.rb
```diff
@@ -0,0 +1,378 @@
...
+}]
+      },
+      git_blame_info: nil
+    }
+  ],
+  analyzer: { name: "PMD CPD", version: "6.23.0" }
+)
+
+s.add_test(
+  "broken_sideci_yml",
```
ybiquitous, line 138: [suggestion] `sideci.yml` is old, so please use `sider.yml` instead. 🙏 


```suggestion
  "broken_sider_yml",
```

lib/runners/processor/pmd_cpd.rb
```diff
@@ -0,0 +1,199 @@
...
+    DEFAULT_FILES = ".".freeze
+
+    register_config_schema(name: :pmd_cpd, schema: Schema.runner_config)
+
+    def analyzer_version
+      @analyzer_version ||= capture3!("show_pmd_version").yield_self { |stdout,| stdout.strip }
+    end
+
+    def analyzer_bin
+      "cpd"
```
yykamei, line 50: [nitpick] You can remove this method overriding if you rename the file `images/pmd_cpd/cpd` to `images/pmd_cpd/pmd_cpd`. Of course, you can keep the code as it is.

images/pmd_cpd/Dockerfile.erb
suetsuguk: Good call!
Fixed. commit ffdbdeeec9214f7f9518a60e0bb4b3f5c2befaa3

lib/runners/processor/pmd_cpd.rb
suetsuguk: Good call!
Fixed. commit 8cc0eb507a0a1c0bac6e954c2f995c8cb7ed8349

lib/runners/processor/pmd_cpd.rb
suetsuguk: Good call!
Fixed. commit 2b3488b8aa47595daf381b39e1bf9aa2d9a13dbb

test/smokes/pmd_cpd/expectations.rb
suetsuguk: peace of cake.
Fixed. commit 50308ffeecbc22479ce5bdb1e8f3ac065b4c0530

lib/runners/processor/pmd_cpd.rb
suetsuguk: Good call!
Fixed. commit eb37f18a034beb2d1682220861dfd1eb9aad6927

images/pmd_cpd/pom.xml
suetsuguk: Good call!
The PMD CPD in pmd-java failed to analyze C/C++ source files. The pmd-dist package contains a support module for each language and enables us to analyze properly.
Fixed. commit 1475a277dacc5f6af25882246fa024115c1efa50

analyzers.yml
```diff
@@ -104,6 +104,10 @@ analyzers:
     name: PHPMD
     github: phpmd/phpmd
     doc: tools/php/phpmd
+  pmd_cpd:
+    name: PMD CPD
+    github: pmd/pmd
+    doc: tools/others/pmd_cpd
```
ybiquitous, line 110: [nitpick]

```suggestion
    doc: tools/others/pmd-cpd
```

README.md
```diff
@@ -36,6 +36,7 @@ All 36 analyzers are provided as a Docker image:
 | Phinder | [docker](https://hub.docker.com/r/sider/runner_phinder), [source](https://github.com/sider/phinder), [doc](https://help.sider.review/tools/php/phinder) | ✅ |
 | PHP_CodeSniffer | [docker](https://hub.docker.com/r/sider/runner_code_sniffer), [source](https://github.com/squizlabs/PHP_CodeSniffer), [doc](https://help.sider.review/tools/php/codesniffer) | ✅ |
 | PHPMD | [docker](https://hub.docker.com/r/sider/runner_phpmd), [source](https://github.com/phpmd/phpmd), [doc](https://help.sider.review/tools/php/phpmd) | ✅ |
+| PMD CPD | [docker](https://hub.docker.com/r/sider/runner_pmd_cpd), [source](https://github.com/pmd/pmd), [doc](https://help.sider.review/tools/others/pmd_cpd) | ✅ |
```
suetsuguk, line 39:
```suggestion
| PMD CPD | [docker](https://hub.docker.com/r/sider/runner_pmd_cpd), [source](https://github.com/pmd/pmd), [doc](https://help.sider.review/tools/others/pmd-cpd) | ✅ |
```

Refer to https://github.com/sider/sider-docs/pull/431#discussion_r430216578

ybiquitous: Yes, right. When you change `analyzers.yml` and run `rake readme:generate`, the README part will be updated correctly. 👌

lib/runners/processor/pmd_cpd.rb
```diff
@@ -0,0 +1,191 @@
...
+      return {
+        lines: lines,
+        tokens: tokens,
+        files: fileobjs,
+        codefragment: codefragment
+      }
+    end
+
+    def option_files
+      Array(config_linter[:files] || DEFAULT_FILES).map { |v| ["--files", v] }.flatten
```
ybiquitous, line 121: [IMO] PMD CPD's `--files` name is so confusing... 😢 (not need to fix)

lib/runners/processor/pmd_cpd.rb
```diff
@@ -0,0 +1,191 @@
...
+        *option_skip_duplicate_files,
+        *option_non_recursive,
+        *option_skip_lexical_errors,
+        *option_ignore_annotations,
+        *option_ignore_identifiers,
+        *option_ignore_literals,
+        *option_ignore_usings,
+        *option_no_skip_blocks,
+        *option_skip_blocks_pattern,
+        "--format", "xml",
```
ybiquitous, line 186: [suggestion] What about setting `--failOnViolation false`?

```suggestion
        "--format", "xml",
        "--failOnViolation", "false",
```

If doing so, `#run_analyzer` method will be simpler (because `capture3!` can be used):

```ruby
def run_analyzer
  stdout, = capture3!(analyzer_bin, *cli_options)
    construct_result(result, stdout)
  end
end
```

`capture3!` method can raise `Shell::ExecError` if the process exits with the non-zero code.

Also, I think the following code is not good. Because `Results::Failure` is returned when `exitstatus` is not `0` or `4`.
`Results::Failure` is due to user's misconfiguration etc. and is not reported to Bugsnag, so we will be hard to notice what happens if unexpected errors occur.

```
if status.success? || status.exitstatus == 4
    construct_result(result, stdout, stderr)
  end
else
  Results::Failure.new(guid: guid, analyzer: analyzer, message: "Unexpected error occurred. Please see the analysis log.")
end
```

lib/runners/processor/pmd_cpd.rb
```diff
@@ -0,0 +1,191 @@
...
+        Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+          construct_result(result, stdout, stderr)
+        end
+      else
+        Results::Failure.new(guid: guid, analyzer: analyzer, message: "Unexpected error occurred. Please see the analysis log.")
+      end
+    end
+
+    def construct_result(result, stdout, stderr)
+      stdout.sub!(/encoding=".*?"/,'') # Remove the encoding attribute to read this XML as UTF-8. The PMD CPD writes an XML report as UTF-8 with the inconsistent encoding attribute value when the --encoding option is specified.
```
ybiquitous, line 64: [suggestion] What about the following refactoring?

- Make the trailing long line comment readable to show it without page scrolling.
- Clarity replacing the XML declaration.

```suggestion
      # HACK: Remove the encoding attribute to read this XML as UTF-8.
      #       The PMD CPD writes an XML report as UTF-8 with the inconsistent encoding attribute value when the --encoding option is specified.
      stdout.sub!(/<?xml version="1.0" encoding=".+"?>/, '<?xml version="1.0"?>')

```

lib/runners/processor/pmd_cpd.rb
yykamei: Agreed. :thinking: Would it be better to use `target` instead of `files`?

ybiquitous: Umm..., it's difficult. Because CPD's flag name is `--files`. If we would use `target`, CPD users might be confused.

yykamei: I see. Let's consider later. Anyway, it's OK to keep the code :+1:

ybiquitous:
> it might be difficult to read the Hashed issue ID like `069e8856ac66563b0c6e50b615209face9958fa7`.

It seems a bad UX, but I don't have any idea... 🤷

analyzers.yml
suetsuguk: LGTM
Fixed. commit 2fb310be18c5c6543dc839f8bf3b9614a37d87d0

README.md
suetsuguk: Fixed. commit 2fb310be18c5c6543dc839f8bf3b9614a37d87d0
Now 'rake readme:generate' succeeds.

suetsuguk:
> Oh, I forgot it. 😅 Can you change `.dependabot/config.yml` to add PMD CPD?
> 
> https://github.com/sider/runners/blob/b1bc3d889987e912ed4430c0ae4416e15314b37c/.dependabot/config.yml#L25-L27

LGTM
Fixed. commit 255027a6e36cd1b77116fb8380971b1844c497b1

lib/runners/processor/pmd_cpd.rb
suetsuguk: I am concerned about changes of the XML declaration in the future, e.g. adding new attributes. But we already have smoke tests to detect the changes, so I have accepted your suggestion.
commit ddaa70724b929a934c6cdff64e86eab062e8f8b5

Also I have replaced `encoding="foo"` with `encoding="UTF-8"` rather than just deleting, so that the parser uses 'UTF-8' instead of the default encoding.
https://github.com/sider/runners/commit/ddaa70724b929a934c6cdff64e86eab062e8f8b5#diff-78e7c4c97abb54a5880bb4582b2322c8R66

ybiquitous: Thanks!

lib/runners/processor/pmd_cpd.rb
suetsuguk: That makes sense. Thank you for the detailed explanation.
Fixed. commit b090cb1fe9f00c829be4f1d976b9665ae0d54b7e

suetsuguk:
I happen to find that PMD CPD writes a list of files being skipped to stderr in the case below:
- found files of the same name and size when --skip-duplicate-files is enabled
- found files which can't be tokenized when --skip-lexical-errors is enabled

I have added the messages to the warning list with the `add_warning` method not to squash them.
https://github.com/sider/runners/pull/1122/files#diff-78e7c4c97abb54a5880bb4582b2322c8R62-R68

```
    def raise_warnings(stderr)
      stderr.each_line do |line|
        if line.start_with?("Skipping")
          add_warning line.gsub(working_dir.to_s + '/', '')
        end
      end
    end

  warnings: [
    {
      message: "Skipping src/bar/baz/こんにちは.1.eucjp.cs. Reason: Lexical error in file src/bar/baz/こんにちは.1.eucjp.cs at line 20, column 13.  Encountered: token recognition error at: '｣'",
      file: nil
    },
    {
      message: "Skipping src/bar/こんにちは.2.sjis.cs since it appears to be a duplicate file and --skip-duplicate-files is set",
      file: nil
    },
    {
      message: "Skipping src/baz/こんにちは.2.sjis.cs since it appears to be a duplicate file and --skip-duplicate-files is set",
      file: nil
    }
  ],
```

lib/runners/processor/pmd_cpd.rb
ybiquitous: Nice! 👏

ybiquitous:
Good catch! 👍

ybiquitous:
Could you please re-run `bundle exec rake readme:generate` and push the change to make the failed commit status green? 🙏

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 
 [Full diff](https://github.com/sider/runners/compare/0.25.4...HEAD)
 
+- [PMD CPD] New support [#1122](https://github.com/sider/runners/pull/1122)
+
```
ybiquitous, line 10: Please update the entry before merge. 🙏 
(0.26.0 was released)

ybiquitous:
@suetsuguk Please *squash and merge*. And ignore Codacy failure. 🙏

Issue #1142, opened 2020-05-26
git-blame sometimes fails


ybiquitous:
git-blame sometimes fails, for example:

> Runners::Shell::ExecError · Command [git blame -p -L 12,12 8cc0eb507a0a1c0bac6e954c2f995c8cb7ed8349 -- images/pmd_cpd/Dockerfile] exited with 128

ybiquitous:
Maybe cause:

https://github.com/sider/runners/blob/b35adb15a8efde210114934385c3d0cd12856d11/lib/runners/ignoring.rb#L51

PR #1143, opened 2020-05-26
Add `git-blame` failed exception

ybiquitous:
To look into the issue #1142.


ybiquitous:
I will ignore Sider errors now.

Issue #1144, opened 2020-05-26
`ignore` option does not work


ybiquitous:
https://sider.review/gh/repos/202486306/pulls/1143/logs/d477d2f5-4776-44d8-b3fb-ced2b68cb216

<img width="1241" alt="image" src="https://user-images.githubusercontent.com/473530/82854051-460aeb00-9f42-11ea-8753-372bfc9233aa.png">

https://github.com/sider/runners/blob/77ab6a3165d8c3796e0d254de0170c234c752bf6/sider.yml#L10-L11

PR #1112, opened 2020-05-19
Fix "No space left" error on Git workspace

ybiquitous:
This change aims to prevent the "No space left on device" error by stopping a copy of the entire working directory.

https://github.com/sider/runners/blob/48e04a96ee7e477394deb6a836edda3ded5e5c3b/lib/runners/workspace/git.rb#L26

This removal makes the `Workspace::Git` code more simple.

Fix #951

PR #1147, opened 2020-05-26
Do not remove .git within #with_gitignore

### LINKED ISSUE #1142: git-blame sometimes fails

yykamei:
We started to reuse Git directory as a `working_dir` since #1112, so a `working_dir` always includes `.git` directory. However, `Ignoring` class still removes `.git` directory after git-check-ignore(1), then it
causes errors afterward.

Fix #1142

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 [Full diff](https://github.com/sider/runners/compare/0.25.1...HEAD)
 
 - Output trace more [#1145](https://github.com/sider/runners/pull/1145)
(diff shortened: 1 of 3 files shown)


ybiquitous:
Maybe this PR can fix the issue #1142.

yykamei:
We're using `Workspace::HTTP`, so I guess we can't remove yet.

ybiquitous:
Why will this PR fix the issue #1144?

yykamei:
#1144 is not related. Thank you.

PR #1149, opened 2020-05-27
Add Pylint

### LINKED ISSUE #1100: Support Pylint
## Information

- Website: https://www.pylint.org/
- Source code: https://github.com/PyCQA/pylint
- Language: Python

## Why do you need the tool?

- Just want to use Pylint

dev-yakuza:
Fix #1100

- [X] Write a [document](https://github.com/sider/sider-docs) for the new runner. PR sider/sider-docs#432


ybiquitous:
Can you fix the failed unit test? ⬇️ 

> ConfigTest#test_content_with_linter_section [/home/runner/work/runners/runners/test/config_test.rb:34]:

https://github.com/sider/runners/pull/1149/checks?check_run_id=711498577#step:5:240

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 [Full diff](https://github.com/sider/runners/compare/0.25.2...HEAD)
 
+New supported:
+
+- **Pylint** [#1149](https://github.com/sider/runners/pull/1149)
+
```
ybiquitous, line 12: CHANGELOG.md is updated for **0.25.3**, so please merge or rebase. 🙏

analyzers.yml
```diff
@@ -153,3 +153,7 @@ analyzers:
     name: TyScan
     github: sider/TyScan
     doc: tools/javascript/tyscan
+  pylint:
+    name: pylint
+    github: PyCQA/pylint
+    doc: tools/python/pylint
```
ybiquitous, line 159: Please fix the position. 🙏 

https://github.com/sider/runners/blob/686926eac1be5572e53471777a52360b2269c2ef/analyzers.yml#L2

.github/workflows/build.yml
```diff
@@ -95,6 +95,7 @@ jobs:
           - swiftlint
           - tslint
           - tyscan
+          - pylint
```
ybiquitous, line 98: Please fix the position according to the alphabetical order. 🙏

analyzers.yml
```diff
@@ -153,3 +153,7 @@ analyzers:
     name: TyScan
     github: sider/TyScan
     doc: tools/javascript/tyscan
+  pylint:
+    name: pylint
```
ybiquitous, line 157: It seems to me that the official name is **Pylint** instead of pylint.

```suggestion
    name: Pylint
```

lib/runners.rb
```diff
@@ -107,6 +107,7 @@
 require "runners/processor/swiftlint"
 require "runners/processor/tslint"
 require "runners/processor/tyscan"
+require "runners/processor/pylint"
```
ybiquitous, line 110: Alphabetical order also here.

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,91 @@
+module Runners
+  class Processor::Pylint < Processor
+    Schema = StrongJSON.new do
+      let :runner_config, Schema::BaseConfig.base.update_fields { |fields|
+        fields.merge!(
+          rcfile: string?,
+          ignore: enum?(string, array(string)),
+          errorsOnly: boolean?,
```
ybiquitous, line 8: Please fix it according to the convention between other tools:

```suggestion
          errors-only: boolean?,
```

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def erros_only
+      config = config_linter[:errorsOnly] || config_linter.dig(:options, :errorsOnly)
+      ["--errors-only"] if config
+    end
+
+    def analyzed_files
+      # Via glob
+      targets = Array(config_linter[:target] || DEFAULT_TARGET)
```
ybiquitous, line 46: Do we need to add `target` into `Schema.runner_config`?

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,91 @@
...
+      let :rule, object(
+        severity: string,
+      )
+    end
+
+    register_config_schema(name: :pylint, schema: Schema.runner_config)
+
+    DEFAULT_TARGET = [
+      "**/*.{py}",
+      { shebang: true },
```
ybiquitous, line 21: [question] Is this `{ shebang: true },` necessary?

lib/runners/processor/pylint.rb
dev-yakuza: Sorry, we don't need it. it's a copy-paste mistake. 😭 
I will modify it.

Thanks.

lib/runners/processor/pylint.rb
ybiquitous: Sorry, quotes are needed:

```suggestion
          "errors-only": boolean?,
```

Usage:

```ruby
config_linter[:"errors-only"]
```

lib/runners/processor/pylint.rb
dev-yakuza: I tested it, it looks not necessary. I will delete it!

Thanks 🙇‍♂️

CHANGELOG.md
dev-yakuza: I will rebase and merge after all reviews!

Thanks

lib/runners/processor/pylint.rb
ybiquitous: I think `target` option is useful for some users, for example:

```yaml
linter:
  pylint:
    target: 
      - "src/**/*.py"
      - "test/**/*.py"
```

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,86 @@
+module Runners
+  class Processor::Pylint < Processor
```
ybiquitous, line 2: [suggestion] How about including `Python` module? See also `processor/flake8.rb`

```suggestion
  class Processor::Pylint < Processor
    include Python

```

lib/runners/processor/pylint.rb
dev-yakuza: OK, I will add the option!

Thanks!

dev-yakuza:
> [question] Can we provide our default configuration for Pylint if users have their own configuration?

I think it's difficult. According to the document([command-line-options](https://docs.pylint.org/en/1.6.0/run.html#command-line-options)), 

> You can specify a configuration file on the command line using the `--rcfile` option. Otherwise, Pylint searches for a configuration file in the following order and uses the first one it finds:
>1. `pylintrc` in the current working directory
>2. `.pylintrc` in the current working directory

So, if the user has `pylintrc` or `.pylintrc` file in the working directory, Pylint refers to the file first.

we can set `--rcfile` option for our default configuration, but it will ignore the user configuration file if the user didn't set `--rcfile` option in `sider.yml`.

ybiquitous:
Thanks, I understand it. When users turn on Pylint on Sider and they don't have any configuration files, then what happens?

BTW, the following link seems the latest:
https://github.com/PyCQA/pylint/blob/master/doc/user_guide/run.rst

dev-yakuza:
> Thanks, I understand it. When users turn on Pylint on Sider and they don't have any configuration files, then what happens?

Pylint has the default configuration([pylintrc](https://github.com/PyCQA/pylint/blob/master/pylintrc)), so If the user doesn't have `pytlintrc` file, Pylint would use the default configuration.

So, I think we don't need to make our default configuration file.

ybiquitous:
Great. 👍 

BTW, it seems better to explicitly set `--jobs=2` option when running it on Sider. Because users can overwrite `jobs` via their `pylintrc` file.

https://github.com/PyCQA/pylint/blob/1fc490c0a1d0df5d4d6a64da0ccf2dd711b9bf67/pylintrc#L23

There may be other options like `jobs`. What do you think?

dev-yakuza:
>BTW, it seems better to explicitly set --jobs=2 option when running it on Sider. Because users can overwrite jobs via their pylintrc file.

OK, I will set `--jobs=2`.
I think other options are fine.

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,91 @@
...
+      trace_writer.message "Analyzing #{files.size} file(s)..."
+
+      stdout, stderr = capture3(
+        analyzer_bin,
+        *files,
+        *rcfile,
+        *ignore,
+        *erros_only,
+        '--output-format=json',
+      )
```
ybiquitous, line 80: [suggestion] How about setting `--exit-zero`?

> --exit-zero
> Always return a 0 (non-error) status code, even if
> lint errors are found. This is primarily useful in
> continuous integration scripts.

If so, we can use `capture3!` instead of `capture3` as follows:

```ruby
stdout, stderr = capture3!(...)
```

dev-yakuza: Oh! it looks better! I will add and test it.

Thanks! 🙏

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    def analyze(changes)
+      run_analyzer
+    end
+
+    private
+
+    def rcfile
+      file = config_linter[:rcfile]
+      Array(file ? "--rcfile=#{file}" : nil)
```
ybiquitous, line 32:
```suggestion
      config_linter[:rcfile].then { |file| file ? ["--rcfile=#{file}"] : [] }
```

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    def ignore
+      Array(config_linter[:ignore] || []).then { |arr| arr.empty? ? nil : "--ignore=#{arr.join(",")}" }
+    end
+
+    def erros_only
+      config = config_linter[:'errors-only'] || config_linter.dig(:options, :'errors-only')
+      ["--errors-only"] if config
```
ybiquitous, line 41:
```suggestion
      config_linter[:'errors-only'].then { |value| value ? ["--errors-only"] : [] }
```

`:options` does not exist 😏

test/smokes/pylint/expectations.rb
```diff
@@ -0,0 +1,152 @@
...
+s.add_test(
+  "success",
+  type: "success",
+  issues: [],
+  analyzer: { name: "Pylint", version: "2.5.2" }
+)
+
+s.add_test(
+  "failure",
+  timestamp: :_,
```
ybiquitous, line 12: [nitpick] This can be omitted:

```suggestion
  timestamp: :_,
```

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,91 @@
...
+      targets = Array(config_linter[:target] || DEFAULT_TARGET)
+      globs = targets.select { |glob| glob.is_a? String }
+      Dir.glob(globs, File::FNM_EXTGLOB, base: current_dir)
+    end
+
+    def parse_result(output)
+      json = JSON.parse(output, symbolize_names: true)
+      json.flat_map do |issue|
+        yield Issue.new(
+          id: "[#{issue[:'message-id']}]#{issue[:symbol]}",
```
ybiquitous, line 53: [question] What differences between the following?

- `id: "[#{issue[:'message-id']}]#{issue[:symbol]}",`
- `id: issue[:'message-id'],`

What advantages to set `:symbol`?

dev-yakuza: I thought it is a more user-friendly message.

- message-id only: `E0602`
- message-id with the symbol: `[E0602]undefined-variable`

It is just my opinion. What do you think about it?

ybiquitous: `message-id` seems an internal expression of Pylint to me. 🤔 

Is there a document to explain about `message-id` and `symbol`?

ybiquitous: I think that the `issue.id` field should be simple and clear for users, e.g. ESLint's [no-empty](https://eslint.org/docs/rules/no-empty).

- Users see the `id` on Sider.
- Users configure tool's config file and `sider.yml` via the `id`.

So, showing two IDs like `[E0602]undefined-variable` can make users confused.

dev-yakuza: There is the doc about the messages. ([basic-checker-messages](http://pylint.pycqa.org/en/latest/technical_reference/features.html#basic-checker-messages))

OK, I got it. I will change to display only `message-id`.

Thanks!

ybiquitous: It seems good to add some metadata to a `rule` object as follows:

```ruby
let :rule, object(
  severity: string,
  symbol: string,
  module: string,
)
```

We will be able to use them in the future.

ybiquitous: I found the doc on GitHub:
https://github.com/PyCQA/pylint/blob/master/doc/user_guide/message-control.rst

ybiquitous: > OK, I got it. I will change to display only `message-id`.

Sorry, isn't showing only `symbol` better? It's more readable.

ybiquitous:
We can close all Sider issues. 👍

ybiquitous:
Can you please merge master to this branch? Then, if CI passes, you can merge it.

yykamei:
I'm seeing this PR. Sorry to late :pray:

yykamei:
In my opinion, the default configuration of Pylint is too noisy for me. For example, it issues the violations of the "Missing docstring." Of course, it's better to write document for modules, functions, and classes, but this does not fit for every project. So, I think we should our default configuration like runner_flake8 in the near future. Fortunately, Pylint sees the configuration exists on the HOME directory (See the description about `--rcfile` of [Command line options](http://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options)).

However, our new `runner_pylint` will be beta, so it's good to inject the default configuration later.

ybiquitous:
@yykamei's idea makes sense to me. But `missing-docstring` is disabled by default?

https://github.com/PyCQA/pylint/blob/1ad2f2550ca34f12da0a1065cd1e61cdeefba90b/pylintrc#L56-L64

yykamei:
To be accurate, `missing-function-docstring`, `missing-class-docstring`, and `missing-module-docstring` are still enabled by default.

ybiquitous:
Thanks, I see. 👍

ybiquitous:
> However, our new `runner_pylint` will be beta, so it's good to inject the default configuration later.

Agreed 👍

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,97 @@
...
+      stdout, stderr = capture3!(
+        analyzer_bin,
+        *files,
+        *rcfile,
+        *ignore,
+        *erros_only,
+        '--output-format=json',
+        '--jobs=2',
+        '--exit-zero',
+      )
```
yykamei, line 85: Maybe, we should add `--disable=import-error` in the command line arguments because the violation `import-error` would be issued against almost all projects. This is because Pylint tries to [import modules](https://github.com/PyCQA/pylint/blob/1ad2f2550ca34f12da0a1065cd1e61cdeefba90b/pylint/checkers/imports.py#L784), but our Runners does not install dependencies from users' repository. Ideally, it's good to support the violation `import-error`, but it would not be worth our while to deal with just one violation.

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,98 @@
...
+      stdout, stderr = capture3!(
+        analyzer_bin,
+        *files,
+        *rcfile,
+        *ignore,
+        *erros_only,
+        '--output-format=json',
+        '--jobs=2',
+        '--exit-zero',
+        '--disable=E0401',
```
ybiquitous, line 85: [question] Always disable `E0401`? Why?

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+        parse_result(stdout) do |issue|
+          result.add_issue issue
+        end
+      end
+    end
+
+  end
+end
```
ybiquitous, line 98: [question] Do you setup EditorConfig?

https://github.com/sider/runners/blob/529e3411ce704e257ad232fe2fc9c0b86a3d50ff/.editorconfig#L7

https://github.com/sider/runners#prerequisites

lib/runners/processor/pylint.rb
dev-yakuza: >Maybe, we should add --disable=import-error in the command line arguments because the violation import-error would be issued against almost all projects. 

`E0401` is the message-id of `import-error` symbol.

ybiquitous: I see. Then, can you please add a comment and use `import-error` instead of `E0401`?
It's better to add a link to a document about Pylint's importing modules.

```suggestion
        # NOTE: We don't support importing modules yet. See https://...
        '--disable=import-error',
```

lib/runners/processor/pylint.rb
dev-yakuza: Sorry, I will setup it.

lib/runners/processor/pylint.rb
```diff
@@ -0,0 +1,102 @@
...
+      stdout, stderr = capture3!(
+        analyzer_bin,
+        *files,
+        *rcfile,
+        *ignore,
+        *erros_only,
+        '--output-format=json',
+        # NOTE: If jobs option is not 1, the output is different
+        # @see https://github.com/PyCQA/pylint/issues/374
+        '--jobs=1',
```
ybiquitous, line 85: Nice comment. 👍

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 [Full diff](https://github.com/sider/runners/compare/0.25.5...HEAD)
 
+New supported:
+
+- **Pylint** [#1149](https://github.com/sider/runners/pull/1149)
+
```
ybiquitous, line 12: 0.25.6 was released:

https://github.com/sider/runners/blob/720548da48caa6472b543e4a37f9842138701114/CHANGELOG.md

PR #1148, opened 2020-05-26
Use git-ls-files instead of git-check-ignore

ybiquitous:
[git-check-ignore](https://git-scm.com/docs/git-check-ignore) does not work but I cannot find the cause. 🤷 
So, I will try [git-ls-files](https://git-scm.com/docs/git-ls-files) instead to fix the bug.

Fix #1144

PR #1150, opened 2020-05-27
Generated ignoring file sometimes is missing

ybiquitous:
Related to #1148

Example:

```console
$ git ls-files --ignored --exclude-from /tmp/gitignore-20200527-6-i7ma6c
fatal: cannot use /tmp/gitignore-20200527-6-i7ma6c as an exclude file
```

I recommend [viewing it with hidden whitespaces](https://github.com/sider/runners/pull/1150/files?w=1).

### DIFF
--- a/lib/runners/ignoring.rb
+++ b/lib/runners/ignoring.rb
@@ -32,22 +32,22 @@ module Runners
     end
 
     def each_ignored_file(&block)
-      gitignore = Tempfile.open(["gitignore-"]) do |file|
-        file.write ignores.join("\n")
-        file.path
+      Tempfile.create("gitignore-") do |file|
+        gitignore = file.path
+        File.write gitignore, ignores.join("\n")
+
+        shell = Shell.new(current_dir: working_dir, trace_writer: trace_writer, env_hash: {})
+        shell.capture3! "git", "init"
+        shell.capture3! "git", "add", "."
+
+        # @see https://git-scm.com/docs/git-ls-files
+        stdout, = shell.capture3!(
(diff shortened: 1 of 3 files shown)


lib/runners/ignoring.rb
```diff
@@ -32,22 +32,22 @@ def ignores
     end
 
     def each_ignored_file(&block)
-      gitignore = Tempfile.open(["gitignore-"]) do |file|
-        file.write ignores.join("\n")
-        file.path
+      Tempfile.create("gitignore-") do |file|
```
ybiquitous, line 35: The reason to use `Tempfile.create` instead of `Tempfile.open`:

> If a block is given, then a File object will be constructed, and the block is invoked with the object as the argument. The File object will be automatically closed and the temporary file is removed after the block terminates.

https://ruby-doc.org/stdlib-2.7.1/libdoc/tempfile/rdoc/Tempfile.html#method-c-create

lib/runners/ignoring.rb
```diff
@@ -32,22 +32,22 @@ def ignores
...
+        gitignore = file.path
+        File.write gitignore, ignores.join("\n")
```
ybiquitous, line 37: [note] The reason to use [`File.write`](https://ruby-doc.org/core-2.7.1/IO.html#method-c-write) instead of [`#write`](https://ruby-doc.org/core-2.7.1/IO.html#method-i-write):

The former:

> ensures the file is closed before returning.

ybiquitous:
CI passed. I will try it.

PR #1158, opened 2020-05-29
Fix no file error in Pylint

dev-yakuza:
If there is no file in the folder, the error occurs in Pylint.

### DIFF
--- a/lib/runners/processor/pylint.rb
+++ b/lib/runners/processor/pylint.rb
@@ -73,6 +73,10 @@ module Runners
 
       trace_writer.message "Analyzing #{files.size} file(s)..."
 
+      if files.size == 0
+        return Results::Success.new(guid: guid, analyzer: analyzer)
+      end
+
       stdout, stderr = capture3!(
         analyzer_bin,
         *files,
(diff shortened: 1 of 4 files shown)


lib/runners/processor/pylint.rb
```diff
@@ -73,6 +73,10 @@ def run_analyzer
 
       trace_writer.message "Analyzing #{files.size} file(s)..."
 
+      if files.size == 0
+        return Results::Success.new(guid: guid, analyzer: analyzer)
+      end
```
ybiquitous, line 78: [suggestion] What about this?

- Fix the position of the message "Analyzing...".
- Add a message for no files found.
- Use `empty?` instead of `size == 0`.

```suggestion
      if files.empty?
        trace_writer.message "No files found."
        return Results::Success.new(guid: guid, analyzer: analyzer)
      end

      trace_writer.message "Analyzing #{files.size} file(s)..."
```

test/smokes/pylint/no_file/sider.yml
```diff
@@ -0,0 +1,3 @@
+linter:
+  pylint:
+    errors-only: true
```
ybiquitous, line 3: Can you replace this `sider.yml` with `.keep` file to clarify the test purpose?

PR #1166, opened 2020-06-05
[Cppcheck] Support "Bug hunting"

### LINKED ISSUE #1138: [Cppcheck] Support "Bug hunting"
> ### Bug hunting
>
> There is a new "soundy" analysis in Cppcheck that should detect most bugs. You should expect false alarms, however the false alarms will not be overwhelming.
>
> This new "soundy" analysis is not intended to replace normal Cppcheck analysis. > There are use cases where false alarms can not be tolerated.
>
> We have added 1 checker and that checks for division by zero:
>
>    It detects all "integer division by zero" bugs in the Juliet test suite.
>    It detects all "division by zero" bugs in the ITC test suite.
>    There was 28 division by zero CVEs published in 2019 for C/C++ open source projects, and we could quickly see that 21 of the bugs are found by Cppcheck. > There is no CVE bug that we know Cppcheck fails to diagnose. But there are 7 CVEs that would require additional investigation to establish if it is really detected or not.

https://github.com/danmar/cppcheck/releases/tag/2.0

New feature! 🎉 Let's support it!

See also:
https://github.com/danmar/cppcheck/blob/master/man/manual.md#bug-hunting

_Originally posted by @ybiquitous in https://github.com/sider/runners/pull/1083#issuecomment-629028050_

suetsuguk:
Fix #1138

- [x] Build cppcheck with z3
- [x] Run cppcheck twice and combine results
- [x] Write smoke tests accordingly

- [ ] Write a [document](https://github.com/sider/sider-docs) and link the PR sider/sider-docs#<PULL_NUMBER>


images/cppcheck/Dockerfile
```diff
@@ -9,15 +9,21 @@ FROM sider/devon_rex_python:2.19.0
 # See https://github.com/danmar/cppcheck#gnu-make
 
 ARG CPPCHECK_VERSION=2.0
+ARG LIBZ3_VERSION=4.4.1-1~deb10u1
```
ybiquitous, line 12: [question] Will we need to maintain and update `LIBZ3_VERSION` from now on? Can we make the version relaxed as follows?

```diff
-apt-get install -y --no-install-recommends libz3-dev=${LIBZ3_VERSION}
+apt-get install -y --no-install-recommends libz3-dev
```

test/smokes/cppcheck/expectations.rb
```diff
@@ -492,3 +492,243 @@
...
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "src/bar/c_sample.c",
+      location: { start_line: 18 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
```
dev-yakuza, line 522: [Question] is it possible to reduce the same issues by modifying the test code? I think one issue is enough to check the same issues.

lib/runners/processor/cppcheck.rb
```diff
@@ -65,18 +66,40 @@ def language
...
+        ret = step_analyzer(results, *args)
+        if ret
+          return ret
+        end
+      end
+
+      return results
+    end
+
+    def step_analyzer(results, *args)
```
dev-yakuza, line 93: [Question] I think there is a reason to make `step_analyzer`. Could you explain why you separate it? (I just want to learn personally. 😅 )

images/cppcheck/Dockerfile
suetsuguk: I found out that cppcheck compiled with the latest version of Z3 (v.4.8.8) failed due to segmentation fault. Current version of the debian package worked fine. But it will be updated sometime in the future, and we may get the failure then. In case potential trouble happen, I have specified the version number.

We also have made smoke tests to detect such a potential trouble. Shall we make the version relaxed for the time being?

lib/runners/processor/cppcheck.rb
suetsuguk: My aim is to reduce code clones. We need to to run `cppcheck` twice as pseudo code below;

```
def run_analyzer()
  results = Result()

  ret = cppcheck("normal")
  if ret is error1
    return ret
  if ret is error2
    return ret
  results.add(ret.issues)

  ret = cppcheck("bug-hunting")
  if ret is error1
    return ret
  if ret is error2
    return ret
  results.add(ret.issues)

  return  results
```

The `step_analyzer` make the code simple.

```
def run_analyer()
  results = Result()

  for args in ["normal", "bug-hunting"]
    ret = step_analyze(results, args)
    if ret
      return ret

  return results

def step_analyzer(results, args)
  ret = cppcheck(args)
  if ret is error1
    return ret
  if ret is error2
    return ret
  results.add(ret.issues)
```

images/cppcheck/Dockerfile
ybiquitous: > We also have made smoke tests to detect such a potential trouble. Shall we make the version relaxed for the time being?

Nice! 👏 
I probably that we will forget to update `libz3-dev` in the future... 😅

lib/runners/processor/cppcheck.rb
dev-yakuza: Thanks for your explanation 🙇‍♂️

lib/runners/processor/cppcheck.rb
```diff
@@ -65,18 +66,40 @@ def language
       Array(lang ? "--language=#{lang}" : nil)
     end
 
+    def bug_hunting
+      Array(do_bug_hunting? ? "--bug-hunting" : nil)
+    end
+
+    def do_bug_hunting?
```
ybiquitous, line 73: [suggestion] How about the following refactoring? `do_bug_hunting?` can be removed. 💪 

```diff
diff --git a/lib/runners/processor/cppcheck.rb b/lib/runners/processor/cppcheck.rb
index 35d9f563..0f91bf53 100644
--- a/lib/runners/processor/cppcheck.rb
+++ b/lib/runners/processor/cppcheck.rb
@@ -67,20 +67,15 @@ module Runners
     end
 
     def bug_hunting
-      Array(do_bug_hunting? ? "--bug-hunting" : nil)
-    end
-
-    def do_bug_hunting?
-      return config_linter[:'bug-hunting']
+      config_linter[:'bug-hunting'] ? ["--bug-hunting"] : nil
     end
 
     def run_analyzer
       results = Results::Success.new(guid: guid, analyzer: analyzer)
 
       args_normal = [*enable, *std, *addon]
-      args_bughunting = do_bug_hunting? ? [ *bug_hunting ] : nil
 
-      [args_normal, args_bughunting].compact.each do |args|
+      [args_normal, bug_hunting].compact.each do |args|
         ret = step_analyzer(results, *args)
         if ret
           return ret
```

lib/runners/processor/cppcheck.rb
```diff
@@ -65,18 +66,40 @@ def language
...
-        *std,
         *project,
         *language,
-        *target,
-        *addon,
+        *args,
+        *target
       )
 
       if status.exitstatus == 1 && stdout.strip == "cppcheck: error: could not find or open any of the paths given."
```
ybiquitous, line 105: If the first step succeeds with "No linting files", it seems that the second step does not be executed. 🤔

lib/runners/processor/cppcheck.rb
ybiquitous: [suggestion] I think a bit confusing about passing the `results` argument and returning `nil`.
So, how about the following refactoring?

- `step_analyzer` returns `Results::{Success,Failure}` object always.
- Merge issues in `run_analyzer`.

```diff
diff --git a/lib/runners/processor/cppcheck.rb b/lib/runners/processor/cppcheck.rb
index 35d9f563..7925bd19 100644
--- a/lib/runners/processor/cppcheck.rb
+++ b/lib/runners/processor/cppcheck.rb
@@ -75,22 +75,25 @@ module Runners
     end
 
     def run_analyzer
-      results = Results::Success.new(guid: guid, analyzer: analyzer)
-
       args_normal = [*enable, *std, *addon]
       args_bughunting = do_bug_hunting? ? [ *bug_hunting ] : nil
 
+      issues = []
       [args_normal, args_bughunting].compact.each do |args|
-        ret = step_analyzer(results, *args)
-        if ret
+        ret = step_analyzer(*args)
+        if ret.is_a? Results::Success
+          issues.push(*ret.issues)
+        else
           return ret
         end
       end
 
-      return results
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+        result.issues.push(*issues)
+      end
     end
 
-    def step_analyzer(results, *args)
+    def step_analyzer(*args)
       stdout, stderr, status = capture3(
         analyzer_bin,
         "--quiet",
@@ -120,13 +123,11 @@ module Runners
         return Results::Failure.new(guid: guid, analyzer: analyzer, message: "Invalid XML output!")
       end
 
-      results.tap do |result|
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
         parse_result(xml_output) do |issue|
           result.add_issue issue
         end
       end
-
-      return nil # return nil to indicate normal exit
     end
 
     # @see https://github.com/danmar/cppcheck/blob/master/man/manual.md#xml-output
```

ybiquitous:
`code_sniffer` CI failure does not seem due to this PR change. I'll look into it. 👀 
https://github.com/sider/runners/pull/1166/checks?check_run_id=749280019

ybiquitous:
I've fixed via #1178. Please merge or rebase `master`.

test/smokes/cppcheck/expectations.rb
suetsuguk: Having an in-depth study for cppcheck, I need to check the followings. It leads me to get the same issues.
- check for --bug-hunting with `mul1 = 1 / (dx * dx + dy * dy);`
- check for --addon=misra with `mul2 = 1 / dz; and mul3 = 1 / mul2;`
- normal check with `mul2 = 1 / dz;`

And also we need to cause at least two issues in order to confirm that the loop block works.
```
parse_result(xml_output) do |issue|
    result.add_issue issue
end
```

images/cppcheck/Dockerfile
```diff
@@ -9,15 +9,21 @@ FROM sider/devon_rex_python:2.19.0
...
+    cd /tmp && \
     curl -sSL https://github.com/danmar/cppcheck/archive/${CPPCHECK_VERSION}.tar.gz | tar -xz && \
     cd cppcheck-${CPPCHECK_VERSION} && \
     make install -s \
       MATCHCOMPILER=yes \
       FILESDIR=/usr/share/cppcheck \
       HAVE_RULES=yes \
-      CXXFLAGS="-O2 -DNDEBUG -Wall -Wno-sign-compare -Wno-unused-function" && \
+      USE_Z3=yes \
+      CPPFLAGS="-DUSE_Z3 -Wno-deprecated-declarations -Wno-unused-function" && \
```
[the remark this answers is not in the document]
suetsuguk: That's what I thought. But it did not. Passing the CPPFLAGS value to the make command as a command-line argument, the CPPFLAGS value in Makefile is overwritten by the argument. So we need to specify both `USE_Z3=yes` and `CPPFLAGS="-DUSE_Z3 foo bar"`.
See https://www.gnu.org/software/make/manual/html_node/Overriding.html#Overriding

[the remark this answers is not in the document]
yykamei: I understand. Thank you. This `USE_Z3` is only used for assigning `-lz3` to `LIBS`.

lib/runners/processor/cppcheck.rb
suetsuguk: That is what I thought. But with a lot of thinking, I decided to use the `nil`-return approach for the following reasons:
- The original `run_analyzer` method returns Results::Success() even when `cppcheck` exits with error. https://github.com/sider/runners/blob/1b9782f009f9cc977ff431bb0b7a1804b5be4dfe/lib/runners/processor/cppcheck.rb#L82-L85 I cannot distinguish between error and success from the Results::Success instance.
- I try to avoid a buffer array `issues = []` to hold temporary results which increases memory usage. Memory usage might increase only when copying a temporary array to a final result array. Though, the effect is negligible because Ruby has an efficient memory management.

lib/runners/processor/cppcheck.rb
suetsuguk: Seen from the error message, the second step will cause the same error if it is executed. I think it is reasonable to stop the second step.

lib/runners/processor/cppcheck.rb
ybiquitous: > * The original `run_analyzer` method returns Results::Success() even when `cppcheck` exits with error.

This case is not an error for Runners. Because the situation of "No linting files" can usually occur, e.g. a user may try to enable it.
So, returning `Results::Success` is **success**, and returning except `Results::Success` is **error**.

> * I try to avoid a buffer array `issues = []` to hold temporary results which increases memory usage.

As you said, I also think the memory influence is negligible. I've never seen a memory error due to too many issues.
So, what about re-considering when a performance issue will be raised in the future?

lib/runners/processor/cppcheck.rb
ybiquitous: Sorry, I understand your comment on https://github.com/sider/runners/pull/1166#discussion_r437138059.

A simple solution is to return another value instead of `Results::Success` as follows:

```diff
- return Results::Success.new(guid: guid, analyzer: analyzer)
+ return :no_target_files
```

The caller code looks like:

```ruby
ret = step_analyzer(*args)
case ret
when Results::Success
  issues.push(*ret.issues)
when :no_target_files
  return Results::Success.new(guid: guid, analyzer: analyzer)
else
  return ret
end
```

lib/runners/processor/cppcheck.rb
suetsuguk: Good catch!
In order to make variable names symmetrical (args_*), how about this?

```diff
diff --git a/lib/runners/processor/cppcheck.rb b/lib/runners/processor/cppcheck.rb
index 35d9f563..924aae6a 100644
--- a/lib/runners/processor/cppcheck.rb
+++ b/lib/runners/processor/cppcheck.rb
@@ -66,20 +66,17 @@ module Runners
       Array(lang ? "--language=#{lang}" : nil)
     end

-    def bug_hunting
-      Array(do_bug_hunting? ? "--bug-hunting" : nil)
+    def args_normal
+      [*enable, *std, *addon]
     end

-    def do_bug_hunting?
-      return config_linter[:'bug-hunting']
+    def args_bughunting
+      config_linter[:'bug-hunting'] ? ["--bug-hunting"] : nil
     end

     def run_analyzer
       results = Results::Success.new(guid: guid, analyzer: analyzer)

-      args_normal = [*enable, *std, *addon]
-      args_bughunting = do_bug_hunting? ? [ *bug_hunting ] : nil
-
       [args_normal, args_bughunting].compact.each do |args|
         ret = step_analyzer(results, *args)
         if ret
```

ybiquitous: Looks nice! 👍

suetsuguk:
> I've fixed via #1178. Please merge or rebase `master`.

Thank you for fixing the issue. Now CI passes.

lib/runners/processor/cppcheck.rb
suetsuguk: That makes good sense to me.
Fixed. commit 79c464fafa44d7a02ecb964a14bf0e25eeab2f9e

lib/runners/processor/cppcheck.rb
suetsuguk: Fixed. commit 4d28900281a1507df62d0c83ddc7eb6ca98271d0

suetsuguk:
> It looks good, but this PR contains the change of PMD Java. I think it should be excluded here.

I think changes for PMD Java appear due to merging the current master branch. I will `squash and merge` to merge only my changes to the master.
As for "Files changed" of this PR, I tried to change a base branch by the `Edit` button, but didn't. Do you know how to compare between the current master and this branch? I do not know which branch(or point of commit) GitHub compares to.

ybiquitous:
@suetsuguk Can I push the successfully merged commit (solves the conflict), OK?

`$ git diff master...` on my local machine:

<details>

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 37396215..3c2f4f3c 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - **Flake8** 3.8.1 -> 3.8.2 [#1127](https://github.com/sider/runners/pull/1127)
 - **Reek** 6.0.0 -> 6.0.1 [#1161](https://github.com/sider/runners/pull/1161)
 - **Checkstyle** 8.32 -> 8.33 [#1162](https://github.com/sider/runners/pull/1162)
+- **Cppcheck** Add `bug-hunting` option [#1166](https://github.com/sider/runners/pull/1166)
 
 ## 0.27.0
 
diff --git a/images/cppcheck/Dockerfile b/images/cppcheck/Dockerfile
index 67c20f2d..4e4d1af9 100644
--- a/images/cppcheck/Dockerfile
+++ b/images/cppcheck/Dockerfile
@@ -9,15 +9,21 @@ FROM sider/devon_rex_python:2.19.0
 # See https://github.com/danmar/cppcheck#gnu-make
 
 ARG CPPCHECK_VERSION=2.0
+ARG LIBZ3_VERSION=4.4.1-1~deb10u1
 USER root
-RUN cd /tmp && \
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libz3-dev=${LIBZ3_VERSION} && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    cd /tmp && \
     curl -sSL https://github.com/danmar/cppcheck/archive/${CPPCHECK_VERSION}.tar.gz | tar -xz && \
     cd cppcheck-${CPPCHECK_VERSION} && \
     make install -s \
       MATCHCOMPILER=yes \
       FILESDIR=/usr/share/cppcheck \
       HAVE_RULES=yes \
-      CXXFLAGS="-O2 -DNDEBUG -Wall -Wno-sign-compare -Wno-unused-function" && \
+      USE_Z3=yes \
+      CPPFLAGS="-DUSE_Z3 -Wno-deprecated-declarations -Wno-unused-function" && \
     rm -rf cppcheck-${CPPCHECK_VERSION}
 
 
diff --git a/images/cppcheck/Dockerfile.erb b/images/cppcheck/Dockerfile.erb
index 18cb35a0..1e3453ee 100644
--- a/images/cppcheck/Dockerfile.erb
+++ b/images/cppcheck/Dockerfile.erb
@@ -5,15 +5,21 @@ FROM sider/devon_rex_python:2.19.0
 # See https://github.com/danmar/cppcheck#gnu-make
 
 ARG CPPCHECK_VERSION=2.0
+ARG LIBZ3_VERSION=4.4.1-1~deb10u1
 USER root
-RUN cd /tmp && \
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libz3-dev=${LIBZ3_VERSION} && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    cd /tmp && \
     curl -sSL https://github.com/danmar/cppcheck/archive/${CPPCHECK_VERSION}.tar.gz | tar -xz && \
     cd cppcheck-${CPPCHECK_VERSION} && \
     make install -s \
       MATCHCOMPILER=yes \
       FILESDIR=/usr/share/cppcheck \
       HAVE_RULES=yes \
-      CXXFLAGS="-O2 -DNDEBUG -Wall -Wno-sign-compare -Wno-unused-function" && \
+      USE_Z3=yes \
+      CPPFLAGS="-DUSE_Z3 -Wno-deprecated-declarations -Wno-unused-function" && \
     rm -rf cppcheck-${CPPCHECK_VERSION}
 
 <%= render_erb 'images/Dockerfile.base.erb' %>
diff --git a/lib/runners/processor/cppcheck.rb b/lib/runners/processor/cppcheck.rb
index ddac98cd..7214d6d8 100644
--- a/lib/runners/processor/cppcheck.rb
+++ b/lib/runners/processor/cppcheck.rb
@@ -10,6 +10,7 @@ module Runners
           std: string?,
           project: string?,
           language: string?,
+          'bug-hunting': boolean?
         )
       }
 
@@ -65,23 +66,49 @@ module Runners
       Array(lang ? "--language=#{lang}" : nil)
     end
 
+    def args_normal
+      [*enable, *std, *addon]
+    end
+
+    def args_bughunting
+      config_linter[:'bug-hunting'] ? ["--bug-hunting"] : nil
+    end
+
     def run_analyzer
+      issues = []
+
+      [args_normal, args_bughunting].compact.each do |args|
+        ret = step_analyzer(*args)
+        case ret
+        when Results::Success
+          issues.push(*ret.issues)
+        when :no_target_files
+          return Results::Success.new(guid: guid, analyzer: analyzer)
+        else
+          return ret
+        end
+      end
+
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+        result.issues.push(*issues)
+      end
+    end
+
+    def step_analyzer(*args)
       stdout, stderr, status = capture3(
         analyzer_bin,
         "--quiet",
         "--xml",
         *ignore,
-        *enable,
-        *std,
         *project,
         *language,
-        *target,
-        *addon,
+        *args,
+        *target
       )
 
       if status.exitstatus == 1 && stdout.strip == "cppcheck: error: could not find or open any of the paths given."
         add_warning "No linting files."
-        return Results::Success.new(guid: guid, analyzer: analyzer)
+        return :no_target_files
       end
 
       unless status.success?
diff --git a/test/smokes/cppcheck/bug_hunting_addon/c_sample.c b/test/smokes/cppcheck/bug_hunting_addon/c_sample.c
new file mode 100644
index 00000000..116a7542
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_addon/c_sample.c
@@ -0,0 +1,15 @@
+
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_addon/sider.yml b/test/smokes/cppcheck/bug_hunting_addon/sider.yml
new file mode 100644
index 00000000..e071d377
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_addon/sider.yml
@@ -0,0 +1,4 @@
+linter:
+  cppcheck:
+    addon: misra
+    bug-hunting: true
diff --git a/test/smokes/cppcheck/bug_hunting_project/sider.yml b/test/smokes/cppcheck/bug_hunting_project/sider.yml
new file mode 100644
index 00000000..b9ab9190
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_project/sider.yml
@@ -0,0 +1,4 @@
+linter:
+  cppcheck:
+    project: test.vcxproj
+    bug-hunting: true
diff --git a/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_1.c++ b/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_1.c++
new file mode 100644
index 00000000..cb89d386
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_1.c++
@@ -0,0 +1,24 @@
+void bar()
+{
+    double v;
+
+    int ar[] = {1, 3, 5, 7, 9, 0, 11};
+    for(int &x : ar) {
+        v = 1 / x;
+    }
+}
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_2.c++ b/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_2.c++
new file mode 100644
index 00000000..44c78c38
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_2.c++
@@ -0,0 +1,26 @@
+
+
+void bar()
+{
+    double v;
+
+    int ar[] = {1, 3, 5, 7, 9, 0, 11};
+    for(int &x : ar) {
+        v = 1 / x;
+    }
+}
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_3.c++ b/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_3.c++
new file mode 100644
index 00000000..56bf3ccc
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_project/src/cpp_sample_3.c++
@@ -0,0 +1,28 @@
+
+
+
+
+void bar()
+{
+    double v;
+
+    int ar[] = {1, 3, 5, 7, 9, 0, 11};
+    for(int &x : ar) {
+        v = 1 / x;
+    }
+}
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_project/test.vcxproj b/test/smokes/cppcheck/bug_hunting_project/test.vcxproj
new file mode 100644
index 00000000..2ae83415
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_project/test.vcxproj
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|Win32">
+      <Configuration>Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <VCProjectVersion>15.0</VCProjectVersion>
+    <ProjectGuid>{7319858B-261C-4F0D-B022-92BB896242DD}</ProjectGuid>
+    <RootNamespace>helloworld</RootNamespace>
+    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
+    <UseDebugLibraries>true</UseDebugLibraries>
+    <PlatformToolset>v141</PlatformToolset>
+    <CharacterSet>MultiByte</CharacterSet>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
+    <UseDebugLibraries>false</UseDebugLibraries>
+    <PlatformToolset>v141</PlatformToolset>
+    <WholeProgramOptimization>true</WholeProgramOptimization>
+    <CharacterSet>MultiByte</CharacterSet>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
+    <UseDebugLibraries>true</UseDebugLibraries>
+    <PlatformToolset>v141</PlatformToolset>
+    <CharacterSet>MultiByte</CharacterSet>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
+    <UseDebugLibraries>false</UseDebugLibraries>
+    <PlatformToolset>v141</PlatformToolset>
+    <WholeProgramOptimization>true</WholeProgramOptimization>
+    <CharacterSet>MultiByte</CharacterSet>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
+  <ImportGroup Label="ExtensionSettings">
+  </ImportGroup>
+  <ImportGroup Label="Shared">
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup />
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+    <ClCompile>
+      <WarningLevel>Level3</WarningLevel>
+      <Optimization>Disabled</Optimization>
+      <SDLCheck>true</SDLCheck>
+      <ConformanceMode>true</ConformanceMode>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <ClCompile>
+      <WarningLevel>Level3</WarningLevel>
+      <Optimization>Disabled</Optimization>
+      <SDLCheck>true</SDLCheck>
+      <ConformanceMode>true</ConformanceMode>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
+    <ClCompile>
+      <WarningLevel>Level3</WarningLevel>
+      <Optimization>MaxSpeed</Optimization>
+      <FunctionLevelLinking>true</FunctionLevelLinking>
+      <IntrinsicFunctions>true</IntrinsicFunctions>
+      <SDLCheck>true</SDLCheck>
+      <ConformanceMode>true</ConformanceMode>
+    </ClCompile>
+    <Link>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+    </Link>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <ClCompile>
+      <WarningLevel>Level3</WarningLevel>
+      <Optimization>MaxSpeed</Optimization>
+      <FunctionLevelLinking>true</FunctionLevelLinking>
+      <IntrinsicFunctions>true</IntrinsicFunctions>
+      <SDLCheck>true</SDLCheck>
+      <ConformanceMode>true</ConformanceMode>
+    </ClCompile>
+    <Link>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+    </Link>
+  </ItemDefinitionGroup>
+  <ItemGroup>
+    <ClCompile Include="src/cpp_sample_1.c++" />
+    <ClCompile Include="src/cpp_sample_3.c++" />
+  </ItemGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <ImportGroup Label="ExtensionTargets">
+  </ImportGroup>
+</Project>
diff --git a/test/smokes/cppcheck/bug_hunting_target/sider.yml b/test/smokes/cppcheck/bug_hunting_target/sider.yml
new file mode 100644
index 00000000..ba759474
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_target/sider.yml
@@ -0,0 +1,9 @@
+linter:
+  cppcheck:
+    root_dir: src
+    bug-hunting: true
+    language: c++
+    target:
+      - cpp_sample_2.c++
+      - bar
+    ignore: bar/foo
diff --git a/test/smokes/cppcheck/bug_hunting_target/src/bar/c_sample.c b/test/smokes/cppcheck/bug_hunting_target/src/bar/c_sample.c
new file mode 100644
index 00000000..14e482c4
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_target/src/bar/c_sample.c
@@ -0,0 +1,23 @@
+class Color
+{
+    public:
+        int r;
+        int g;
+        int b;
+};
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2, mul3;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+    mul3 = 1 / mul2;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_target/src/bar/cpp_sample.c++ b/test/smokes/cppcheck/bug_hunting_target/src/bar/cpp_sample.c++
new file mode 100644
index 00000000..56bf3ccc
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_target/src/bar/cpp_sample.c++
@@ -0,0 +1,28 @@
+
+
+
+
+void bar()
+{
+    double v;
+
+    int ar[] = {1, 3, 5, 7, 9, 0, 11};
+    for(int &x : ar) {
+        v = 1 / x;
+    }
+}
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_target/src/bar/foo/cpp_sample_0.c++ b/test/smokes/cppcheck/bug_hunting_target/src/bar/foo/cpp_sample_0.c++
new file mode 100644
index 00000000..cb89d386
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_target/src/bar/foo/cpp_sample_0.c++
@@ -0,0 +1,24 @@
+void bar()
+{
+    double v;
+
+    int ar[] = {1, 3, 5, 7, 9, 0, 11};
+    for(int &x : ar) {
+        v = 1 / x;
+    }
+}
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_target/src/cpp_sample_1.c++ b/test/smokes/cppcheck/bug_hunting_target/src/cpp_sample_1.c++
new file mode 100644
index 00000000..cb89d386
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_target/src/cpp_sample_1.c++
@@ -0,0 +1,24 @@
+void bar()
+{
+    double v;
+
+    int ar[] = {1, 3, 5, 7, 9, 0, 11};
+    for(int &x : ar) {
+        v = 1 / x;
+    }
+}
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/bug_hunting_target/src/cpp_sample_2.c++ b/test/smokes/cppcheck/bug_hunting_target/src/cpp_sample_2.c++
new file mode 100644
index 00000000..44c78c38
--- /dev/null
+++ b/test/smokes/cppcheck/bug_hunting_target/src/cpp_sample_2.c++
@@ -0,0 +1,26 @@
+
+
+void bar()
+{
+    double v;
+
+    int ar[] = {1, 3, 5, 7, 9, 0, 11};
+    for(int &x : ar) {
+        v = 1 / x;
+    }
+}
+
+double calc(double x0, double x1, double y0, double y1)
+{
+    double dx, dy, dz;
+    double mul1, mul2;
+
+    dx = x1 - x0;
+    dy = y1 - y0;
+    mul1 = dz;
+
+    mul1 = 1 / (dx * dx + dy * dy);
+    mul2 = 1 / dz;
+
+    return mul1;
+}
diff --git a/test/smokes/cppcheck/expectations.rb b/test/smokes/cppcheck/expectations.rb
index eee0c1c6..e5a00066 100644
--- a/test/smokes/cppcheck/expectations.rb
+++ b/test/smokes/cppcheck/expectations.rb
@@ -492,3 +492,243 @@ s.add_test(
   ],
   analyzer: { name: "Cppcheck", version: "2.0" }
 )
+
+s.add_test(
+  "bug_hunting_target",
+  type: "success",
+  issues: [
+    {
+      id: "bughuntingDivByZero",
+      path: "src/bar/cpp_sample.c++",
+      location: { start_line: 11 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZero",
+      path: "src/cpp_sample_2.c++",
+      location: { start_line: 9 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "src/bar/c_sample.c",
+      location: { start_line: 18 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "src/bar/c_sample.c",
+      location: { start_line: 20 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "src/bar/cpp_sample.c++",
+      location: { start_line: 24 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "src/cpp_sample_2.c++",
+      location: { start_line: 22 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/bar/c_sample.c",
+      location: { start_line: 16 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/bar/c_sample.c",
+      location: { start_line: 19 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/bar/cpp_sample.c++",
+      location: { start_line: 22 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/bar/cpp_sample.c++",
+      location: { start_line: 25 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/cpp_sample_2.c++",
+      location: { start_line: 20 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/cpp_sample_2.c++",
+      location: { start_line: 23 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    }
+  ],
+  analyzer: { name: "Cppcheck", version: "2.0" }
+)
+
+s.add_test(
+  "bug_hunting_project",
+  type: "success",
+  issues: [
+    {
+      id: "bughuntingDivByZero",
+      path: "src/cpp_sample_1.c++",
+      location: { start_line: 7 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZero",
+      path: "src/cpp_sample_3.c++",
+      location: { start_line: 11 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "src/cpp_sample_1.c++",
+      location: { start_line: 20 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "src/cpp_sample_3.c++",
+      location: { start_line: 24 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/cpp_sample_1.c++",
+      location: { start_line: 18 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/cpp_sample_1.c++",
+      location: { start_line: 21 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/cpp_sample_3.c++",
+      location: { start_line: 22 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "src/cpp_sample_3.c++",
+      location: { start_line: 25 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    }
+  ],
+  analyzer: { name: "Cppcheck", version: "2.0" }
+)
+
+s.add_test(
+  "bug_hunting_addon",
+  type: "success",
+  issues: [
+    {
+      id: "bughuntingDivByZeroFloat",
+      path: "c_sample.c",
+      location: { start_line: 12 },
+      message: "There is division, cannot determine that there can't be a division by zero.",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "369", location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "misra-c2012-10.4",
+      path: "c_sample.c",
+      location: { start_line: 12 },
+      message: "misra violation (use --rule-texts=<file> to get proper output)",
+      links: [],
+      object: { severity: "style", verbose: nil, inconclusive: false, cwe: nil, location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "misra-c2012-12.3",
+      path: "c_sample.c",
+      location: { start_line: 5 },
+      message: "misra violation (use --rule-texts=<file> to get proper output)",
+      links: [],
+      object: { severity: "style", verbose: nil, inconclusive: false, cwe: nil, location_info: nil },
+      git_blame_info: nil
+    },
+    {
+      id: "uninitvar",
+      path: "c_sample.c",
+      location: { start_line: 10 },
+      message: "Uninitialized variable: dz",
+      links: [],
+      object: { severity: "error", verbose: nil, inconclusive: false, cwe: "457", location_info: nil },
+      git_blame_info: nil
+    }
+  ],
+  analyzer: { name: "Cppcheck", version: "2.0" }
+)
diff --git a/test/smokes/cppcheck/single_target/sider.yml b/test/smokes/cppcheck/single_target/sider.yml
index 761b73cf..08bc59f7 100644
--- a/test/smokes/cppcheck/single_target/sider.yml
+++ b/test/smokes/cppcheck/single_target/sider.yml
@@ -1,3 +1,4 @@
 linter:
   cppcheck:
     target: "autoVariables.c"
+    bug-hunting: false
```

</details>

suetsuguk:
> @suetsuguk Can I push the successfully merged commit (solves the conflict), OK?
Yes, please.

ybiquitous:
The PR diff seems good. I'll merge this when the CI is all green. 👍 

@suetsuguk Can you update the document, please?
https://help.sider.review/tools/cplusplus/cppcheck

ybiquitous:
We can ignore Codacy issues.

suetsuguk:
> @suetsuguk Can you update the document, please?
> https://help.sider.review/tools/cplusplus/cppcheck

Yes, I'm working on it in the sider-docs depot.

ybiquitous:
Thank you! 😄

PR #1200, opened 2020-06-15
[PMD CPD] Analyze multiple languages by default

### LINKED ISSUE #1181: [PMD CPD] Analyze multiple languages by default
Currently, users can specify only one language via `sider.yml` (default: `java`).
This is not useful, so I think it's better to analyze multiple languages by default.

https://github.com/sider/runners/blob/01fae1b32766810284c23092970aabb2e8326971/lib/runners/processor/pmd_cpd.rb#L10

suetsuguk:
Fix #1181

- [x] Run `bundle exec rake readme:generate` and commit the updated `README.md`.
- [x] Write a [document](https://github.com/sider/sider-docs) and link the PR sider/sider-docs#445


test/smokes/pmd_cpd/expectations.rb
```diff
@@ -1277,3 +1277,1380 @@
...
+          {
+            id: "f596d2f766fb7b912331c3691b493b5d028f8ce3",
+            path: "foo/hello_cpp.cpp",
+            start_line: 3,
+            start_column: 1,
+            end_line: 17,
+            end_column: 1
+          }
+        ],
+        codefragment: %[void print_hello();
```
ybiquitous, line 1317: [suggestion] I think that too many lines of `codefragment` make it hard for us to read the test code.
Because we can test the content of `codefragment` in other test cases, how about using regex as follows?

```ruby
codefragment: /void print_hello();/,
```

lib/runners/processor/pmd_cpd.rb
```diff
@@ -50,12 +51,17 @@ def analyze(changes)
...
+      languages.each do |language|
+        stdout, stderr = capture3!(analyzer_bin, *cli_options(language))
+        raise_warnings(stderr)
+        ret = construct_result(stdout)
+        issues.push(*ret)
+      end
 
       Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
-        construct_result(result, stdout)
+        issues.each { |i| result.add_issue(i) }
```
yykamei, line 64: [nitpick] Thanks to 0fdb3a62, `add_issue` can multiple issues at one time.

```suggestion
        result.add_issue(*issues)
```

ybiquitous:
It's strange that the CI via Actions does not run... 🤔

test/smokes/pmd_cpd/expectations.rb
suetsuguk: Good catch!
I never thought of regex capabilities.
Fixed. commit 07a88f372f1496f2f889df688102c6670a05a1b4

lib/runners/processor/pmd_cpd.rb
suetsuguk: LGTM. Thank you for letting me know.
Fixed. commit 15d456b4ba1fbfb2197c404402fc500a530a5d64

lib/runners/processor/pmd_cpd.rb
```diff
@@ -114,9 +124,9 @@ def create_issue_object(elem_dupli, files)
         id: f[:id],
         path: f[:path].to_s,
         start_line: f[:location].start_line,
-        start_column: f[:location].start_column,
+        start_column: f[:location].start_column.to_i,
         end_line: f[:location].end_line,
-        end_column: f[:location].end_column,
+        end_column: f[:location].end_column.to_i,
```
ybiquitous, line 129: [suggestion] I'm concerned that `start_column` and `end_column` can be `nil` because `nil.to_i` returns `0`:

```
$ irb
irb(main):001:0> nil.to_i
=> 0
```

What case that `start_column` or `end_column` are  `nil`?

test/smokes/pmd_cpd/option_multiple_languages_invalid/sider.yml
```diff
@@ -0,0 +1,5 @@
+linter:
+  pmd_cpd:
+    minimum-tokens: 15
+    language: [tcl, cs]
+    ignore-usings: true
```
ybiquitous, line 5: [question] This case seems that there are no errors or warnings. Can you tell me the reason for to name as `option_multiple_languages_invalid`?

lib/runners/processor/pmd_cpd.rb
suetsuguk: There are two cases when `start_column` and `end_column` are `nil`.

1. Analyze PHP files
2. Try to analyze unsupported languages

For both cases analysis is executed successfully, but the `column` and `endcolumn` attributes do not appear in the result XML file.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<pmd-cpd>
   <duplication lines="9" tokens="20">
      <file endline="9" line="1"
            path="./src/bar/hello_tcl.tcl"/>
      <file endline="9" line="1"
            path="./src/foo/hello_tcl.tcl"/>
      <codefragment><![CDATA[proc main {} {
    print_hello ;
}

proc print_hello {} {
    puts "Hello, world!"
}

main]]></codefragment>
   </duplication>
</pmd-cpd>
```

ybiquitous: I understand it. 👍 
The `location` field of the analysis result allows the data format:

```
start_line: integer, end_line: integer, start_column: nil, end_column: nil
```

https://github.com/sider/runners/blob/da8de9370e69046f5f515aeeacacb44cf229dff8/lib/runners/schema/result.rb#L7

So, I think it better to change from `integer` to `integer?` instead of `.to_i` as follows:

```ruby
let :issue, object(
  # ...
  files: array(object(
    id: string,
    path: string,
    start_line: integer,
    start_column: integer?,
    end_line: integer,
    end_column: integer?
  )),
  # ...
)
```

test/smokes/pmd_cpd/option_multiple_languages_invalid/sider.yml
suetsuguk: I intended to test an unsupported language. When I specified unsupported `tcl` and tried to analyze, but analysis is executed successfully against my expectations. The return code is `0` (success) and the result XML goes to https://github.com/sider/runners/pull/1200#discussion_r440671965. When I also set a random string to the `language` option, analysis is also executed successfully with return code `0`. But the result XML is empty.

```shel
$ ~/work/pmd-bin-6.24.0/bin/run.sh cpd --format xml --minimum-tokens 15 --failOnViolation false --files . --language asdfasdf
<?xml version="1.0" encoding="UTF-8"?>
<pmd-cpd/>
$ echo $?
0
```

I assume that the PMD CPD tries to find files with extension `(--language) tcl` and analyze them after applying a simple tokenizer.

lib/runners/processor/pmd_cpd.rb
suetsuguk: I thought the same idea. But my concern was a nullable integer might affect the `sideci` code. Do you think `sideci` can accept it?

test/smokes/pmd_cpd/option_multiple_languages_invalid/sider.yml
ybiquitous: I think this behavior is a bug of PMD CPD (has an issue been reported?). 😅 

Then, I think we need to fix this behavior on Sider. If not, users would be very confused. For example, here is one way:

```ruby
Schema = StrongJSON.new do
  let :available_languages, enum(
    literal("apex"),
    literal("cpp"),
    literal("cs"),
    # ...
  )
  
  # ...

  language: enum?(available_languages, array(available_languages)),
  
  # ...
end
```

ybiquitous: I think better for Sider and PMD CPD users to open an issue and send a patch to PMD CPD. 😅

lib/runners/processor/pmd_cpd.rb
ybiquitous: No problem. `start_column` and `end_column` are not used yet. 👍

test/smokes/pmd_cpd/option_multiple_languages_invalid/sider.yml
suetsuguk: That is the intended behavior. The author says that CPD supports a so called "AnyLanguage" tokenizer, that is used as a fallback.
- https://github.com/pmd/pmd/issues/2453#issuecomment-622456506
- Section "Beyond Java" at https://www.datarespons.com/code-quality-assurance-with-pmd/

Seen from the screen image at https://github.com/pmd/pmd/issues/2453#issue-608151326 , users can choose "Language: by extension..." and input a file extension to analyze.

Now we need to make a decision. Do we allow users to input any strings, or choose from supported languages as you suggested?
In order to avoid confusion, I prefer allowing users choosing from supported languages.

ybiquitous: Thanks for the detailed investigation. I deeply understood the problem.

Here seems the fallback code:

https://github.com/pmd/pmd/blob/ca0f6535a4992233715b7338120080bf0ecf317b/pmd-core/src/main/java/net/sourceforge/pmd/cpd/LanguageFactory.java#L78-L82

I also agree that this fallback mechanism is very confusing and we should address the problem on our side.
So, I recommend my suggestion in <https://github.com/sider/runners/pull/1200#discussion_r440698111>.

But, if there might be users who expect this fallback, we could output warnings when invalid languages are specified. For example:

`sider.yml`:

```yaml
linter:
  pmd_cpd:
    language: [javaa, rubyy, go]
```

warnings:

```
The specified language `javaa` is invalid, so the fallback to "any language" will be performed.
The specified language `rubyy` is invalid, so the fallback to "any language" will be performed.
```

But these warnings may confuse users... 🤔

suetsuguk: If we adopt the suggestion <https://github.com/sider/runners/pull/1200#discussion_r440698111>, it implies that we **do not** support the fallback mechanism.
So the warning message should be simple:

```
The specified language `javaa` is not supported.
```

We also need to write a note for the unsupported fallback mechanism in our document.

ybiquitous: Sorry, my comment is not enough. 🙇 

My suggestions are:
1. Restrict the `language` option. The way of https://github.com/sider/runners/pull/1200#discussion_r440698111
2. Allow the fallback and output warnings.

> We also need to write a note for the unsupported fallback mechanism in our document.

Yes. 👍 Honestly, I doubt that users will really expect the fallback. And such a fallback document can make users confused instead.

So, I prefer the **1** idea.

lib/runners/processor/pmd_cpd.rb
suetsuguk: Fixed. commit 2ee275fc529e4441be1253f4fc66e380d2e4d7d6

test/smokes/pmd_cpd/option_multiple_languages_invalid/sider.yml
suetsuguk: Fixed. commit bc3ed6d694b741bc889a51f074f4791b04bb7683
I will mention **nothing** about fallback in the document.

ybiquitous:
We can ignore the Codacy failure.

PR #1213, opened 2020-06-23
[Rails Best Practices] Improvements

ybiquitous:

- Output the result STDOUT to the trace.
- Use `YAML.safe_load` instead of `YAML.load`.
- Optimize the result location.

None.

> Check the following items.


lib/runners/processor/rails_best_practices.rb
```diff
@@ -115,29 +115,25 @@ def run_analyzer(options)
...
-          loc = Location.new(
-            start_line: yaml.fetch(:line_number).to_i,
-            start_column: 0,
-            end_line: yaml.fetch(:line_number).to_i,
-            end_column: 0
-          )
+        #       We want to just parse a YAML, so we remove YAML tags before passing.
+        output = read_report_file
+        output.gsub!('- !ruby/object:RailsBestPractices::Core::Error', '-')
+        YAML.safe_load(output, symbolize_names: true, filename: report_file).each do |issue|
```
ybiquitous, line 130: [note] `filename:` is useful to debug when an error occurs. An occurred exception message includes the filename.

```ruby
YAML.safe_load ":", filename: "foo.yml"
#=> Psych::SyntaxError ((foo.yml): did not find expected key while parsing a block mapping at line 1 column 1)

YAML.safe_load ":"
#=> Psych::SyntaxError ((<unknown>): did not find expected key while parsing a block mapping at line 1 column 1)
```

See https://ruby-doc.org/stdlib-2.7.1/libdoc/psych/rdoc/Psych.html#safe_load-method

lib/runners/processor/rails_best_practices.rb
```diff
@@ -115,29 +115,25 @@ def run_analyzer(options)
...
         )
         # NOTE: rails_best_practices returns top-level YAML array with tags.
-        #       We want to just parse a YAML, so we remove YAML tags before passing to `YAML.load`.
-        YAML.load(stdout.gsub('- !ruby/object:RailsBestPractices::Core::Error', '-'), symbolize_names: true).each do |yaml|
-          loc = Location.new(
-            start_line: yaml.fetch(:line_number).to_i,
-            start_column: 0,
-            end_line: yaml.fetch(:line_number).to_i,
-            end_column: 0
-          )
```
ybiquitous, line 134: [note] The fields except for `start_line` are needless:

https://github.com/sider/runners/blob/b11e8317ec5a7cb3a38dd7a48b29054c70f2477b/lib/runners/schema/result.rb#L6-L8

I believe that this data format change will not cause breaking changes.

ybiquitous:
Thank you! 😄

PR #1216, opened 2020-06-24
[Rails Best Practices] Fix invalid `line_number`

ybiquitous:

The following error occurs. This may be a bug of RBP, but Runners keep it as a warning.

```
ArgumentError: invalid value for Integer(): "call"
```

None.

> Check the following items.

### DIFF
--- a/lib/runners/processor/rails_best_practices.rb
+++ b/lib/runners/processor/rails_best_practices.rb
@@ -128,9 +128,19 @@ module Runners
         output = read_report_file
         output.gsub!('- !ruby/object:RailsBestPractices::Core::Error', '-')
         YAML.safe_load(output, symbolize_names: true, filename: report_file).each do |issue|
+          # HACK: `line_number` sometimes is not a number.
+          line_number = issue.fetch(:line_number)
+          start_line = Integer(line_number, exception: false)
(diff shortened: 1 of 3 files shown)


ybiquitous:
I want to add a smoke test case for this bug, but I could not reproduce it... 😓

lib/runners/processor/rails_best_practices.rb
```diff
@@ -128,9 +128,19 @@ def run_analyzer(options)
         output = read_report_file
         output.gsub!('- !ruby/object:RailsBestPractices::Core::Error', '-')
         YAML.safe_load(output, symbolize_names: true, filename: report_file).each do |issue|
+          # HACK: `line_number` sometimes is not a number.
+          line_number = issue.fetch(:line_number)
+          start_line = Integer(line_number, exception: false)
```
ybiquitous, line 133: [note] `Integer(num, exception: false)` can return `nil` instead of raising an error.

https://ruby-doc.org/core-2.7.1/Kernel.html#Integer-method

ybiquitous:
@yykamei Sorry, can you please review this?

ybiquitous:
Thank you!

PR #1254, opened 2020-07-02
Support Clang-Tidy

### LINKED ISSUE #996: Support clang-tidy
## Information

- Website: https://clang.llvm.org/extra/clang-tidy/
- Source code: https://reviews.llvm.org/diffusion/L/browse/clang-tools-extra/trunk/clang-tidy/cppcoreguidelines/
- Language: C++

## Why do you need the tool?

The development members of our company mainly use C++ for our products.
The Core Guidelines provided by Standard C++ Foundation will help them to use modern C++ effectively.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

## Additional context

There are two possible candidates of the Core Guidelines implementation.

- clang-tidy
- Visual Studio built-in checker https://docs.microsoft.com/cpp/code-quality/using-the-cpp-core-guidelines-checkers?view=vs-2019

### LINKED ISSUE #1066: Support build management tool for C/C++
We are considering the support of clang-tidy #996.
Clang-tidy is able to analyze a single file by passing the file name as an command-line option.
In addition to that, the clang-tidy requires a compiler command database so as to read header files.
There two options to specify the command database; first, the command-line option -I and second, compile_commands.json.
In any case, we need to deploy the libraries to be included by the target source code on the runner instance.

CMake is a possible solution. If users load libraries by setting up ExternalProject_Add() in CMakeLists.txt properly, the runners can create the entire build environment.

Another simple solution is for the user to list development packages so that the runner deploys them on the runner instance.

suetsuguk:
Fix #996
Fix #1066

- [x] Run `bundle exec rake readme:generate` and commit the updated `README.md`.
- [x] Add a new tool to the [Dependabot configuration](https://github.com/sider/runners/blob/master/.github/dependabot.yml) if needed.
- [x] Write a [document](https://github.com/sider/sider-docs) and link the PR sider/sider-docs#469


lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,70 @@
...
+      stdout.each_line do |line|
+        # group: 1      2      3         4           5          6
+        #        <path>:<line>:<column>: <severity>: <message> [<id>]
+        match = line.strip.match(/^(?<path>.+):(?<line>\d+):(?<column>\d+): (?<severity>[^:]+): (?<message>.+) \[(?<id>[^\[]+)\]$/)
+        if match
+          start_line = Integer(match[:line])
+          start_column = Integer(match[:column])
+          issues << Issue.new(
+            path: relative_path(match[:path]),
+            location: Location.new(start_line: start_line, start_column: start_column, end_line: start_line, end_column: start_column),
```
ybiquitous, line 56: The following seems more natural to me. But currently, we cannot do it.

```suggestion
            location: Location.new(start_line: start_line, start_column: start_column),
```

I suggested allowing this form above via #1257.

suetsuguk: That's exactly what I want. Thank you.

images/clang_tidy/Dockerfile.erb
```diff
@@ -0,0 +1,11 @@
+FROM sider/devon_rex_base:2.21.0
+
+USER root
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends clang-tidy && \
```
yykamei, line 5: Is it OK to use this version?

```
root@7441fae702d3:/# clang-tidy --version
LLVM (http://llvm.org/):
  LLVM version 7.0.1
  
  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: skylake
```

lib/runners/processor/clang_tidy.rb
ybiquitous: @suetsuguk I've merged #1257. 💪

suetsuguk: Thanks to #1257, I have removed end_line and end_column from my code at 023c68d. Looks nice. Thank you for your great work.

images/clang_tidy/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+# NOTE: DO *NOT* EDIT THIS FILE. IT IS GENERATED.
+# PLEASE UPDATE Dockerfile.erb INSTEAD OF THIS FILE.
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+FROM sider/devon_rex_base:2.21.0
+
+USER root
```
suetsuguk, line 7: > __[Goodcheck]__ Try to avoid using the `root` user as much as possible. ([view](https://sider.review/gh/repos/202486306/pulls/1254#issue-4082550))
>
> <details>
>
> Each runner is executed by the non-root user for security reasons.
> Using the `USER` command leads to more layers and bigger image size.
>
> | Rule |
> | ---- |
> | `sider.runners.root_user_in_dockerfile` |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1254#issue-4082550) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

以下の理由でこのIssueをクローズしました：
- When you need to use system-wide commands such as `apt-get`.

images/clang_tidy/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN apt-get update && \
```
suetsuguk, line 8: > __[hadolint]__ Delete the apt-get lists after installing something ([view](https://sider.review/gh/repos/202486306/pulls/1254#issue-4099674))
>
> <details>
>
> | Rule | Severity |
> | ---- | -------- |
> | `DL3009` | `info` |
>
> References:
> - <https://github.com/hadolint/hadolint/wiki/DL3009>
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1254#issue-4099674) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

We need the apt-get lists when executing apt-get in clang-tidy.rb. If the apt-get lists don't exist, apt-get fails.

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,90 @@
...
+    def run_analyzer(changes)
+      issues = []
+
+      changes
+        .changed_paths
+        .select { |path| VALID_EXTENSIONS.include?(path.extname.downcase) }
+        .map{ |path| relative_path(working_dir / path, from: current_dir) }
+        .each do |path|
+          stdout, stderr = capture3!(analyzer_bin, path.to_s, "--",
+            is_success: ->(status) { [0, 1].include?(status.exitstatus) })
```
suetsuguk, line 56: > __[Querly]__ Note that `Process#exitstatus` can return `nil`. ([view](https://sider.review/gh/repos/202486306/pulls/1254#issue-4099677))
>
> <details>
>
> `Process#exitstatus` returns `nil` when the process does not exit normally.
> This means the process may be abort(2)-ed or core-dumped.
> If you want to know if it is aborted, consider using also `Process#exited?`.
>
> See also:
> * https://ruby-doc.org/core/Process/Status.html#method-i-exitstatus
>
> | Rule |
> | ---- |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1254#issue-4099677) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

If `exitstatus` is `nil`, the `is_success` returns `false`([0, 1].include?(nil) => false), then `capture3!()` throws `ExecError` as expected.

images/clang_tidy/Dockerfile.erb
ybiquitous: Additionally, we need to track the future updates of Clang-Tidy. So, we usually avoid installing an implicit version of a tool.

My questions:

- Can we install the fixed version of Clang-Tidy? (e.g. v2.3.4)
- Can we track the future changes of Clang-Tidy? (e.g. changelog or release note etc.)

suetsuguk:
As discussed before, we will deploy frequently used packages on our Docker image in advance.
Here are some thoughts on how to choose packages to deploy: Search for packages named "lib*-dev" in the result of Debian Popularity Contest [https://popcon.debian.org/]. Select them from the top until the sum of file size reaches about 300MB. They might be 20 to 30 packages.

images/clang_tidy/Dockerfile.erb
suetsuguk: - Yes, this is how to install the package with exact version: `apt-get install clang-tidy=1:7.0-47`
- Yes, we can follow the updates and also get the e-mail notification at https://tracker.debian.org/

ybiquitous: Thanks for the response. One more question.

If users request to use the latest version of Clang-Tidy, can we address the request soon?
I've heard the package versions by Debian are old, so I'm worried about it. 🤔

suetsuguk: That's true. We provide the old and stable version of clang-tidy 7.0 to users as long as we adopt the Docker 'debian:buster' image.
The Debian project provides mainly three releases: stable, testing and unstable. https://www.debian.or.jp/using/release.html
The testing release called 'bullseye' contains the clang-tidy 9.0. The unstable release 'sid' also contains the clang-tidy 9.0.
I think we can provide 'bullseye' to users who want to use newer version of clang-tidy. In that case, we will have to change the Docker image to 'debian:bullseye' accordingly.

ybiquitous: Umm..., is the latest version of clang-tidy **10.0.0**?
https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/ReleaseNotes.html

**9.0** seems too old for some users. Can we install clang-tidy in other ways?

yykamei: I think so. That's why I asked about the version. Perhaps, we have to compile from the source.

suetsuguk: Clang-tidy is a part of LLVM which depends on lots of packages. Even if we compile it on 'bullseye' or 'sid', we may not be able to get sufficient versions of dependent packages. Alternatively, we can use the Debian 'experimental' release as a last resort. The 'experimental' release is a rolling development version, but it contains the latest version of Clang-Tidy 10.
Anyway, we need to try to build LLVM on any release of Debian.

suetsuguk: I found the clang-tidy-**10** package on `sid`. That could be a possible solution.

ybiquitous: Can we extract the binary of clang-tidy from somewhere, without installing it from the source?

ybiquitous: Because we have strongly depended on the base image defined sider/devon_rex, I think it too big cost to change the mechanism only for clang-tidy... 😓 

https://github.com/sider/devon_rex/blob/8df9de0038460fe23c76d47128057f866754fc83/base/Dockerfile#L5

suetsuguk: Here is an outline to get the binary:
1. `wget` the clang-tidy deb file from the Debian package server.
2. `dpkg-deb -x the_deb_file my_dir` extracts files into my_dir

However, we need to prepare all dependent packages accordingly. Otherwise, the single clang-tidy binary fails to run.

suetsuguk: We can upgrade our `buster` image to `bullseye` with Dockerfile of clang-tidy only as:
1. Change the word `buster` to `bullseye` in /etc/apt/sources.list with `sed`
2. Invoke commands: `apt-get update`, `apt-get upgrade` and `apt-get dist-upgrade`

I will see what's going on.

ybiquitous: Umm..., I think the Debian-based installation **itself** can be a problem.

Can we upgrade it when a new version of clang-tidy will be released and the Debian clang-tidy package will not be upgraded for a while?

yykamei: I search a bit on the internet and found out there seemed not to be available binaries anywhere. Again, I think we should compile, unfortunately.

ybiquitous: How long the compile time? I'm concerned it may have an influence on the CI time.

yykamei: I'm compiling on my local machine. Would you wait for a minute...

My machine spec: Thinkpad X1 Carbon gen7

```console
$ cat /proc/cpuinfo
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 142
model name      : Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
stepping        : 12
microcode       : 0xd6
cpu MHz         : 1800.001
cache size      : 8192 KB
...
```


```console
$ cat /proc/meminfo
MemTotal:       16068420 kB
MemFree:         2043304 kB
MemAvailable:    3478268 kB
Buffers:            3264 kB
Cached:          4165056 kB
SwapCached:            0 kB
...
```

suetsuguk: > Can we upgrade it when a new version of clang-tidy will be released and the Debian clang-tidy package will not be upgraded for a while?
Yes, I agree.

For reference, I have modified the clang-tidy Dockerfile to upgrade to `bulleye` and install clang-tidy-10.
```
root@41475902c5b5:/home/analyzer_runner# clang-tidy-10
Error: no input files specified.
USAGE: clang-tidy-10 [options] <source0> [... <sourceN>]
```

yykamei:
```
cmake --build .  23624.74s user 1239.76s system 734% cpu 56:24.39 total
```

:thinking: Using binary Debian sid may be better.

ybiquitous: It sounds good to me if we use `sid` (unstable version):

```
deb http://apt.llvm.org/unstable/ llvm-toolchain-10 main
deb-src http://apt.llvm.org/unstable/ llvm-toolchain-10 main
```

https://apt.llvm.org/

suetsuguk: For reference, I have modified the clang-tidy Dockerfile to upgrade to `bullseye` and install clang-tidy-10.
```
root@41475902c5b5:/home/analyzer_runner# clang-tidy-10
Error: no input files specified.
USAGE: clang-tidy-10 [options] <source0> [... <sourceN>]
```

```Dockerfile
USER root
RUN sed -i -e 's/buster main/bullseye main/g' -e 's/buster\/updates/bullseye-security\/updates/g' -e 's/buster-updates/bullseye-updates/g' /etc/apt/sources.list && \
    apt-get update -y && \
    apt-get upgrade -y && \
    apt-get dist-upgrade -y && \
    cat ${RUNNER_USER_HOME}/packages.list | xargs apt-get install -y --no-install-recommends && \
    apt-get clean && \
    echo "analyzer_runner ALL=NOPASSWD: /usr/bin/apt-get" >> /etc/sudoers
USER $RUNNER_USER
```

```shell
$ bundle exec rake docker:build docker:smoke ANALYZER=clang_tidy
docker build --tag sider/runner_clang_tidy:dev --file images/clang_tidy/Dockerfile .
[+] Building 192.1s (6/12)
 => [internal] load build definition from Dockerfile                                                                                                                                                           2.1s
[+] Building 253.7s (13/13) FINISHED
```

Smoke tests are passed.

ybiquitous: Looks good! Are there some considerations about using the unstable version like `bullseye` or `sid`?

suetsuguk: It seems that we can get the stable clang-10 packages for buster from https://apt.llvm.org/ you found.
It is better for us to use the packages rather than upgrading to `bullseye`.
I will try to install them and see what's going on.

yykamei: I'm concerned about using `apt-get dist-upgrade` because it tries to upgrade many packages system-wide. However, assets build on devon_rex_base are generated for Debian Buster.

I think preparing a base image, for example: `devon_rex_base_sid`, is better.

yykamei: Using https://apt.llvm.org/ is good :+1:

suetsuguk: Work fine! commit 81a14f888114c5b8f84d93248f58eaa9dcb90be8

```shell
kentaro@debian103amd64:~/depot/runners$ bundle exec rake docker:build docker:smoke ANALYZER=clang_tidy
docker build --tag sider/runner_clang_tidy:dev --file images/clang_tidy/Dockerfile .
[+] Building 59.2s (13/13) FINISHED
...
bin/runners_smoke sider/runner_clang_tidy:dev test/smokes/clang_tidy/expectations.rb
Running smoke tests...
...
??  Smoke tests passed! -- 6 passed, 0 failed, 6 total in 74.3 seconds
```

ybiquitous:
@suetsuguk Can you please merge the master branch to this branch?

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+          end
+        end
+
+      unless packages.empty?
+        capture3!("sudo", "apt-get", "install", "-y", "--no-install-recommends", *packages)
+      end
+    end
+
+    def analyzer_bin
+      "clang-tidy-10"
```
ybiquitous, line 48: [suggestion] When users see the `clang-tidy-10` command in the log, they may feel strange.
How about creating a symbolic link `clang-tidy`?

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      let :issue, object(
+        severity: string,
+      )
+    end
+
+    register_config_schema(name: :clang_tidy, schema: Schema.runner_config)
+
+    VALID_EXTENSIONS = [".c", ".cc", ".cpp", ".c++", ".cp", ".cxx"].freeze
+    GLOB_HEADERS = "**/*.{h,hpp,h++}".freeze
```
ybiquitous, line 21: [question] Is it reasonable to support the extensions by Linguist? 

https://github.com/github/linguist/blob/775b07d40c04ef6e0051a541886a8f1e30a892f4/lib/linguist/languages.yml#L532-L535

https://github.com/github/linguist/blob/775b07d40c04ef6e0051a541886a8f1e30a892f4/lib/linguist/languages.yml#L568-L584

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def analyze(changes)
+      deploy_packages
```
ybiquitous, line 24: [suggestion] It is better to use `setup`:

```suggestion
    def setup
      deploy_packages
      yield
    end

    def analyze(changes)
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+      run_analyzer(changes)
```
ybiquitous, line 25: [suggestion] How about inlining `run_analyzer`? Because `analyzer` is equal to `run_analyzer`.

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+            false
+          end
+        end
+
+      unless packages.empty?
+        capture3!("sudo", "apt-get", "install", "-y", "--no-install-recommends", *packages)
+      end
+    end
+
+    def analyzer_bin
```
ybiquitous, line 47: [suggestion] Can you put `analyzer_bin` above `analyze`? Because `analyzer_bin` is the overridden method. 

```ruby
def analyzer_bin
end

def analyze(changes)
end
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+      issues = []
+
+      stdout.each_line do |line|
+        # issue format
+        # <path>:<line>:<column>: <severity>: <message> [<id>]
+        match = line.strip.match(/^(?<path>.+):(?<line>\d+):(?<column>\d+): (?<severity>[^:]+): (?<message>.+) \[(?<id>[^\[]+)\]$/)
+        if match
+          issues << Issue.new(
+            path: relative_path(match[:path]),
+            location: Location.new(start_line: Integer(match[:line]), start_column: Integer(match[:column])),
```
ybiquitous, line 80: [suggestion] `Location.new` internally converts the arguments to `Integer`:

```suggestion
            location: Location.new(start_line: match[:line], start_column: match[:column]),
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+    private
+
+    def deploy_packages
+      # select "lib*" and report others as warning for security concerns
+      packages = Array(config_linter[:apt])
+        .select do |pkg|
+          if pkg.start_with?("lib")
+            true
+          else
+            add_warning "Installing the package '%s' is blocked." % pkg
```
ybiquitous, line 37: [suggestion]

- It seems more natural to use string interpolation instead of `%` method.
- Warnings can be rendered in Markdown in the Sider Web UI.

```suggestion
            add_warning "Installing the package `#{pkg}` is blocked."
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+      issues
+    end
+
+    def option_includes
+      includes = config_linter.dig(:'compilation-options', :'-I') || find_paths_containing_headers
+      includes.map { |v| "-I" + v }
+    end
+
+    def find_paths_containing_headers
+      working_dir.glob(GLOB_HEADERS, File::FNM_EXTGLOB | File::FNM_CASEFOLD)
```
ybiquitous, line 100: [suggestion] `FNM_CASEFOLD` seems unavailable for `Dir.glob`:

> Makes File.fnmatch patterns case insensitive (but not Dir.glob patterns).

```suggestion
      working_dir.glob(GLOB_HEADERS, File::FNM_EXTGLOB)
```

https://ruby-doc.org/core/File/Constants.html#FNM_CASEFOLD

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
+module Runners
+  class Processor::ClangTidy < Processor
+    Schema = StrongJSON.new do
+      let :runner_config, Schema::BaseConfig.base.update_fields { |fields|
+        fields.merge!({
+          apt: enum?(string, array(string)),
+          'compilation-options': object?(
+            '-I': enum?(string, array(string))
```
ybiquitous, line 8: [question] There are no options starting `-` in the other tools, so I feel a bit strange. Is not there any other good ideas?

test/smokes/clang_tidy/expectations.rb
```diff
@@ -0,0 +1,224 @@
...
+      id: "clang-analyzer-deadcode.DeadStores",
+      message: "Value stored to 't' during its initialization is never read",
+      object: {
+        severity: "warning"
+      },
+      git_blame_info: nil,
+      links: []
+    }
+  ],
+  analyzer: { name: "Clang-Tidy", version: "10.0.1" }
```
ybiquitous, line 41: [suggestion] How about making a variable as follows to reduce diffs when bumping the version?

https://github.com/sider/runners/blob/22c3d8e3ac3972e7aa12d36780ca07970028c847/test/smokes/eslint/expectations.rb#L3

```suggestion
  analyzer: { name: "Clang-Tidy", version: default_version }
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+          if pkg.start_with?("lib")
+            true
+          else
+            add_warning "Installing the package '%s' is blocked." % pkg
+            false
+          end
+        end
+
+      unless packages.empty?
+        capture3!("sudo", "apt-get", "install", "-y", "--no-install-recommends", *packages)
```
ybiquitous, line 43: [suggestion] How about suppress the verbose log as follows?

```suggestion
        capture3!("sudo", "apt-get", "install", "-qq", "-y", "--no-install-recommends", *packages)
```

images/clang_tidy/Dockerfile.erb
```diff
@@ -0,0 +1,15 @@
+FROM sider/devon_rex_base:2.21.0
```
ybiquitous, line 1: [suggestion] Please bump the version as follows. See also #1278

```suggestion
FROM sider/devon_rex_base:2.22.0
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def analyzer_bin
+      "clang-tidy-10"
+    end
+
+    def run_analyzer(changes)
+      issues = []
+
+      changes
+        .changed_paths
```
yykamei, line 55: Can we use `changed_paths`? For example, I wonder if Clang-Tidy can analyze without any headers in the situation that users change an only single file `foo.cpp`?

lib/runners/processor/clang_tidy.rb
suetsuguk: Agreed. commit 7c766b767e2fd1ebededdf3736c1fcab4d51e585

lib/runners/processor/clang_tidy.rb
suetsuguk: Good catch! The `setup` method suits this purpose.
commit b3ac6bad627c92d2bd7a65fcc067a02025ac0898

lib/runners/processor/clang_tidy.rb
suetsuguk: Now that the `deploy_packages` method is moved to `setup`, only `run_analyzer(changes)` statement is in the `analyze(changes)` method. I have inlined `run_analyzer` in `analyze`.
commit bfe1a014b2218643ea69feca89c283fd134c4723

lib/runners/processor/clang_tidy.rb
suetsuguk: Fixed. commit bfe1a014b2218643ea69feca89c283fd134c4723

lib/runners/processor/clang_tidy.rb
suetsuguk: Good catch!
commit 52da9302cbd109e2164b041a70901472cd7eb28b

lib/runners/processor/clang_tidy.rb
suetsuguk: Good catch!
commit 3aad06d90bc7ae0aa68981300a7ec00bd49a8047

test/smokes/clang_tidy/expectations.rb
suetsuguk: Fixed. commit 334b6dbbb8f51c0c4d9500e853e88c617cbc7649

lib/runners/processor/clang_tidy.rb
suetsuguk: Good catch!
commit 44b237bda22e18255686d5cf0abd5f184386548c

images/clang_tidy/Dockerfile.erb
suetsuguk: Fixed. commit 617ce2886a8e07b498301eba8bbf0af051232d7d

lib/runners/processor/clang_tidy.rb
suetsuguk: The `VALID_EXTENSIONS` array already covers all C/C++ files. But I missed some header files in `GLOB_HEADERS`.
C compilers can deal with the header files in either format, so that there are many extensions.
I will add header extensions except ".re", ".cats" and ".idc". They are extremely rare case.

ybiquitous: OK, I understand it. 👍

lib/runners/processor/clang_tidy.rb
suetsuguk: I thought that **all** files are copied in the temporary folder where analyzers run.
If it is true, Clang-Tidy is able to find necessary header files from the folder and its sub-folders.
Our `clang_tidy.rb` finds only changed C/C++ files in `changed_paths` and passes to the `clang-tidy` command so that we can reduce analysis time.

yykamei: Sorry, I misunderstood. Files themselves remain.

By the way, Clang-Tidy does not receive multiple file paths?

lib/runners/processor/clang_tidy.rb
suetsuguk: The clang-tidy's `-I` options should be relative from the working path where the `clang-tidy` command has been executed.

The `working_dir` variable is an instance of `Pathname` class which accepts `File::FNM_CASEFOLD` as bellow.
Indeed, the smoke test named `option_I_default` covers a test for the capital letters.

```shell
kentaro@debian103amd64:~/depot/runners/test/smokes/clang_tidy/success$ tree
.
+-- foo
|   +-- テスト.cpp
+-- myinclude
|   +-- テスト.hpp
+-- test.c

2 directories, 3 files
kentaro@debian103amd64:~/depot/runners/test/smokes/clang_tidy/success$ ruby -rpathname -e 'p Pathname(".").glob("**/*.HpP", File::FNM_EXTGLOB | File::FNM_CASEFOLD)'
[#<Pathname:myinclude/テスト.hpp>]
kentaro@debian103amd64:~/depot/runners/test/smokes/clang_tidy/success$ ruby -rpathname -e 'p Pathname(".").glob("**/*.HpP", File::FNM_EXTGLOB)'
[]
kentaro@debian103amd64:~/depot/runners/test/smokes/clang_tidy/success$
```

suetsuguk: Could it be that the `current_dir` is where the shell command is executed?
I'm not sure exact difference between `current_dir` and `working_dir`.

lib/runners/processor/clang_tidy.rb
suetsuguk: Yes, Clang-Tidy accepts multiple target files. But I execute one-by-one to avoid the limit of shell command-line length.
The error will occur only when tons of files are passed, though.

lib/runners/processor/clang_tidy.rb
ybiquitous: Using `File::FNM_CASEFOLD` seems good. 👍 

---

> I'm not sure exact difference between `current_dir` and `working_dir`.

`current_dir` is a location which the `root_dir` option points at:

https://github.com/sider/runners/blob/87985b85b13fec23274aa719c2a3c5231dd9e10b/lib/runners/harness.rb#L47

https://github.com/sider/runners/blob/87985b85b13fec23274aa719c2a3c5231dd9e10b/lib/runners/processor.rb#L126-L132

https://github.com/sider/runners/blob/87985b85b13fec23274aa719c2a3c5231dd9e10b/lib/runners/shell.rb#L50

On the other hand, `working_dir` is the root of a repository.

lib/runners/processor/clang_tidy.rb
suetsuguk: I searched manpages of gcc, clang and cl (Microsoft C compiler) for the command line usage, but I could not find a long name of the `-I` option. And what is worse, clang provides the `-include` option for other purpose.

`include-search-paths:` is a possible solution. But most C/C++ programmers are familiar with compilation options of `-I`, `-W`, `-D`, etc.

ybiquitous: I understand `-I` is familiar, but I think we should make the option name readable for unfamiliar users. And, I'm worried a too short name may occur typo potentially.

Also, this option can be used by other tools (e.g. Cppcheck; see #1280).

I think `linter.clang_tidy.compilation-options.include-search-paths` too long, so what about just `linter.clang_tidy.include-path`?

```yaml
linter:
  clang_tidy:
    include-path: [my/lib/headers, /usr/include]
  
  cppcheck:
    include-path: my/lib/headers
```

lib/runners/processor/clang_tidy.rb
yykamei: OK, that's understandable. It seems good as long as it doesn't give rise to performance issues by invoking subprocesses one by one.

lib/runners/processor/clang_tidy.rb
suetsuguk: Added header extensions to `GLOB_HEADERS` and made smoke tests accordingly.
commit 2d508e0096f3e6f45757d85c8749b4d5a8efce2f

lib/runners/processor/clang_tidy.rb
suetsuguk: Thank you for making it clear. `current_dir` is suitable.
commit 027d4c5e0ba8c1f520f17c3a67af4b825078e6ef

lib/runners/processor/clang_tidy.rb
ybiquitous: Thanks. 👍

lib/runners/processor/clang_tidy.rb
suetsuguk: Agreed. Now that the short name `-I` is dropped, `compilation-options` is also meaningless and dropped.
commit 626a8bb686d3ba75dac716f53317227f1556debd and 277d1f1ae3db144f1c9944e2692ff47394afc7a0

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def option_include_path
+      includes = config_linter[:'include-path'] || find_paths_containing_headers
+      includes.map { |v| "-I" + v }
+    end
+
+    def find_paths_containing_headers
+      current_dir.glob(GLOB_HEADERS, File::FNM_EXTGLOB | File::FNM_CASEFOLD)
+        .select { |path| path.file? }
+        .map { |path| relative_path(path.parent, from: current_dir).to_path }
```
ybiquitous, line 103: [suggestion] Maybe this refactoring may be better (but I don't try it locally...):

```suggestion
      Pathname.glob(GLOB_HEADERS, File::FNM_EXTGLOB | File::FNM_CASEFOLD)
        .filter_map { |path| path.parent.to_path if path.file? }
```

See https://ruby-doc.org/core/Enumerable.html#filter_map-method

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    def construct_result(stdout)
+      issues = []
+
+      stdout.each_line do |line|
+        # issue format
+        # <path>:<line>:<column>: <severity>: <message> [<id>]
+        match = line.strip.match(/^(?<path>.+):(?<line>\d+):(?<column>\d+): (?<severity>[^:]+): (?<message>.+) \[(?<id>[^\[]+)\]$/)
+        if match
```
ybiquitous, line 78: [suggestion] Maybe we can refactor it via `String#scan`:

```ruby
pattern = /^(.+):(\d+):(\d+): ([^:]+): (.+) \[([^\[]+)\]$/
stdout.scan(pattern) do |path, line, column, severity, message, id|
  # Issue.new(...)
end
```

https://ruby-doc.org/core/String.html#scan-method

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,107 @@
...
+      changes
+        .changed_paths
+        .select { |path| VALID_EXTENSIONS.include?(path.extname.downcase) }
+        .map{ |path| relative_path(working_dir / path, from: current_dir) }
+        .reject { |path| path.to_s.start_with?("../") } # reject files outside the current_dir
+        .each do |path|
+          stdout, = capture3!(analyzer_bin, path.to_s, "--", *option_include_path,
+            is_success: ->(status) { [0, 1].include?(status.exitstatus) })
+          ret = construct_result(stdout)
+          issues.push(*ret)
```
ybiquitous, line 44: [suggestion] If `construct_result` can receive a block, maybe we can prevent from extra array:

E.g.

```ruby
construct_result(stdout).each { |issue| issues << issue }

def construct_result(stdout)
  stdout.each_line do
    yield Issue.new(...)
  end
end
```

lib/runners/processor/clang_tidy.rb
ybiquitous: Thanks. 😃

lib/runners/processor/clang_tidy.rb
suetsuguk: LGTM. The `match` variable is omitted.
commit 2e41674cb5b9861c351ad650b78c1cb9a7f09497

lib/runners/processor/clang_tidy.rb
suetsuguk: LGTM. The `issues` variable is omitted.
commit 1b9e2416bfc1a217ab966e326342bf53f511f587

lib/runners/processor/clang_tidy.rb
suetsuguk: LGTM. The `.select` and `.map` are converted to `.filter_map`.
commit e62a5d4236988b1474601bd97eb1d080d732659c

The `relative_path` method is left unchanged. Users need to set relative paths to `include-path:'. We should set the same format as users so that we can test the clang-tidy's `-I` option. Also users may get confused about absolute paths on the Sider console log.

ybiquitous: Maybe `Pathname.glob(GLOB_HEADERS)` should return relative paths, so it seems needless to use `relative_path`. 🤔 Right?

```console
$ irb
irb(main):001:0> Pathname.glob("*.md")
=> [#<Pathname:CHANGELOG.md>, #<Pathname:README.md>, #<Pathname:CONTRIBUTING.md>]
```

suetsuguk:
> Here are some thoughts on how to choose packages to deploy: Search for packages named "lib*-dev" in the result of Debian Popularity Contest [https://popcon.debian.org/]. Select them from the top until the sum of file size reaches about 300MB. They might be 20 to 30 packages.

```shell
> curl -fsSL https://popcon.debian.org/by_inst | tr -s ' ' | cut -d' ' -f 2 | grep "lib.*-dev" | head -200 > packages.list
> apt-get install `cat packages.list | head -30`
```

Disk space will be used.
head -30 -> 188 MB
head -35 -> 418 MB
head -40 -> 421 MB
head -50 -> 423 MB
head -70 -> 433 MB
head -100 -> 533 MB
head -120 -> 954 MB
head -150 -> 1,324 MB

I removed bellow packages which cause an error "E: Unable to locate package" on installing.
libgcc-6-dev
libstdc++-6-dev
libgcc-4.9-dev
libstdc++-4.9-dev
libcrypt-dev
libgcc-9-dev
libgcc-4.8-dev
libstdc++-9-dev
libpython3.5-dev
libpng12-dev

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    def analyze(changes)
+      issues = []
+
+      changes
+        .changed_paths
+        .select { |path| VALID_EXTENSIONS.include?(path.extname.downcase) }
+        .map{ |path| relative_path(working_dir / path, from: current_dir) }
+        .reject { |path| path.to_s.start_with?("../") } # reject files outside the current_dir
```
ybiquitous, line 39: [suggestion] Maybe the following code seems more safe and clear to me:

```suggestion
        .reject { |path| path.parent.to_path == "..") } # reject files outside the current_dir
```

lib/runners/processor/clang_tidy.rb
suetsuguk: We have an instance of `Pathname` which points `/tmp/blablabla`.

```shell
/tmp$ mkdir -p aaa/bbb/ccc/
/tmp$ touch aaa/bbb/ccc/test.md
/tmp$ cd
~$ irb
irb(main):001:0> path = Pathname("/tmp/aaa")
irb(main):002:0> path.glob("*.md")
=> []
irb(main):003:0> path.glob("**/*.md")
=> [#<Pathname:/tmp/aaa/bbb/ccc/test.md>]
irb(main):004:0>
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,99 @@
...
+    def analyzer_bin
+      "clang-tidy"
+    end
+
+    def analyze(changes)
+      issues = []
+
+      changes
+        .changed_paths
+        .select { |path| VALID_EXTENSIONS.include?(path.extname.downcase) }
```
ybiquitous, line 37: [suggestion] We can use also glob like GLOB_HEADERS via `fnmatch?`:

```ruby
GLOB_SOURCES = "*.{c,cpp,c++,cc,cp,cxx}".freeze
```

```suggestion
        .select { |path| path.fnmatch?(GLOB_SOURCES, File::FNM_EXTGLOB | File::FNM_CASEFOLD) }
```

See also:

- https://ruby-doc.org/stdlib/libdoc/pathname/rdoc/Pathname.html#fnmatch-3F-method
- https://ruby-doc.org/core/File.html#fnmatch-method

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,99 @@
...
+    private
+
+    def deploy_packages
+      # select "lib*" and report others as warning for security concerns
+      packages = Array(config_linter[:apt])
+        .select do |pkg|
+          if pkg.start_with?("lib")
+            true
+          else
+            add_warning "Installing the package `#{pkg}` is blocked."
```
ybiquitous, line 60: [suggestion] The warning is due to `sider.yml`, so can you add `file`, please?

```suggestion
            add_warning "Installing the package `#{pkg}` is blocked.", file: config.path_name
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,99 @@
...
+          object: {
+            severity: severity,
+          },
+          schema: Schema.issue,
+        )
+      end
+    end
+
+    def option_include_path
+      includes = config_linter[:'include-path'] || find_paths_containing_headers
```
ybiquitous, line 89: [suggestion] When `config_linter[:'include-path']` is a string, it may fail. How about using `Array()`?

```suggestion
      includes = Array(config_linter[:'include-path'] || find_paths_containing_headers)
```

lib/runners/processor/clang_tidy.rb
suetsuguk: Ohhh, I missed `Pathname.glob()` in your suggestion.

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,99 @@
...
+    def analyze(changes)
+      issues = []
+
+      changes
+        .changed_paths
+        .select { |path| VALID_EXTENSIONS.include?(path.extname.downcase) }
+        .map{ |path| relative_path(working_dir / path, from: current_dir) }
+        .reject { |path| path.to_s.start_with?("../") } # reject files outside the current_dir
+        .each do |path|
+          stdout, = capture3!(analyzer_bin, path.to_s, "--", *option_include_path,
```
ybiquitous, line 41: [suggestion] For example, how about linting multiple files together?

```suggestion
        .each_slice(100) do |paths| # Avoid too many arguments 
          stdout, = capture3!(analyzer_bin, *paths.to_map(&:to_path), "--", *option_include_path,
```

https://ruby-doc.org/core/Enumerable.html#each_slice-method

yykamei:
https://github.com/sider/runners/pull/1254#issuecomment-656428095

The algorithm of choosing popular packages is good :+1: I think lines with "Not in sid" could be excluded because these packages seem not to be maintained.

suetsuguk:
I take a quick look at the top 200 packages and find they are mostly **system** libraries.
In addition, we have to consider some **user** libraries listed below.
https://github.com/fffaraz/awesome-cpp
https://en.cppreference.com/w/cpp/links/libs

lib/runners/processor/clang_tidy.rb
suetsuguk: I'm stuck. Results of `capture3!()` with multiple files are different from results of `clang-tidy` on `bash`.

(1) Three issues found on `bash`.
```shell
$ clang-tidy baz/test.cp foo/テスト.C++ test.CC -- -I bar/myinc
...
/home/kentaro/depot/runners/test/smokes/clang_tidy/option_root_dir/src/baz/test.cp:5:15: warning: Value stored to 't' during its initialization is never read [clang-analyzer-deadcode.DeadStores]
...
/home/kentaro/depot/runners/test/smokes/clang_tidy/option_root_dir/src/foo/テスト.C++:6:5: warning: 2nd function call argument is an uninitialized value [clang-analyzer-core.CallAndMessage]
...
/home/kentaro/depot/runners/test/smokes/clang_tidy/option_root_dir/src/test.CC:5:3: warning: Assigned value is garbage or undefined [clang-analyzer-core.uninitialized.Assign]
...
```
When I loop over the files and call `capture3!()` with a single file, I get the same result.

(2) Only two issues reported from `clang_tidy.rb`.
"/tmp/d20200710-6-zwq123/src/foo/テスト.C++:6:5: warning: 2nd function call argument is an uninitialized value [clang-analyzer-core.CallAndMessage]\n
...
/tmp/d20200710-6-zwq123/src/test.CC:5:3: warning: Assigned value is garbage or undefined [clang-analyzer-core.uninitialized.Assign]\n
..."
With literal arguments `stdout, = capture3!(analyzer_bin, "baz/test.cp", "foo/テスト.C++", "test.CC", "--", *option_include_path,`, I got the same results.

I cannot find the cause of this problem so far. I will give up your suggestion.

ybiquitous: OK. My suggestion is nice to have. Not must. 👍

lib/runners/processor/clang_tidy.rb
suetsuguk: Let's take an example below:

```
/tmp/blablabla/WORKING_DIR/CURRENT_DIR/foo/bar.c
                          /aaa/bbb/ccc.c
```

We have `../aaa/bbb/ccc.c` (and `foo/bar.c`). Apply it to the `path.parent.to_path`, then we get "../aaa/bbb" which is not equal to "..".
Alternatively, we can use `path.parent.to_path.split(File::SEPARATOR)[0] == ".."`, but it is a bit complicated.

lib/runners/processor/clang_tidy.rb
suetsuguk: I missed a test case for that case. Thank you for pointing out.
commit 0cc349ab4eed6ee378fec1884182560611d947e3

lib/runners/processor/clang_tidy.rb
suetsuguk: LGTM.
commit c142082debb6e8e4ff1a0143c94519647018c232

lib/runners/processor/clang_tidy.rb
suetsuguk: Fixed. commit c142082debb6e8e4ff1a0143c94519647018c232

lib/runners/processor/clang_tidy.rb
suetsuguk: LGTM.
commit f0fa9e6c31248e842659e94fad41774d04d63a50

lib/runners/processor/clang_tidy.rb
ybiquitous: Make sense. 👍

suetsuguk:
I have added about 60 packages to `packages.list` as our built-in packages. commit 2dab6361bb61e1e2db2481f446171132a1e5b67b
- pick up the top 30 of Debian popcon; `curl -fsSL https://popcon.debian.org/by_inst | grep -v "Not in sid" | tr -s ' ' | cut -d' ' -f 2 | grep "lib.*-dev" | head -30`
- pick up several packages with a large number of stars for each section in https://github.com/fffaraz/awesome-cpp

images/clang_tidy/Dockerfile.erb
```diff
@@ -0,0 +1,16 @@
...
+COPY --chown=<%= chown %> images/clang_tidy/packages.list ${RUNNER_USER_HOME}/
+
+USER root
+RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
+    add-apt-repository "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-10 main" && \
+    apt-get update && \
+    xargs apt-get install -qq -y --no-install-recommends < ${RUNNER_USER_HOME}/packages.list && \
+    apt-get clean && \
+    update-alternatives --install /usr/local/bin/clang-tidy clang-tidy /usr/lib/llvm-10/bin/clang-tidy 20 && \
+    echo "analyzer_runner ALL=NOPASSWD: /usr/bin/apt-get" >> /etc/sudoers
```
ybiquitous, line 12: [suggestion] How about the following suggestions?

- Replace `analyzer_runner` with `${RUNNER_USER}`.
- Create a new file into `/etc/sudoers.d/` instead of modifying the `/etc/sudoers` file.

```suggestion
    echo "${RUNNER_USER} ALL=NOPASSWD: /usr/bin/apt-get" > /etc/sudoers.d/runner
```

suetsuguk: Good catch!
Fixed. commit 5ad9e819efea4b4dd7f5e1204d8709a87aabf4fd

images/clang_tidy/Dockerfile.erb
```diff
@@ -0,0 +1,16 @@
+FROM sider/devon_rex_base:2.22.0
+
+COPY --chown=<%= chown %> images/clang_tidy/packages.list ${RUNNER_USER_HOME}/
+
+USER root
+RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
+    add-apt-repository "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-10 main" && \
+    apt-get update && \
+    xargs apt-get install -qq -y --no-install-recommends < ${RUNNER_USER_HOME}/packages.list && \
+    apt-get clean && \
```
ybiquitous, line 10: [question] Does removing `/var/lib/apt/lists/*` make sense?

```suggestion
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* && \
```

E.g. https://github.com/sider/runners/blob/840af184a3e5b95e82661a8f5748a24e630ca5e1/images/cppcheck/Dockerfile.erb#L16

images/clang_tidy/Dockerfile.erb
```diff
@@ -0,0 +1,16 @@
+FROM sider/devon_rex_base:2.22.0
+
+COPY --chown=<%= chown %> images/clang_tidy/packages.list ${RUNNER_USER_HOME}/
+
+USER root
+RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
+    add-apt-repository "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-10 main" && \
+    apt-get update && \
+    xargs apt-get install -qq -y --no-install-recommends < ${RUNNER_USER_HOME}/packages.list && \
```
ybiquitous, line 9: [suggestion] I think it important to left comments about how we decided the installed packages (see https://github.com/sider/runners/pull/1254#issuecomment-657333085).

So, how about using an ERB file instead of just a plain-text file as follows?

The `clang-tidy` package is written directly to distinguish the main package and the extra packages.

```diff
diff --git a/images/clang_tidy/Dockerfile.erb b/images/clang_tidy/Dockerfile.erb
index f80efa39..ed6e1bdb 100644
--- a/images/clang_tidy/Dockerfile.erb
+++ b/images/clang_tidy/Dockerfile.erb
@@ -1,12 +1,13 @@
 FROM sider/devon_rex_base:2.22.0
 
-COPY --chown=<%= chown %> images/clang_tidy/packages.list ${RUNNER_USER_HOME}/
-
 USER root
 RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
     add-apt-repository "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-10 main" && \
     apt-get update && \
-    xargs apt-get install -qq -y --no-install-recommends < ${RUNNER_USER_HOME}/packages.list && \
+    apt-get install -qq -y --no-install-recommends clang-tidy-10=1:10.0.1~++20200708124224+ef32c611aa2-1~exp1~20200707224822.188 && \
+    apt-get install -qq -y --no-install-recommends \
+<%= render_erb('images/clang_tidy/packages.erb').lines(chomp: true).reject(&:empty?).map{ "      #{_1} \\\n" }.join %>
+      && \
     apt-get clean && \
     update-alternatives --install /usr/local/bin/clang-tidy clang-tidy /usr/lib/llvm-10/bin/clang-tidy 20 && \
     echo "${RUNNER_USER} ALL=NOPASSWD: /usr/bin/apt-get" > /etc/sudoers.d/runner
```

Replacing `images/clang_tidy/packages.list` with `images/clang_tidy/packages.erb`:

```erb
<%# Popular packages of Debian popcon %>
<%# `curl -fsSL https://popcon.debian.org/by_inst | grep -v "Not in sid" | tr -s ' ' | cut -d' ' -f 2 | grep "lib.*-dev" | head -30` %>
libapr1-dev
libasio-dev
...


<%# Popular packages of https://github.com/fffaraz/awesome-cpp %>
libc6-dev
linux-libc-dev
...
```

images/clang_tidy/Dockerfile.erb
suetsuguk: If we delete files in `/var/lib/apt/lists/`, we will fail to execute `apt-get install` at `deploy_packages` in `clang-tidy.rb`.
Otherwise, we must execute `apt-get update` before `apt-get install` at `deploy_packages`.

ybiquitous: Make sense. 👍

images/clang_tidy/Dockerfile.erb
suetsuguk: Applying `grep` to the plain-text file is more simple. We can leave the comment lines starting with "#".
What do you think?

```shell
...
apt-get update && \
grep -v "^\s*#" ${RUNNER_USER_HOME}/packages.list | xargs apt-get install -qq -y --no-install-recommends && \
apt-get clean && \
...
```

```text
# Main Clang-Tidy package
clang-tidy-10=1:10.0.1~++20200708124224+ef32c611aa2-1~exp1~20200707224822.188

# Popular packages of https://github.com/fffaraz/awesome-cpp
libapr1-dev
libasio-dev
...

# Popular packages of Debian popcon
# `curl -fsSL https://popcon.debian.org/by_inst | grep -v "Not in sid" | tr -s ' ' | cut -d' ' -f 2 | grep "lib.*-dev" | head -30`
libc6-dev
linux-libc-dev
...
```

ybiquitous: Looks good! 👍

suetsuguk: Fixed. commit 3fe5ab7e3a07084c62549a22f338108333d4f1f3

ybiquitous: Thank you!

lib/runners/processor/clang_tidy.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    def deploy_packages
+      # select "lib*" and report others as warning for security concerns
+      packages = Array(config_linter[:apt])
+        .select do |pkg|
+          if pkg.start_with?("lib")
+            true
+          else
+            add_warning "Installing the package `#{pkg}` is blocked.", file: config.path_name
+            false
```
ybiquitous, line 61: [question] Are there other cases like `zlib1g-dev` (not starting with `lib`)?

suetsuguk: We might have misunderstood. There are development packages that do not start with `lib`.
Indeed, the `zlib1g-dev` package contains `/usr/include/zlib.h`.
We must select the packages with "-dev$" suffix in names.

```shell
$ curl -fsSL https://popcon.debian.org/by_inst | grep -v "Not in sid" | tr -s ' ' | cut -d' ' -f 2 > packages.all.list
$ cat packages.all.list | wc -l
65285
$ grep -e "-dev$" packages.all.list | wc -l
9421
$ grep -e "-dev$" packages.all.list | grep -e "^lib" | wc -l
7376
$ grep -e "-dev$" packages.all.list | grep -v -e "^lib" | wc -l
2045
```

ybiquitous: Make sense. Please update the logic. 👍

ybiquitous:
@suetsuguk Can I merge this PR?

suetsuguk:
@ybiquitous  Yes, please. Thank you for taking time.

PR #1257, opened 2020-07-02
Allow `location(start_line,start_column)` format

ybiquitous:

Many tools report an issue with not only a line but also a column.
This newly added column will be useful in the future.

None.

> Check the following items.


ybiquitous:
@yykamei What do you think about this change? Can we allow this new form?

yykamei:
> What do you think about this change? Can we allow this new form?

Yes, I think so. It's reasonable to add the new schema. I think it's safe to add it.

ybiquitous:
Thanks for your review!

PR #1264, opened 2020-07-03
[ESLint] Add `target` option instead of `dir` option

### LINKED ISSUE #891: [ESLint] `root_dir` and `dir` options are confusing
`root_dir` and `dir` are confusing and I think users can easily misconfigure `sider.yml`.

E.g.

```yaml
linter:
  eslint:
    root_dir: frontend/
    dir: frontend/src/
```

So, I suggest:

- Mark `linter.eslint.dir` option as **deprecated**
- Add `linter.eslint.target` option

See also:

https://github.com/sider/runners/blob/1d1e1253e177b860ffdeb62d1573de6e835a2474/lib/runners/processor/eslint.rb#L8

ybiquitous:

The `dir` option can actually allow files, so the naming is confusing (of course, the `root_dir` option).
This change adds a new option `target` (used by other tools also) and deprecates the `dir` option.

Fix #891

> Check the following items.


ybiquitous:
Oh, good catch! I forget it. 😅

ybiquitous:
The next version of Runners will include the ESLint update, so I will write them together.

https://github.com/sider/runners/blob/4ff91a5a59f01086ace90efe71a769013cfe6746/CHANGELOG.md

Thanks for your reminder! 🥰

PR #1265, opened 2020-07-03
[HAML-Lint] Add `target` option

### LINKED ISSUE #1065: [HAML-Lint] Rename `file` option
This `file` option actually can receive a directory.

https://github.com/sider/runners/blob/7b94afb3dcc6ebd43104f73d46765709368a9fa3/lib/runners/processor/haml_lint.rb#L10

ybiquitous:

The current `file` option allows directories actually, so the naming is confusing.
The newly added `target` option has a more clear name and supports an array format. For example:

```yaml
linter:
  haml_lint:
    target:
      - src/
      - lib/test.*
```

Fix #1065

> Check the following items.


lib/runners/processor/haml_lint.rb
```diff
@@ -57,6 +61,12 @@ def default_gem_specs
     def setup
       add_warning_if_deprecated_options([:options])
 
+      if config_linter[:file]
+        add_warning <<~MSG, file: config.path_name
+          The `file` option is deprecated. Use the `target` option instead.
+        MSG
+      end
+
```
yykamei, line 69: Isn't sufficient to add an argument to `add_warning_if_deprecated_options`?

```suggestion
      add_warning_if_deprecated_options([:options, :file])
```

lib/runners/processor/haml_lint.rb
```diff
@@ -66,48 +76,42 @@ def setup
...
     def setup_default_config
-      # NOTE: We expect default_rubocop.yml to be located in $HOME.
-      default_config = (Pathname(Dir.home) / './default_rubocop.yml').realpath
-      path = Pathname("#{current_dir}/.rubocop.yml")
-      return false if path.exist?
-      path.parent.mkpath
-      FileUtils.cp(default_config, path)
-      true
+      config_file = ".rubocop.yml"
+      return if File.exist? config_file
```
yykamei, line 86: This change is excellent :+1: In addition to this, I think the presence of `haml_lint_config` should be checked before copying the default config in case the custom RuboCop configuration requires an older version of Meowcop.

```suggestion
      return if File.exist? config_file
      return unless haml_lint_config.empty?
```

lib/runners/processor/haml_lint.rb
ybiquitous: First, I thought about it. But `add_warning_if_deprecated_options` does not suggest an alternative option, so I thought it's not kindful for users.

Perhaps, we might make `add_warning_if_deprecated_options` customizable, but I think it's out of the scope of this PR. What do you think?

yykamei: Reasonable. Thank you.

lib/runners/processor/haml_lint.rb
ybiquitous: Good catch! 👍

ybiquitous: Umm..., when I think again, this `setup_default_config` method is for `.rubocop.yml`, not for [`.haml-lint.yml`](https://github.com/sds/haml-lint#configuration).

So, it seems to me that we should just rename `setup_default_config` to `setup_rubocop_default_config` and keep this method's logic.
What do you think?

yykamei: OK, I misunderstood the configuration is for RuboCop, but it is for HAML-Lint. Please just rename it.

ybiquitous: Thanks for the feedback! I've fixed it via fcdb1b62.

PR #1270, opened 2020-07-06
[ESLint] Update the default configuration

ybiquitous:

This change enables the [`no-unused-vars`](https://eslint.org/docs/rules/no-unused-vars) rule in the default configuration. Because:

- it is included in the `eslint:recommended` ruleset.
- it is useful in most cases, especially in refactoring.
- it is designed not to produce false positives as possible by default.

Also, this change sets `root: true` to make the ESLint behavior clear.

> in successive parent directories all the way up to the root directory of the filesystem (unless `root: true` is specified)

https://eslint.org/docs/user-guide/configuring#using-configuration-files-1

None.

> Check the following items.


ybiquitous:
Thanks!

Issue #900, opened 2020-03-27
Add pre-installed plugins and configs


ybiquitous:
For pluggable tools such as ESLint or stylelint, there are often used plugins and sharable configurations, e.g. `eslint-plugin-react` or `stylelint-config-standard`.

I think that to pre-install such plugins and configs is useful for many users.
Because users don't need to manage and maintain such plugins and configs (almost npm packages) by themselves.

For example, users can just write own `.eslintrc.yml` as follows,

```yaml
extends:
  - eslint:recommended
  - plugin:react/recommended
```

then Sider can analyze well their code with the `.eslitrc.yml` configuration!
(but now, Sider fails with "not installed `eslint-plugin-react`" 😓 )

Perhaps, to realize the feature, we can just add the famous or often used packages into each `package.json`:

https://github.com/sider/runners/blob/214886b5dc1cc0828b2ff972fa8aa8ed1ae65a89/images/eslint/package.json#L2-L4

https://github.com/sider/runners/blob/214886b5dc1cc0828b2ff972fa8aa8ed1ae65a89/images/stylelint/package.json#L2-L5

ybiquitous:
ESLint and stylelint were done.

PR #1271, opened 2020-07-06
[ESLint] Pre-install popular configs and plugins

ybiquitous:

This change aims to make Sider available with **zero configurations** of `sider.yml` and `package.json`.
It is enough for users to prepare their `.eslintrc.*` files to succeed in Sider analysis.

In other words, users will be free from their dependency management for using ESLint configurations and plugins.

About this **popularity**, I decided which packages should be included in the default list by the following npm search.

- https://www.npmjs.com/search?q=eslint-config&ranking=popularity
- https://www.npmjs.com/search?q=eslint-plugin&ranking=popularity

Also, I removed some packages with low-frequency updates from the list even if the packages are popular.

Note that [eslint-plugin-vue](https://www.npmjs.com/package/eslint-plugin-vue) is still alpha requiring ESLint v7, so I did **not** include it.
I think we should include the popular package for Vue users when it will be published with alpha.

A part of #900

> Check the following items.


ybiquitous:
Thank you, let's try!

PR #1277, opened 2020-07-07
Remove codeql-analysis.yml

yykamei:

Our main code is written in Ruby, but it is not supported yet.
Supported languages are described [here](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning).

None

> Check the following items.

    - I think it's not needed.


yykamei:
I decided to remove codeql-analysis.yml because CodeQL Analysis has not yet supported Ruby.

Issue #1233, opened 2020-06-26
git-fetch cause "429 Too Many Requests"


ybiquitous:
The following error sometimes occurs:

> git-fetch failed: fatal: unable to access 'https://[FILTERED]@github.com/org/repo/': The requested URL returned error: 429 

The code location:

https://github.com/sider/runners/blob/224ad45bfe07974ee85037c272413fd088306052/lib/runners/workspace/git.rb#L33

ybiquitous:
If the error is raised due to a rate limit over, we may need to return a new error.
The Runners caller can retry the execution of Runners when it receives the new error.

yykamei:
* eightcard: There are [17](https://app.bugsnag.com/sleeek-corporation/runners/errors/5ec4f901d42bbb001725abad?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=429%20&filters[search][1][type]=eq&filters[search][1][value]=eightcard) errors within 2 weeks
* crowdworksjp: There [21](https://app.bugsnag.com/sleeek-corporation/runners/errors/5ec4f901d42bbb001725abad?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=429&filters[search][1][type]=eq&filters[search][1][value]=crowdworksjp) errors within 2 weeks
* padlet: There [40](https://app.bugsnag.com/sleeek-corporation/runners/errors/5ec4f901d42bbb001725abad?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=429&filters[search][1][type]=eq&filters[search][1][value]=mozart) errors within 2 weeks

The correlation between the number of repositories and the error rate does not look sufficient to explain the error :thinking:

ybiquitous:
@yykamei How about increasing retry count and interval?

yykamei:
Yes, the only thing we can is just retrying for now. I will implement it.

ybiquitous:
@yykamei Can we close this issue?

yykamei:
Yes, I closed.

PR #1279, opened 2020-07-08
Increase :tries and :sleep of Retryable for git-fetch(1)

yykamei:
git-fetch(1) sometimes fails because of Rate-Limit.  Ideally, we should see the HTTP header `X-RateLimit-Reset` from GitHub, but I just made Runners try fetching again more.

### Related issues

#1233 

### Checklist

### DIFF
--- a/lib/runners/shell.rb
+++ b/lib/runners/shell.rb
@@ -80,11 +80,11 @@ module Runners
       [stdout, stderr]
     end
 
-    def capture3_with_retry!(command, *args, tries: 3)
+    def capture3_with_retry!(command, *args, tries: 3, sleep: -> (n) { n + 1 })
       Retryable.retryable(
         tries: tries,
         on: ExecError,
-        sleep: -> (n) { n + 1 },
+        sleep: sleep,
         exception_cb: -> (_ex) { trace_writer.message "Command failed. Retrying..." }
       ) do
         capture3!(command, *args)
--- a/sig/runners.rbi
+++ b/sig/runners.rbi
@@ -127,7 +127,7 @@ class Runners::Processor
   def self.register_config_schema: (**any) -> void
   def capture3: (String, *String, **capture3_options) -> [String, String, Process::Status]
   def capture3!: (String, *String, **capture3_options) -> [String, String]
-  def capture3_with_retry!: (String, *String, ?tries: Integer) -> [String, String]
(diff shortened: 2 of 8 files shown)


sig/runners.rbi
```diff
@@ -174,7 +174,7 @@ class Runners::Shell
 
   def capture3: (String, *String, **capture3_options) -> [String, String, Process::Status]
   def capture3!: (String, *String, **capture3_options) -> [String, String]
-  def capture3_with_retry!: (String, *String, ?tries: Integer) -> [String, String]
+  def capture3_with_retry!: (String, *String, ?tries: Integer, ?sleep: ^(Integer) -> Numeric) -> [String, String]
```
yykamei, line 177: I couldn't write the signature like `^(Integer) -> Integer | Float` because of a parse error.

lib/runners/shell.rb
```diff
@@ -80,11 +80,11 @@ def capture3!(command, *args, **options)
       [stdout, stderr]
     end
 
-    def capture3_with_retry!(command, *args, tries: 3)
+    def capture3_with_retry!(command, *args, tries: 3, sleep: -> (n) { n + 1 })
       Retryable.retryable(
         tries: tries,
         on: ExecError,
-        sleep: -> (n) { n + 1 },
+        sleep: sleep,
```
ybiquitous, line 87: [question] Are there no problems when changing the default from `sleep: 1` to `sleep: ->(n) { n + 1 }`?

https://github.com/nfedyashev/retryable/blob/14fc479054546dfc507920b1fa357d6476f150ca/lib/retryable/configuration.rb#L29

sig/runners.rbi
ybiquitous: Maybe the following works:

```diff
-?sleep: ^(Integer) -> Numeric)
+?sleep: ^(Integer) -> (Integer | Float)
```

But I think `Numeric` is enough 👍

yykamei: Thanks

lib/runners/shell.rb
yykamei: No, I think there are no problems to change the value from the default one. I don't want to change the value. If we should consider it, I want to deal with it on another PR.

ybiquitous: 🙆

yykamei:
Thank you!

PR #1290, opened 2020-07-10
[ESLint] Set `ecmaVersion:2015` to the default configuration

ybiquitous:

This change aims to reduce potential parse errors in users' JavaScript code.

Our current ESLint default configuration has used `ecmaVersion: 5` (from the ESLint default).
But I think the default setting has caused many parse errors.
Because most JavaScript developers use the modern syntax in 2020 now.

According to the State of JavaScript 2019, **97%** of developers use arrow functions.
Also, almost modern browsers support the modern syntax natively,
and transpilers like Babel or TypeScript are widely used for old browsers.

See also:

- https://eslint.org/docs/user-guide/configuring#specifying-parser-options
- https://2019.stateofjs.com/features/syntax/
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility

None.

> Check the following items.


test/smokes/eslint/no_config/index.js
```diff
@@ -1,2 +1,5 @@
 for (var i = 0; i < 10; i--) {
 }
+
+let a = () => ""
```
ybiquitous, line 4: [note] Unless `ecmaVersion: 2015`, ESLint reports the following error:

```
Parsing error: Unexpected token a
```

ybiquitous:
Thank you! 😄

PR #1301, opened 2020-07-13
Prevent unexpected error when missing issue ID

ybiquitous:

This change aims to prevent an unexpected error when an issue ID is missing.
Some processors for tools like ESLint have addressed missing IDs, but some never do it.
Today we discovered missing IDs also when running remark-lint.

Thus, this change addresses missing IDs for all processors.

None.

> Check the following items.


ybiquitous:
Thanks. 👍

Issue #1121, opened 2020-05-22
Remove `missing_files` from result


ybiquitous:
This `missing_files` type is needless actually because Sider considers it as the same as `failure`.

https://github.com/sider/runners/blob/6380f8c9a70216e344ccfb123a8be05da048a18f/lib/runners/schema/result.rb#L23

By removing this type, we can reduce the many needless codes and make the codebase simpler.

PR #1304, opened 2020-07-14
[Refactor] Remove `Processor#ensure_files` method

ybiquitous:

I will remove `Results::MissingFilesFailure` via #1121, but the `Processor#ensure_files` method depends on `Results::MissingFilesFailure`.
So, this change aims to remove `ensure_files`.

Related to #1121

> Check the following items.

- [] ~~Add a new [changelog](https://github.com/sider/runners/blob/master/CHANGELOG.md) entry if this change is notable.~~
  → Internal change.

### DIFF
--- a/lib/runners/processor/goodcheck.rb
+++ b/lib/runners/processor/goodcheck.rb
@@ -26,8 +26,8 @@ module Runners
     DEFAULT_TARGET = ".".freeze
     DEFAULT_CONFIG_FILE = "goodcheck.yml".freeze
 
-    def analyzer_version
-      @analyzer_version ||= extract_version! ruby_analyzer_bin, "version"
+    def extract_version_option
+      "version"
     end
 
     def goodcheck_test
@@ -109,41 +109,20 @@ module Runners
       end
     end
 
-    def setup
-      ret = install_gems default_gem_specs, constraints: CONSTRAINTS do
-        yield
+    def setup(&block)
+      if !config_linter[:config] && !File.exist?(DEFAULT_CONFIG_FILE)
+        return missing_config_file_result(DEFAULT_CONFIG_FILE)
       end
 
-      # NOTE: Exceptionally MissingFileFailure is treated as successful
-      if ret.instance_of? Results::MissingFilesFailure
-        trace_writer.error "File not found: goodcheck.yml"
-        add_warning(<<~MESSAGE)
-          Sider cannot find the required configuration file `goodcheck.yml`.
-          Please set up Goodcheck by following the instructions, or you can disable it in the repository settings.
-
-          - https://github.com/sider/goodcheck
-          - #{analyzer_doc}
-        MESSAGE
-        Results::Success.new(guid: guid, analyzer: analyzer)
-      else
-        ret
+      begin
+        install_gems default_gem_specs, constraints: CONSTRAINTS, &block
+      rescue InstallGemsFailure => exn
+        trace_writer.error exn.message
+        return Results::Failure.new(guid: guid, message: exn.message, analyzer: nil)
       end
-    rescue InstallGemsFailure => exn
-      trace_writer.error exn.message
-      return Results::Failure.new(guid: guid, message: exn.message, analyzer: nil)
     end
 
     def analyze(changes)
-      if !config_linter[:config] && !File.exist?(DEFAULT_CONFIG_FILE)
-        add_warning <<~MSG, file: DEFAULT_CONFIG_FILE
-          Sider could not find the required configuration file `#{DEFAULT_CONFIG_FILE}`.
-          Please create the file according to the following documents:
-          - #{analyzer_github}
-          - #{analyzer_doc}
-        MSG
-        return Results::Success.new(guid: guid, analyzer: analyzer)
-      end
-
       delete_unchanged_files(changes, except: ["*.yml", "*.yaml"])
 
       error_message = goodcheck_test
--- a/lib/runners/processor/goodcheck.rb
+++ b/lib/runners/processor/goodcheck.rb
@@ -110,10 +110,6 @@ module Runners
     end
 
     def setup(&block)
-      if !config_linter[:config] && !File.exist?(DEFAULT_CONFIG_FILE)
-        return missing_config_file_result(DEFAULT_CONFIG_FILE)
-      end
-
       begin
         install_gems default_gem_specs, constraints: CONSTRAINTS, &block
       rescue InstallGemsFailure => exn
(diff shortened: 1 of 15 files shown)


lib/runners/processor/javasee.rb
```diff
@@ -4,15 +4,10 @@ class Processor::Javasee < Processor
...
-          when String
-            string
-          when Array
-            array(string)
-          end
+        hash.merge!({
+          dir: enum?(string, array(string)),
+          config: string?,
         })
-        hash[:config] = string?
```
ybiquitous, line 15: [note] I think simplifying the code is better than too strict schema check.

lib/runners/processor/javasee.rb
```diff
@@ -80,14 +77,5 @@ def construct_result(result, stdout, stderr)
...
     end
-
-    def check_runner_config
-      dirs = Array(config_linter[:dir]).map { |dir| Pathname(dir) }
-
-      config_path = config_linter[:config]
-      config_path = Pathname(config_path) if config_path
-
-      yield dirs, config_path
-    end
```
ybiquitous, line 91: [note] Make the code more simple.

lib/runners/processor/goodcheck.rb
```diff
@@ -109,41 +109,20 @@ def goodcheck_check
       end
     end
 
-    def setup
-      ret = install_gems default_gem_specs, constraints: CONSTRAINTS do
-        yield
+    def setup(&block)
+      if !config_linter[:config] && !File.exist?(DEFAULT_CONFIG_FILE)
+        return missing_config_file_result(DEFAULT_CONFIG_FILE)
       end
```
ybiquitous, line 115: [note] Checking before `bundle install` to reduce useless installation time.

lib/runners/ruby.rb
```diff
@@ -77,6 +77,12 @@ def ruby_analyzer_bin
 
     def analyzer_version
       @analyzer_version ||= extract_version! ruby_analyzer_bin
+    rescue Shell::ExecError => exn
+      if exn.stderr_str.include? "Could not locate Gemfile or .bundle/ directory"
+        extract_version! analyzer_bin # fallback to default
+      else
+        raise
+      end
```
ybiquitous, line 85: [note] Prevent an error of `bundle exec <command>`.

lib/runners/processor/javasee.rb
```diff
@@ -24,39 +19,41 @@ class Processor::Javasee < Processor
...
-        elsif stderr.match?(/Configuration file .+ does not look a file/)
-          add_warning stderr
-          return Results::Success.new(guid: guid, analyzer: analyzer)
-        else
-          Results::Failure.new(
-            guid: guid,
-            analyzer: analyzer,
-            message: stderr,
-          )
+      if [0, 2].include?(status.exitstatus)
```
ybiquitous, line 47: > __[Querly]__ Note that `Process#exitstatus` can return `nil`. ([view](https://sider.review/gh/repos/202486306/pulls/1304#issue-4175240))
>
> <details>
>
> `Process#exitstatus` returns `nil` when the process does not exit normally.
> This means the process may be abort(2)-ed or core-dumped.
> If you want to know if it is aborted, consider using also `Process#exited?`.
>
> See also:
> * https://ruby-doc.org/core/Process/Status.html#method-i-exitstatus
>
> | Rule |
> | ---- |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1304#issue-4175240) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

I closed this issue, because:
- When you are not interested in aborted or core-dumped.

ybiquitous:
@yykamei Could you please review this pull request? Sorry for the many diffs... 🙇

yykamei:
I read it soon. No problem.

lib/runners/ruby.rb
yykamei: I wonder if we should include this change in this PR. If this fallback is necessary, the call of `ruby_analyzer_bin` in each processor should also get fall back? For example, `Brakeman#analyze` uses `ruby_analyzer_bin` [here](https://github.com/sider/runners/blob/f31fef373b38f91c206b9abe076002ffea36f08d/lib/runners/processor/brakeman.rb#L30). I think there are other things to be considered when making it handle the exception.

ybiquitous: Yes, this fallback code is ugly but necessary to prevent the following error:

```console
$ bundle exec rake docker:build docker:smoke ANALYZER=goodcheck ONLY=no_config_file SHOW_TRACE=true
...
:inspect => "#<Runners::Shell::ExecError: type=capture3, args=[\"bundle\", \"exec\", \"goodcheck\", \"version\"], status=#<Process::Status: pid 14 exit 10>, dir=#<Pathname:/tmp/d20200714-6-mzausr>, stdout_str=..., stderr_str=...>",
```

image

This `analyzer_version` is called from `analyzer`:

https://github.com/sider/runners/blob/e14a888c5216b1a8827321a41ce353b10e59bf37/lib/runners/processor.rb#L93-L95

https://github.com/sider/runners/blob/e14a888c5216b1a8827321a41ce353b10e59bf37/lib/runners/processor.rb#L150

Is there any good way...? 🤔 

---

If we give up calling `missing_config_file_result` in `setup`, the error is not raised.

```diff
def setup
- missing_config_file_result
end

def analyze
+ missing_config_file_result
end
```

ybiquitous: This error occurs in Querly and Goodcheck, so it seems better to call `missing_config_file_result` from `analyze` instead of `setup`.
What do you think?

yykamei: I agree. `missing_config_file_result` should be called in `analyze`. First, I thought it's OK to call `missing_config_file_result` in `setup`, but I want to change my idea. In the first place, other tools fail in **`analyze`** if the configuration file does not exist. Checking the files in `analyze` is reasonable.

Note: I know the reasons we should use `missing_config_file_result`.  Goodcheck, JavaSee, Phinder, Querly, and TyScan are created by us, and it's not acceptable for users to receive "Failure" without the configuration file.

ybiquitous: Thanks. I will call `missing_config_file_result` in `analyze` only for Goodcheck and Querl.
Because `install_gems` is required to call `missing_config_file_result`.

But for other tools, I will keep `missing_config_file_result` in `analyze`. Right?

yykamei: I think all tools should call it in `analyze` for the code consistency, but I leave it to you. Of course, I know calling `setup` reduces the analysis time.

ybiquitous: Thanks. I pushed 6d38df12.

PR #1307, opened 2020-07-15
Avoid git-blame when base commit is not given

ybiquitous:

Only when head and base commit are given, git-blame is useful.

None.

> Check the following items.

### DIFF
--- a/test/harness_test.rb
+++ b/test/harness_test.rb
@@ -49,7 +49,6 @@ class HarnessTest < Minitest::Test
   def test_run
     mktmpdir do |working_dir|
       with_options do |options|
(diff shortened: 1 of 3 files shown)


test/harness_test.rb
```diff
@@ -49,7 +49,6 @@ def initialize(*args)
   def test_run
     mktmpdir do |working_dir|
       with_options do |options|
-        mock.proxy(Workspace).prepare(options: options, working_dir: working_dir, trace_writer: trace_writer)
```
ybiquitous, line 52: [note] This mocking is needless.

PR #1312, opened 2020-07-15
Optimize git-blame

ybiquitous:

An output of git-blame is **idempotent**.
Thus, I think this optimization is effective, especially when there are many issues to be reported.

None.

> Check the following items.


ybiquitous:
Cached `GitBlameInfo` instances are just converted to `Hash`, so I think no problems to share them in the cache.

https://github.com/sider/runners/blob/75d1f96fd05d47a4b4843f4ee68a2f50de32446d/lib/runners/issue.rb#L57

https://github.com/sider/runners/blob/75d1f96fd05d47a4b4843f4ee68a2f50de32446d/lib/runners/issue.rb#L63-L64

ybiquitous:
Thanks for the review. I'll merge this when the CI passes.

PR #1314, opened 2020-07-16
[Cppcheck] Support the option `-I`

### LINKED ISSUE #1280: [cppcheck] Support the option `-I`
Cppcheck's preprocessor basically handles include paths. We consider support the `-I` option to make analysis more precise.

https://github.com/danmar/cppcheck/blob/main/man/manual.md#include-paths

See also #996

suetsuguk:

The -I option makes analysis more precise.

Fix #1280

> Check the following items.

- [ ] Write a document and link the PR sider/sider-docs#476


lib/runners/processor/cppcheck.rb
```diff
@@ -11,6 +13,8 @@ class Processor::Cppcheck < Processor
           project: string?,
           language: string?,
           'bug-hunting': boolean?
+        ).merge!(
+          { 'include-path': enum?(string, array(string)) } # TODO use FIELD_INCLUDE_PATH at cplusplus.rb
```
ybiquitous, line 17: [suggestion] It seems better to add `Schema::BaseConfig.cplusplus` like this:

https://github.com/sider/runners/blob/2203cb38aec088cc6d2ecf3498c441971d744702/lib/runners/schema/config.rb#L16

lib/runners/cplusplus.rb
```diff
@@ -0,0 +1,22 @@
+module Runners
+  module CPlusPlus
+#    FIELD_INCLUDE_PATH = { 'include-path': enum?(string, array(string)) }
+
+    def option_include_path
+      includes = Array(config_linter[:'include-path'] || find_paths_containing_headers)
+      includes.map { |v| "-I" + v }
+    end
```
ybiquitous, line 8: [suggestion]

- `config_include_path` seems better from `config_linter`.
- [nitpick] we can inline the `includes` variable.

```suggestion
    def config_include_path
      Array(config_linter[:'include-path'] || find_paths_containing_headers).map { |v| "-I#{v}" }
    end
```

lib/runners/cplusplus.rb
```diff
@@ -0,0 +1,24 @@
...
+    def is_source_file?(path)
+      path.fnmatch?(GLOB_SOURCES, File::FNM_EXTGLOB | File::FNM_CASEFOLD)
+    end
+
+    private
+
+    # @see https://github.com/github/linguist/blob/775b07d40c04ef6e0051a541886a8f1e30a892f4/lib/linguist/languages.yml#L532-L535
+    # @see https://github.com/github/linguist/blob/775b07d40c04ef6e0051a541886a8f1e30a892f4/lib/linguist/languages.yml#L568-L584
+    GLOB_SOURCES = "*.{c,cpp,c++,cc,cp,cxx}".freeze
+    GLOB_HEADERS = "**/*.{h,h++,hh,hpp,hxx,inc,inl,ipp,tcc,tpp}".freeze
```
yykamei, line 16: [`private_constant`](https://ruby-doc.org/core-2.7.1/Module.html#method-i-private_constant) is appropriate if you want to make these constants invisible.

lib/runners/cplusplus.rb
```diff
@@ -0,0 +1,24 @@
+module Runners
+  module CPlusPlus
+    def config_include_path
+      Array(config_linter[:'include-path'] || find_paths_containing_headers).map { |v| "-I#{v}" }
+    end
+
+    def is_source_file?(path)
```
yykamei, line 7: This is just an opinion, and you don't have to update this PR, but the name of this method is a bit too general. Perhaps, this method could be defined in `Processor` with the `#glob_sources` method, which is supposed to be overridden by each Language-specific module.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - **stylelint** Add `stylelint-order` to pre-installed list [#1310](https://github.com/sider/runners/pull/1310)
 - Optimize git-blame [#1312](https://github.com/sider/runners/pull/1312)
 - Remove unused `Result::MissingFilesFailure` [#1315](https://github.com/sider/runners/pull/1315)
+- **cppcheck** Support `-I` option to handle include paths [#1314](https://github.com/sider/runners/pull/1314)
```
ybiquitous, line 13: [nitpick] See `analyzers.yml`

```suggestion
- **Cppcheck** Support `-I` option to handle include paths [#1314](https://github.com/sider/runners/pull/1314)
```

lib/runners/cplusplus.rb
ybiquitous: I agree with @yykamei. So, what about the method name `cpp_file?`

lib/runners/cplusplus.rb
ybiquitous: Agree 👍 Also, how about them? 👇 

- Put the constants at the top of this module so that we can find them easily.
- Rename them to clarify they are related to C/C++.

```ruby
module CPlusPlus
  CPP_SOURCES_GLOB = "*.{c,cpp,c++,cc,cp,cxx}".freeze
  CPP_HEADERS_GLOB = "**/*.{h,h++,hh,hpp,hxx,inc,inl,ipp,tcc,tpp}".freeze
  private_constant :CPP_SOURCES_GLOB, :CPP_HEADERS_GLOB

  # some methods...
end
```

lib/runners/processor/clang_tidy.rb
```diff
@@ -1,10 +1,11 @@
...
   class Processor::ClangTidy < Processor
+    include CPlusPlus
+
     Schema = StrongJSON.new do
-      let :runner_config, Schema::BaseConfig.base.update_fields { |fields|
+      let :runner_config, Schema::BaseConfig.cplusplus.update_fields { |fields|
         fields.merge!({
-          apt: enum?(string, array(string)),
-          'include-path': enum?(string, array(string))
+          apt: enum?(string, array(string))
```
ybiquitous, line 8: What about moving the `apt` option to `Schema::BaseConfig.cplusplus`?

suetsuguk: I think it is not too late to tackle this when we face a new analyze that requires development packages.
But, now that we are working on both Clang-Tidy and Cppcheck, we should make use of this opportunity. They say strike while the iron is hot. I will move it to the base config.

suetsuguk:
Can anyone tell me how to solve the Steep error?

```console
$ bundle exec rake  typecheck
...
lib/runners/cplusplus.rb:7:4: NoMethodError: type=(::Module & ::Runners::CPlusPlus.module), method=private_constant (private_constant :CPP_SOURCES_GLOB, :CPP_HEADERS_GLOB)
rake aborted!
...
```

 Simply removing the line makes the check succeed. But I'm wondering if you have a proper solution.

yykamei:
Ah, Steep v0.11 lacks some standard library APIs. Perhaps, this resolves the problem.

```diff
diff --git a/sig/polyfill.rbi b/sig/polyfill.rbi
index 5f2a4b22..af8c2af2 100644
--- a/sig/polyfill.rbi
+++ b/sig/polyfill.rbi
@@ -43,6 +43,7 @@ extension Module (Polyfill)
   def instance_method: (Symbol) -> UnboundMethod
   def define_method: (Symbol) { () -> any } -> Symbol
   def method_defined?: (Symbol, ?bool) -> bool
+  def private_constant: (*Symbol) -> self
 end
 
 extension String (Polyfill)
```

suetsuguk:
This fix works fine. Thank you.

lib/runners/cplusplus.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def config_include_path
+      Array(config_linter[:'include-path'] || find_paths_containing_headers).map { |v| "-I#{v}" }
+    end
+
+    def cpp_file?(path)
+      path.fnmatch?(CPP_SOURCES_GLOB, File::FNM_EXTGLOB | File::FNM_CASEFOLD)
+    end
+
+    def deploy_packages
```
ybiquitous, line 17: [suggestion] I think the name `install_apt_packages` describes the method behavior more clearly. What do you think?

```suggestion
    def install_apt_packages
```

lib/runners/cplusplus.rb
```diff
@@ -0,0 +1,44 @@
...
+          if pkg.match?(/-dev(=.+)?$/)
+            true
+          else
+            add_warning "Installing the package `#{pkg}` is blocked.", file: config.path_name
+            false
+          end
+        end
+
+      if packages.empty?
+        trace_writer.message "No packages to install."
```
ybiquitous, line 30: [nitpick] How about this? 👇 

```suggestion
        trace_writer.message "No apt packages to install."
```

lib/runners/cplusplus.rb
```diff
@@ -0,0 +1,44 @@
...
+    def config_include_path
+      Array(config_linter[:'include-path'] || find_paths_containing_headers).map { |v| "-I#{v}" }
+    end
+
+    def cpp_file?(path)
+      path.fnmatch?(CPP_SOURCES_GLOB, File::FNM_EXTGLOB | File::FNM_CASEFOLD)
+    end
+
+    def deploy_packages
+      # select development packages and report others as warning for security concerns
```
ybiquitous, line 18: [suggestion] How about adding the following trace message to make our debugging easier?

```suggestion
      trace_writer.message "Installing apt packages..."

      # select development packages and report others as warning for security concerns
```

lib/runners/cplusplus.rb
suetsuguk: Agreed. commit 035f9656607796996feb63b06b8bce200e43cd95

lib/runners/cplusplus.rb
suetsuguk: commit 035f9656607796996feb63b06b8bce200e43cd95

lib/runners/cplusplus.rb
suetsuguk: commit 035f9656607796996feb63b06b8bce200e43cd95

suetsuguk:
Thank you all for taking your time.
I will merge this to master.

PR #1315, opened 2020-07-16
Remove unused `Result::MissingFilesFailure`

### LINKED ISSUE #1121: Remove `missing_files` from result

ybiquitous:

It's self-explanatory.

Fix #1121

> Check the following items.

  → ~~Internal change.~~
  → This removed class is unused but the interface change. So, I re-think it should be included in the changelog.


ybiquitous:
Oh, I missed removing below, so I'll update it soon:

https://github.com/sider/runners/blob/2203cb38aec088cc6d2ecf3498c441971d744702/lib/runners/schema/result.rb#L24

> Let's remove the code related to this on our application server later 💪

Good catch! I'll do it later. 👍

yykamei:
I missed it, too. LGTM

ybiquitous:
Thank you for the review! 😺

PR #1320, opened 2020-07-20
Avoid needless calculation of changed files

### LINKED ISSUE #1317: Avoid needless calculation of changed files
When a base commit is **not** given, the calculation of changed files is needless:

https://github.com/sider/runners/blob/265cb5a12907f57eab10fe6e64e7879cb9b399f4/lib/runners/workspace.rb#L53

ybiquitous:

This change aims to avoid the needless calculation in the `Changes` class
by improving the method signatures of the class.

> Link related issues or pull requests, e.g. "Fix #<ISSUE_ID>", "Related to #<PULL_NUMBER>", or "None."

Fix #1317

> Check the following items (remove a needless item).


lib/runners/changes.rb
```diff
@@ -47,44 +47,28 @@ def include?(issue)
...
-        .filter_map { |f| f.relative_path_from(working_dir) if f.file? }
-
-      changed_paths = patches.files.map { |f| Pathname(f) }
-
-      new(changed_paths: changed_paths.sort!,
-          unchanged_paths: (all_files - changed_paths).sort!,
-          patches: patches)
+    def self.all_files_in(basedir)
+      basedir.glob("**/*", File::FNM_DOTMATCH).filter_map do |file|
+        if file.file? && !file.fnmatch?("**/.git/**", File::FNM_DOTMATCH)
```
yykamei, line 67: Don't we have to exclude files in the subdirectory of `.git`?

```suggestion
        if file.file? && !file.fnmatch?("**/.git/**/*", File::FNM_DOTMATCH)
```

```irb
irb(main):001:0> Pathname('.').glob('**/.git/**', File::FNM_DOTMATCH).size
=> 15
irb(main):002:0> Pathname('.').glob('**/.git/**/*', File::FNM_DOTMATCH).size
=> 1934
```

ybiquitous: Good catch! But, `**/.git/**/*` does not match `.git/config`... 😓 
I'll reconsider this logic, thanks. 👍

ybiquitous: Fixed via e322efd8. 🎉 

Correct pattern: `**/.git/*`

yykamei: :thinking: Does this differ among system environments? My environment matches with `.git/config`.

```ruby
irb(main):001:0> Pathname('.').glob('**/.git/**/*', File::FNM_DOTMATCH).include?(Pathname('.git/config'))
=> true
```

ybiquitous: Because behavios of `glob` and `fnmatch?` differ:

```
irb(main):005:0> Pathname('.git/config').realpath.fnmatch?('**/.git/*', File::FNM_DOTMATCH)
=> true
irb(main):006:0> Pathname('.git/info/refs').realpath.fnmatch?('**/.git/*', File::FNM_DOTMATCH)
=> true
```

yykamei: Thank you. I misread the method as `glob` (but `fnmatch?` has been used actually).

ybiquitous:
Thanks for the review. 👍

PR #1316, opened 2020-07-16
Remove unused HTTP and File workspace

ybiquitous:
> Explain a summary, purpose, or background for this change.

`Workspace::HTTP` and `Workspace::File` classes have no longer used. (Only `Workspace::Git` is used)
This pull request aims to reduce the code.

> Link related issues, e.g. "Fix #<ISSUE_ID>", "Related to #<ISSUE_ID>", or "None."

Fix #1146 

> Check the following items.

- [] ~~Add a new [changelog](https://github.com/sider/runners/blob/master/CHANGELOG.md) entry if this change is notable.~~
  → Internal change.

PR #1323, opened 2020-07-21
Fix non-ascii filename on git-diff

ybiquitous:

git-diff outputs quoted filenames by default.
`GitDiffParser::Patches` does not handle these quoted filenames.

See below for `core.quotePath`:

https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath

None.

### DIFF
--- a/test/workspace/git_test.rb
+++ b/test/workspace/git_test.rb
@@ -63,16 +63,17 @@ class WorkspaceGitTest < Minitest::Test
     end
   end
 
-  def test_patches_returns_nil_because_base_is_nil
-    with_workspace do |workspace|
+  def test_patches_returns_nil
+    with_workspace(base: nil) do |workspace|
       assert_nil workspace.send(:patches)
     end
   end
 
-  def test_patches_returns_patches
-    with_workspace(base: base_commit) do |workspace|
+  def test_patches
+    with_workspace(head: "836880fdd04e5e1d7d82ed17dae838a16cfa50b2", base: base_commit) do |workspace|
       workspace.send(:prepare_head_source, nil)
-      assert_instance_of GitDiffParser::Patches, workspace.send(:patches)
+      patches = workspace.send(:patches)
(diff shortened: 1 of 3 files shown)


ybiquitous:
NOTE: The current smoke tests use `Workspace::File` instead of `Workspace::Git`, so this bug cannot be detected by the smoke tests. But this problem will be resolved by #1316.

test/workspace/git_test.rb
```diff
@@ -63,16 +63,17 @@ def test_git_fetch_args_without_pull_number
...
+  def test_patches_returns_nil
+    with_workspace(base: nil) do |workspace|
       assert_nil workspace.send(:patches)
     end
   end
 
-  def test_patches_returns_patches
-    with_workspace(base: base_commit) do |workspace|
+  def test_patches
+    with_workspace(head: "836880fdd04e5e1d7d82ed17dae838a16cfa50b2", base: base_commit) do |workspace|
```
ybiquitous, line 73: https://github.com/sider/runners_test/commit/836880fdd04e5e1d7d82ed17dae838a16cfa50b2

ybiquitous:
Good catch! Maybe it is not necessary.

Also, I'm now doubting whether `git init` and `git add .` are necessary. 🤔 

https://github.com/sider/runners/blob/0d828c3eecb83e1581a8159e59942506eb870f1a/lib/runners/ignoring.rb#L41-L42

I will address this issue on another pull request, but no problem?

yykamei:
I think so, git-init(1) and git-add(1) also seem unnecessary.

ybiquitous:
`git init` and `git add .` are **not** necessary. But, if removing the commands, `ignoring_test.rb` fails.
So, I will refactor `ignoring.rb` on another pull request.

ybiquitous:
@yykamei I've opened another PR #1324. Can I merge this PR?

ybiquitous:
Never mind. Thanks for the review. 😄

PR #1324, opened 2020-07-21
Refactor `Ignoring` class

ybiquitous:

This change aims to improve the readability of the code around the `Ignoring` class.
The `Ignoring` class uses internally the `git` command, so I think the class depends on `Workspace::Git` implicitly.

Thus, this change clarifies the dependency on `Workspace::Git`.
After this change is merged, I will remove unnecessary the `git init` command finally.
(The smoke test, which depends on `Workspace::File`, would fail if removing `git init` in this change).

See also #1316 and <https://github.com/sider/runners/pull/1323#pullrequestreview-452240525>

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -7,7 +7,7 @@ target :lib do
   check "lib/runners/cli.rb"
   check "lib/runners/config.rb"
   check "lib/runners/errors.rb"
-  check "lib/runners/ignoring.rb"
+  # check "lib/runners/ignoring.rb"
   check "lib/runners/shell.rb"
   # TODO: Add checking...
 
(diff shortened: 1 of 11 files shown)


Steepfile
```diff
@@ -7,7 +7,7 @@ target :lib do
   check "lib/runners/cli.rb"
   check "lib/runners/config.rb"
   check "lib/runners/errors.rb"
-  check "lib/runners/ignoring.rb"
+  # check "lib/runners/ignoring.rb"
```
yykamei, line 10: Let's add type signatures later :muscle:

ybiquitous: It's a hard task... 😅 (#1215)

ybiquitous:
Thanks for the review. In #1316, we can remove `core.quotePath` in `Ignoring` class. 👍

PR #1351, opened 2020-07-29
Refactor IO classes

ybiquitous:

This change aims to follow the interface of the standard library [`IO`](https://ruby-doc.org/core-2.7.1/IO.html) class.

None.

### DIFF
--- a/lib/runners/io.rb
+++ b/lib/runners/io.rb
@@ -1,21 +1,17 @@
 module Runners
   class IO
-    REQUIRED_METHODS_FOR_IOS = %i[write flush].freeze
-
     attr_reader :ios
 
     def initialize(*ios)
-      @ios = ios.each do |io|
-        raise ArgumentError unless REQUIRED_METHODS_FOR_IOS.all? { |method| io.respond_to?(method) }
-      end
+      @ios = ios
     end
 
     def write(*args)
       ios.each { |io| io.write(*args) }
     end
 
-    def flush(*args)
-      ios.each { |io| io.flush(*args) }
+    def flush
+      ios.each { |io| io.flush }
     end
 
     def flush!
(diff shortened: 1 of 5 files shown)


lib/runners/io.rb
```diff
@@ -1,21 +1,17 @@
...
     end
 
     def write(*args)
       ios.each { |io| io.write(*args) }
     end
 
-    def flush(*args)
-      ios.each { |io| io.flush(*args) }
+    def flush
+      ios.each { |io| io.flush }
```
ybiquitous, line 14: [note] `flush` has no arguments. See https://ruby-doc.org/core-2.7.1/IO.html#flush-method

lib/runners/io.rb
```diff
@@ -1,21 +1,17 @@
...
   class IO
-    REQUIRED_METHODS_FOR_IOS = %i[write flush].freeze
-
     attr_reader :ios
 
     def initialize(*ios)
-      @ios = ios.each do |io|
-        raise ArgumentError unless REQUIRED_METHODS_FOR_IOS.all? { |method| io.respond_to?(method) }
-      end
+      @ios = ios
```
ybiquitous, line 6: [note] The tests check if a passed object has a correct interface.

lib/runners/io/aws_s3.rb
```diff
@@ -31,11 +30,10 @@ def initialize(uri)
...
         instance_profile_credentials_timeout: 3,
-      }.tap do |hash|
-        if ENV["S3_ENDPOINT"]
-          hash[:endpoint] = ENV["S3_ENDPOINT"]
-          hash[:force_path_style] = true
-        end
+      }
+      if ENV["S3_ENDPOINT"]
+        args[:endpoint] = ENV["S3_ENDPOINT"]
+        args[:force_path_style] = true
```
ybiquitous, line 36: [note] I think this usage of `tap` is needless due to extra indentations.

PR #1361, opened 2020-07-30
Install 3rd-party JVM dependencies via `sider.yml`

### LINKED ISSUE #1000: Support third-party dependencies for JVM tools

ybiquitous:

This change allows users to install 3rd-party JVM dependencies via `sider.yml`.

For example, users can do it to use 3rd-party Checkstyle rules:

```yaml
linter:
  checkstyle:
    jvm_deps:
      # [group, name, version] format
      - [com.github.sevntu-checkstyle, sevntu-checks, 1.37.1]
```

NOTE:
I suggested `maven` instead of `jvm_deps` in [this comment](https://github.com/sider/runners/issues/1000#issuecomment-657994815), but I think it not good that naming depends on a specific tool name (e.g. Maven, Gradle, etc.).
So, I renamed it to `jvm_deps` and adopted only the format `[group, name, version]`.

Fix #1000

### DIFF
--- a/lib/runners/schema/config.rb
+++ b/lib/runners/schema/config.rb
@@ -37,6 +37,10 @@ module Runners
       let :npm, (base.update_fields { |fields|
         fields.merge!(npm_install: optional(npm_install))
       })
+
+      let :java, (base.update_fields { |fields|
+        fields.merge!(jvm_deps: array?(array(string)))
+      })
     end
 
     Config = _ = StrongJSON.new do
--- a/lib/runners/schema/config.rb
+++ b/lib/runners/schema/config.rb
@@ -39,7 +39,7 @@ module Runners
       })
 
       let :java, (base.update_fields { |fields|
-        fields.merge!(jvm_deps: array?(array(string)))
+        fields.merge!(jvm_deps: array?(array(enum(string, number))))
       })
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Updated tools:
 Misc:
 
 - Download JVM dependencies via Gradle [#1352](https://github.com/sider/runners/pull/1352)
+- Install 3rd-party JVM dependencies via `sider.yml` [#1361](https://github.com/sider/runners/pull/1361)
 
 ## 0.31.0
 
--- a/lib/runners/java.rb
+++ b/lib/runners/java.rb
@@ -68,7 +68,7 @@ module Runners
           #{deps}
           }
           task deps(type: Copy) {
-            from configurations.runtimeClasspath
+            from configurations.compileClasspath
             into '.'
           }
         GRADLE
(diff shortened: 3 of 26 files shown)


ybiquitous:
@yykamei This pull request is still in progress, but can you please give me feedback if you have?

lib/runners/schema/config.rb
```diff
@@ -37,6 +37,10 @@ module Schema
       let :npm, (base.update_fields { |fields|
         fields.merge!(npm_install: optional(npm_install))
       })
+
+      let :java, (base.update_fields { |fields|
+        fields.merge!(jvm_deps: array?(array(string)))
```
yykamei, line 42: `version` may be `number`?: `[com.foo, bar, 1.2]`.

ybiquitous: Good catch! If we could support `number`, users would feel nice. 👍

ybiquitous:
> Perhaps, we may have to let users add other repositories like JCenter, but it might be rare cases.

Good point. I think this case might occur in the future. It seems good that we consider it now.

Here are my suggestions:

1. Increase repositories by default defined in <https://docs.gradle.org/current/userguide/declaring_repositories.html>:

- `mavenCentral`
- `jcenter`
- `google`

2. Custom repositories, especially in Enterprise, might is necessary; e.g. a company private repository (see also https://cloudsmith.com/maven-repository/)

I'm not sure how to realize it now, but for example:

```yaml
linter:
  checkstyle:
    jvm_repos: # default to [mavenCentral, jcenter, google]
      - foo
      - bar
      - ...
```

In this pull request, I will implement the **1** option, but the **2** option is not necessary for now.
What do you think?

lib/runners/schema/config.rb
ybiquitous: Fixed via 122f02d

yykamei:
> In this pull request, I will implement the **1** option, but the **2** option is not necessary for now.
> What do you think?

I agree. If the option **2** is supplementally implementable later, this PR is enough to implement the option **1**.

ybiquitous:
Thanks. I implemented **1** via cee2aa27.

ybiquitous:
The ktlint processor has a special schema:

https://github.com/sider/runners/blob/7d156cc9d1ad645f96613b3c82aa427935cdea59/lib/runners/processor/ktlint.rb#L28-L31

So, I think it proper that The ktlint processor does not support this feature yet (we may need to reconsider the schema, but breaking).

ybiquitous:
@yykamei Could you please re-review?

ybiquitous:
Thanks for the review. When the CI passes, I will merge this PR. 👌

PR #1371, opened 2020-08-04
[RBP] Improve error message

ybiquitous:

The output is empty in rare cases.
Then, `YAML.safe_load("")` returns `nil`, so `NoMethodError` occurs.

This change aims to make debugging easier by improving the error message.

None.


lib/runners/processor/rails_best_practices.rb
```diff
@@ -111,23 +111,30 @@ def prepare_config
...
+      )
+
+      # NOTE: rails_best_practices returns top-level YAML array with tags.
+      #       We want to just parse a YAML, so we remove YAML tags before passing.
+      output = read_report_file
+      output.gsub!('- !ruby/object:RailsBestPractices::Core::Error', '-')
+
+      issues = YAML.safe_load(output, symbolize_names: true, filename: report_file)
+      unless issues.is_a? Array
+        raise "Invalid YAML result: `#{issues.inspect}`"
```
ybiquitous, line 133: [note] Unfortunately, this case cannot be reproduced in a smoke test. 😢

PR #1411, opened 2020-08-24
Remove :owner and :repo, and rename :git_http_url and :git_http_userinfo

### LINKED ISSUE #1403: Support Git URL that includes more than 2 nested paths like https://example.com/foo/bar/project
Runners assume the Git remote URL has one parent path and its child path, such as `https://example.com/foo/project`. We call `foo` as an `owner` and `project` as a `repo`, and they are used to construct a Git remote URL here.

https://github.com/sider/runners/blob/50ebb5be65433470547126ec0acb022a838b807d/lib/runners/workspace/git.rb#L64

However, I want Runners to handle a remote URL with more than two nested paths like `https://example.com/foo/bar/project`. To support such URLs, I want to make Runners handle `git_http_url` as a complete Git remote URL and remove `owner` and `repo` from `Options::GitSource`.

What do you think?

yykamei:
Runners assume the Git remote URL has one parent path and its child
path, such as `https://example.com/foo/project`. We call foo as an `owner`
and project as a `repo`, and they are used to construct a Git remote URL
here. As you know from these terms, we assumes the Git repositories are
on GitHub.

However, Runners can handle other Git repositories other than on GitHub,
such as GitLab. That's why this change made Runners handle the Git
remote URL fairly universal.

Now, `git_http_url` should be `git_url`, and `git_http_userinfo` should
be `git_url_user_info`.

Close #1403

To-Do

- [x] Add a new [changelog](https://github.com/sider/runners/blob/master/CHANGELOG.md) and explains this is a braking change

### DIFF
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -56,11 +56,6 @@ module Runners
     end
 
     def remote_url
-      # For smoke test
-      if git_source.git_url.start_with? "file://"
-        return URI(git_source.git_url)
-      end
-
       @remote_url ||= URI(git_source.git_url).tap do |uri|
         git_url_userinfo = git_source.git_url_userinfo
         uri.userinfo = git_url_userinfo if git_url_userinfo
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -57,13 +57,13 @@ module Runners
 
     def remote_url
       # For smoke test
-      if git_source.git_http_url.start_with? "file://"
-        return URI(git_source.git_http_url)
+      if git_source.git_url.start_with? "file://"
+        return URI(git_source.git_url)
       end
 
-      @remote_url ||= URI.join(git_source.git_http_url, "#{git_source.owner}/#{git_source.repo}").tap do |uri|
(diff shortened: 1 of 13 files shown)


lib/runners/workspace/git.rb
```diff
@@ -56,14 +56,9 @@ def sleep_lambda
     end
 
     def remote_url
-      # For smoke test
-      if git_source.git_http_url.start_with? "file://"
-        return URI(git_source.git_http_url)
-      end
```
yykamei, line 62: I removed these lines because we no longer have to give special treatments for such URLs.

ybiquitous (approved):
LGTM 👍

yykamei:
`git_url_userinfo` also could be removed, but before doing this, we should ensure the security breach may not happen.

ybiquitous:
I think we should not remove `git_url_userinfo` because it is hard to detect and filter sensitive data embedded in a URL string.

yykamei:
Yes, that's right. This is what I wanted to say. Thank you for clarifying.
Thanks for your review. I merge this PR and release the new version "0.34.0"

ybiquitous:
Agree. I recommend to try the canary test before releasing a new version.

PR #1412, opened 2020-08-24
Replace :pull_number with :refspec

### LINKED ISSUE #1395: Support refspecs for merge requests on GitLab
The current implementation will add a refspec for pull requests on GitHub here.

https://github.com/sider/runners/blob/5133dee755004c9ee4d4b90cd66f847dd07a21e7/lib/runners/workspace/git.rb#L74-L75

Now, Sider supports GitLab, so we should add a valid refspec for each integration. GitLab's refspec is like `+refs/merge-requests/*/head:refs/remotes/origin/merge-requests/*` according to the [document](https://docs.gitlab.com/ee/user/project/merge_requests/reviewing_and_managing_merge_requests.html#checkout-locally-by-modifying-gitconfig-for-a-given-repository).

Alternatively, it might be sufficient to just add the new refspec. Maybe.

yykamei:
This change allows analyses for any Git repositories with additional refspecs. The previous implementation assumes that the Git repositories are on GitHub, so it had prepared the `:pull_number` parameter to construct an additional refspec. We should support other platforms, such as GitLab. So, Runners will receive the new parameter `:refspecs`. From
this change, the Runners users should construct refspecs by themselves and pass them to Runners.

Close #1395

To-Do:

- [x] Add a new [changelog](https://github.com/sider/runners/blob/master/CHANGELOG.md) as it contains a breaking change

### DIFF
--- a/lib/runners/schema/options.rb
+++ b/lib/runners/schema/options.rb
@@ -8,7 +8,7 @@ module Runners
         base: string?,
         git_url: string,
         git_url_userinfo: string?,
-        pull_number: number?,
+        refspecs: array?(string),
       )
 
       let :payload, object(
--- a/lib/runners/schema/options.rb
+++ b/lib/runners/schema/options.rb
@@ -8,7 +8,7 @@ module Runners
         base: string?,
         git_url: string,
         git_url_userinfo: string?,
-        refspecs: array?(string),
+        refspec: enum?(string, array(string)),
       )
 
       let :payload, object(
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -63,12 +63,14 @@ module Runners
     end
 
     def git_fetch_args
-      @git_fetch_args ||= %w[--quiet --no-tags --no-recurse-submodules origin].tap do |command|
-        command << "+refs/heads/*:refs/remotes/origin/*"
-
-        refspecs = git_source.refspecs
-        refspecs.each { |refspec| command << refspec } if refspecs
-      end
+      @git_fetch_args ||= [
+        '--quiet',
+        '--no-tags',
+        '--no-recurse-submodules',
+        'origin',
+        '+refs/heads/*:refs/remotes/origin/*',
+        *Array(git_source.refspec),
+      ]
     end
 
     def patches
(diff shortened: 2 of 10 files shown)


lib/runners/schema/options.rb
```diff
@@ -8,7 +8,7 @@ module Schema
         base: string?,
         git_url: string,
         git_url_userinfo: string?,
-        pull_number: number?,
+        refspecs: array?(string),
```
ybiquitous, line 11: [suggestion] How about using the name `refspec` instead of `refspecs` and allowing one refspec as a string value?

```suggestion
        refspec: enum?(string, array(string)),
```

lib/runners/workspace/git.rb
```diff
@@ -66,8 +66,8 @@ def git_fetch_args
       @git_fetch_args ||= %w[--quiet --no-tags --no-recurse-submodules origin].tap do |command|
         command << "+refs/heads/*:refs/remotes/origin/*"
 
-        num = git_source.pull_number
-        command << "+refs/pull/#{num}/head:refs/remotes/pull/#{num}/head" if num
+        refspecs = git_source.refspecs
+        refspecs.each { |refspec| command << refspec } if refspecs
       end
     end
```
ybiquitous, line 72: [suggestion] How about this refactoring? (`Array(nil)` returns `[]`)

```ruby
def git_fetch_args
  @git_fetch_args ||= [
    "--quiet", "--no-tags", "--no-recurse-submodules", "origin",
    "+refs/heads/*:refs/remotes/origin/*",
    *Array(git_source.refspecs),
  ].freeze
end
```

ybiquitous (commented):
I think this change has no problems. Please check out my refactoring suggestions.

suetsuguk (approved):
LGTM

lib/runners/schema/options.rb
yykamei: I will change it as you said.

Note I deliberately defined it not using `enum?` because I want to make it obvious from the parameter name. I thought this parameter is basically called by developers, not end-users, so clarity was important.

ybiquitous (approved):
LGTM 👍 
(I've edited the pull request title.)

yykamei:
Thanks, guys.

PR #1415, opened 2020-08-24
Small refactoring of `Runners::Ignoring` class

ybiquitous:

It is self-explanatory from the PR title.

None.

### DIFF
--- a/lib/runners/ignoring.rb
+++ b/lib/runners/ignoring.rb
@@ -9,11 +9,12 @@ module Runners
 
     def delete_ignored_files!
       Tempfile.create("gitignore-") do |gitignore|
-        File.write gitignore, ignore_patterns.join("\n")
+        ignore_path = gitignore.path
+        File.write ignore_path, ignore_patterns.join("\n")
 
         # @see https://git-scm.com/docs/git-ls-files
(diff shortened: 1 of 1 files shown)


lib/runners/ignoring.rb
```diff
@@ -9,11 +9,12 @@ def initialize(workspace:, ignore_patterns:)
 
     def delete_ignored_files!
       Tempfile.create("gitignore-") do |gitignore|
```
ybiquitous, line 11: [note] `gitignore` is a `File` object:

> If a block is given, then a File object will be constructed, and the block is invoked with the object as the argument.

https://ruby-doc.org/stdlib-2.7.1/libdoc/tempfile/rdoc/Tempfile.html#create-method

lib/runners/ignoring.rb
```diff
@@ -9,11 +9,12 @@ def initialize(workspace:, ignore_patterns:)
...
-        File.write gitignore, ignore_patterns.join("\n")
+        ignore_path = gitignore.path
+        File.write ignore_path, ignore_patterns.join("\n")
```
ybiquitous, line 13: [note] Explicitly pass a string to `File.write`:

https://ruby-doc.org/core-2.7.1/IO.html#write-method

PR #1424, opened 2020-08-25
Refactor test code

ybiquitous:

- Unify format: `require "test_helper"`.
- Remove the `rr` gem. Instead, use `minitest/mock`.

None.

### DIFF
--- a/lib/runners/workspace.rb
+++ b/lib/runners/workspace.rb
@@ -2,8 +2,6 @@ module Runners
   class Workspace
     include Tmpdir
 
-    class DownloadError < SystemError; end
-
     def self.prepare(options:, working_dir:, trace_writer:)
       Workspace::Git.new(options: options, working_dir: working_dir, trace_writer: trace_writer)
     end
--- a/test/options_test.rb
+++ b/test/options_test.rb
@@ -1,10 +1,10 @@
-require_relative 'test_helper'
+require 'test_helper'
 
 class OptionsTest < Minitest::Test
   include TestHelper
 
-  def setup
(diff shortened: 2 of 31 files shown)


lib/runners/workspace.rb
```diff
@@ -2,8 +2,6 @@ module Runners
   class Workspace
     include Tmpdir
 
-    class DownloadError < SystemError; end
-
```
ybiquitous, line 6: [note] Unused.

test/options_test.rb
```diff
@@ -1,10 +1,10 @@
...
+require 'test_helper'
 
 class OptionsTest < Minitest::Test
   include TestHelper
 
-  def setup
-    stub(Runners::IO::AwsS3).stub? { true }
+  def run
+    Runners::IO::AwsS3.stub(:stub?, true) { super }
   end
```
ybiquitous, line 8: [note] Override `Minitest::Test#run`.

yykamei (commented):
LGTM

Please explain why you did these tasks :pray: 

> * Unify format: require "test_helper".
> * Remove the rr gem. Instead, use minitest/mock.

ybiquitous:
> - Unify format: require "test_helper".

If we are using `require_relative "../test_helper"`, we need to consider a test file location and a relative path to `test_helper.rb`. It is not useful.

> - Remove the `rr` gem. Instead, use `minitest/mock`.

I think that using multiple mock libraries leads to the source of confusion and high learning cost.
Minitest is very popular (de facto standard), while RR is not.

yykamei (approved):

ybiquitous:
Thanks for the review! 👍

PR #1426, opened 2020-08-26
Refactor value classes

ybiquitous:

- Stop using `Struct` in the `GitBlameInfo` class for better type-checking (See #1215)
- Remove the unused `Location.from_json` method
- Fix some trivial bugs

See also #1215

### DIFF
--- a/lib/runners/issue.rb
+++ b/lib/runners/issue.rb
@@ -25,6 +25,7 @@ module Runners
       @message = message
       @links = links
       @object = object
+      @git_blame_info = nil
     end
 
     def missing_id?
@@ -38,7 +39,8 @@ module Runners
         other.id == id &&
         other.message == message &&
         other.links == links &&
-        other.object == object
+        other.object == object &&
+        other.git_blame_info == git_blame_info
     end
     alias eql? ==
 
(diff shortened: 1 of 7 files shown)


lib/runners/location.rb
```diff
@@ -25,31 +25,17 @@ def hash
...
-    def self.from_json(json)
-      self.new(start_line: json[:start_line],
-               start_column: json[:start_column],
-               end_line: json[:end_line],
-               end_column: json[:end_column])
-    end
-
     def as_json
-      # @type _: Hash<json_key, Integer>
-      @as_json ||= _ = Schema::Issue.location.coerce({
```
ybiquitous, line 37: [note] `Issue#as_json` does schema-checking, so this checking is needless:

https://github.com/sider/runners/blob/766833f03179903377131a6e996ee327f0304247/lib/runners/issue.rb#L50

lib/runners/issue.rb
```diff
@@ -38,7 +39,8 @@ def ==(other)
         other.id == id &&
         other.message == message &&
         other.links == links &&
-        other.object == object
+        other.object == object &&
+        other.git_blame_info == git_blame_info
```
ybiquitous, line 43: [note] The comparison of `git_blame_info` is missing, but this has no effects actually.

PR #1459, opened 2020-09-10
Reconsider `Results::Failure` message

ybiquitous:

The message set to `Results::Failure` may be shown in the UI.
So, I think it's not user-friendly to pass the STDOUT or STDERR output to the message.
(users can see the STDOUT or STDERR output in the log page)

None.


lib/runners/processor/hadolint.rb
```diff
@@ -68,8 +68,8 @@ def run_analyzer
         Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
           parse_result(stdout).each { |v| result.add_issue(v) }
         end
-      rescue JSON::ParserError
-        Results::Failure.new(guid: guid, analyzer: analyzer, message: stderr)
+      rescue JSON::ParserError => exn
+        Results::Failure.new(guid: guid, analyzer: analyzer, message: exn.message)
```
ybiquitous, line 72: > __[Querly]__ Be careful whether the `message` is user-friendly. ([view](https://sider.review/gh/repos/202486306/pulls/1459#issue-4552134))
>
> <details>
>
> The message set to `Results::Failure` may be shown in the UI.
> For example, setting the STDERR output as the message may not be user-friendly. (a too long stack trace, etc.)
>
> | Rule |
> | ---- |
> | `sider.runners.failure_message` |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1459#issue-4552134) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

以下の理由でこのIssueをクローズしました：
- When setting a user-friendly message.

test/smokes/eslint/expectations.rb
```diff
@@ -155,7 +155,7 @@
 s.add_test(
   "broken_config",
   type: "failure",
-  message: /Error: ESLint configuration in .eslintrc.yml is invalid:/,
+  message: "The analysis failed due to an unexpected error. See the analysis log for details.",
```
ybiquitous, line 158: [note] In the ESLint cases, the failure messages do not become user-friendly. This is a trade-off between usability and development costs.

If we would make the messages kinder, the following code would be necessary:

```ruby
if stderr.match?(/Error: ESLint configuration in .+ is invalid:/) || stderr.match?(/ESLint couldn't find the plugin ".+"/)
  message = stderr
end
```

Such a code is flaky and may be changed between the different versions of ESLint.

yykamei (approved):

test/smokes/eslint/expectations.rb
yykamei: If the message generated by ESLint is user-friendly, how about remaining here? `Results::Failure` still looks to receive the `:message` argument.

https://github.com/sider/runners/pull/1459/files#diff-56fa6ae5f7dd943b77508242a04a3b61L192

ybiquitous:
Thanks for the review. 😄

test/smokes/eslint/expectations.rb
ybiquitous: > If the message generated by ESLint is user-friendly,

We cannot know whether the message is really user-friendly or not from `stderr`. 😓 
I think the following logic is necessary in order to extract a user-friendly message from `stderr`:

```
if stderr.match?(/Error: ESLint configuration in .+ is invalid:/) || stderr.match?(/ESLint couldn't find the plugin ".+"/)
  message = stderr
end
```

yykamei: OK, I think this PR is good. The kindness of the messages depends on ESLint and ESLint plugins. This is uncontrollable for us. Let's use DEFAULT_MESSAGE.

Issue #1458, opened 2020-09-10
[Cppcheck] Process aborted or coredumped: #<Process::Status: pid 31 SIGUSR2 (signal 12)>


suetsuguk:
According to the log the execution time of the cppcheck process is too long and reaches  the timeout of 30m.

```
...
1795.45s $ cppcheck ...
0.81s    <?xml version="1.0" encoding="UTF-8"?>
0.00s    Process aborted or coredumped: #<Process::Status: pid 31 SIGUSR2 (signal 12)>
...
```

The user has 6190 files in the repository. After reducing target files to 200 with the `ignore` option, the same error still occurs.

suetsuguk:
After discussed with @ybiquitous  and @yykamei , we concluded that we should apply following options to avoid timeout:

- increase the number of CPUs on Docker
- run num-of-CPUs threads of cppcheck #1462
- increase the timeout value #1461

ybiquitous:
@suetsuguk How is your progress?

suetsuguk:
This issue can be closed when the related issue https://github.com/sider/gitlab_sideci/issues/502 is verified.

PR #1462, opened 2020-09-10
[Cppcheck] Run multiple threads with the `-j` option

suetsuguk:
Use the multi-threading `-j` option to improve analysis time.

Related to #1458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,13 @@ All notable changes to this project will be documented in this file.
 
 [Full diff](https://github.com/sider/runners/compare/0.34.1...HEAD)
 
+- **Cppcheck** Run multiple threads with the `-j` option [#1462](https://github.com/sider/runners/pull/1462)
+
 ## 0.34.1
 
 [Full diff](https://github.com/sider/runners/compare/0.34.0...0.34.1)
 
 - Reconsider `Results::Failure` message [#1459](https://github.com/sider/runners/pull/1459)
-- **Cppcheck** Run multiple threads with the `-j` option [#1462](https://github.com/sider/runners/pull/1462)
 
 ## 0.34.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 [Full diff](https://github.com/sider/runners/compare/0.34.0...HEAD)
 
 - Reconsider `Results::Failure` message [#1459](https://github.com/sider/runners/pull/1459)
+- **Cppcheck** Run multiple threads with the `-j` option [#1462](https://github.com/sider/runners/pull/1462)
 
 ## 0.34.0
 
--- a/lib/runners/processor/cppcheck.rb
+++ b/lib/runners/processor/cppcheck.rb
@@ -18,7 +18,7 @@ module Runners
           project: string?,
           language: string?,
           'bug-hunting': boolean?,
-          parallel: boolean?
+          parallel: boolean?,
         )
       }
 
@@ -105,7 +105,6 @@ module Runners
     def step_analyzer(*args)
       stdout, stderr, status = capture3(
         analyzer_bin,
-        *jobs,
         "--quiet",
         "--xml",
         "--output-file=#{report_file}",
@@ -113,6 +112,7 @@ module Runners
         *project,
         *language,
         *config_include_path,
+        *jobs,
         *args,
         *target
       )
--- a/lib/runners/processor/cppcheck.rb
+++ b/lib/runners/processor/cppcheck.rb
@@ -1,3 +1,5 @@
+require "etc"
+
 module Runners
   class Processor::Cppcheck < Processor
     include CPlusPlus
@@ -98,6 +100,7 @@ module Runners
     def step_analyzer(*args)
       stdout, stderr, status = capture3(
         analyzer_bin,
+        "-j", Etc.nprocessors.to_s,
         "--quiet",
         "--xml",
         "--output-file=#{report_file}",
--- a/lib/runners/processor/cppcheck.rb
+++ b/lib/runners/processor/cppcheck.rb
@@ -71,7 +71,20 @@ module Runners
     end
 
     def jobs
-      (config_linter[:parallel] && !config_linter[:project]) ? ["-j", Etc.nprocessors.to_s] : []
+      @jobs ||=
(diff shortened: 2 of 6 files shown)


suetsuguk:
According to the changes 03ed332b2d74b9efdb011725ab7545c6f3316fad in the smoke test , I found out there are several limitations for the `-j` option:

- [`unusedFunction` analysis disabled](https://github.com/danmar/cppcheck/blob/main/cli/cmdlineparser.cpp#L921)
- [`CTU(Whole program analysis)` analysis doesn't work (probably)](http://cppcheck.sourceforge.net/devinfo/doxyoutput/group__Core.html)
- the `project` option doesn't work (probably)

The `unusedFunction` and `CTU` are not big problems for our use.

suetsuguk:
As for the `project` option, we need to consider how to handle it. Seen from the changes at 03ed332b2d74b9efdb011725ab7545c6f3316fad , the cppcheck with `-j` analyzes a file which is not defined in a project file. This implies the `-j` option is regardless of the project file.

If we give priority the project file over multi-threading, we could omit the `-j` option when the `project` is specified:

```ruby
    def jobs
      ["-j", Etc.nprocessors.to_s] unless config_linter[:project]
    end
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 [Full diff](https://github.com/sider/runners/compare/0.34.0...HEAD)
 
 - Reconsider `Results::Failure` message [#1459](https://github.com/sider/runners/pull/1459)
+- **Cppcheck** Run multiple threads with the `-j` option [#1462](https://github.com/sider/runners/pull/1462)
```
ybiquitous, line 10: [suggestion] Please include the latest commit on the `master` branch. 🙏  

https://github.com/sider/runners/blob/5e9edc19e917166f13eb1ea2892a812ddf425903/CHANGELOG.md

ybiquitous:
Thank you for the investigation.

> The `unusedFunction` and `CTU` are not big problems for our use.

Umm..., perhaps there may be users who want to enable these rules.
So, what about providing a new option `parallel` to `sider.yml` as follows?

```yaml
linter:
  cppcheck:
    parallel: true # defaults to `false`
```

Also, we may need to add a note to describe `unusedFunction` and `CTU` rules are disabled when `parallel: true`.

CHANGELOG.md
suetsuguk: Fixed. commit 0506bfd197d544e305c1ebf1027ab1923ae0b4c9

suetsuguk:
The `parallel` option is added. Please review again.

lib/runners/processor/cppcheck.rb
```diff
@@ -15,7 +17,8 @@ class NoTargetFiles < SystemError; end
           std: string?,
           project: string?,
           language: string?,
-          'bug-hunting': boolean?
+          'bug-hunting': boolean?,
+          parallel: boolean?
```
ybiquitous, line 21: [nitpick] How about reducing diffs in the future due to a trailing comma?

```suggestion
          parallel: boolean?,
```

lib/runners/processor/cppcheck.rb
```diff
@@ -98,6 +105,7 @@ def run_analyzer
     def step_analyzer(*args)
       stdout, stderr, status = capture3(
         analyzer_bin,
+        *jobs,
```
ybiquitous, line 108: [suggestion] Can you please move `*jobs,` after `*config_include_path,`?

test/smokes/cppcheck/enable/sider.yml
```diff
@@ -1,3 +1,4 @@
 linter:
   cppcheck:
     enable: "warning,unusedFunction"
+    parallel: false
```
ybiquitous, line 4: [suggestion] How about reverting the file to test the default value `parallel: false`?

```suggestion
```

lib/runners/processor/cppcheck.rb
```diff
@@ -67,6 +70,10 @@ def language
       config_linter[:language].then { |lang| lang ? ["--language=#{lang}"] : [] }
     end
 
+    def jobs
+      (config_linter[:parallel] && !config_linter[:project]) ? ["-j", Etc.nprocessors.to_s] : []
```
ybiquitous, line 74: [suggestion] How about adding a warning as follows?
(Can we give a good reason for the limitation to users?)

```suggestion
      @jobs ||=
        if config_linter[:parallel]
          if config_linter[:project]
            add_warning <<~MSG, file: config.path_name
              The `parallel` option is ignored when the `project` option is specified.
              This limitation is due to the behavior of #{analyzer_name}.
              See also <https://...>
            MSG
            []
          else
            ["-j", Etc.nprocessors.to_s]
          end
        else
          []
        end
      end
```

test/smokes/cppcheck/enable/sider.yml
suetsuguk: This is an intended change because a test for the default value is covered by the other test "default".
The "default" test reports several 'CTU' issues as expected.

lib/runners/processor/cppcheck.rb
suetsuguk: I tried to search the Web but could not find any notes, reports, or issues about this limitation. Shall we omit the "See also" line as of now?

ybiquitous: Yes, of course. 👍

test/smokes/cppcheck/enable/sider.yml
ybiquitous: Make sense. 👍

lib/runners/processor/cppcheck.rb
suetsuguk: Fixed. commit ce5fc1446e28541f76c265df2c5f123d6a20efc4

lib/runners/processor/cppcheck.rb
suetsuguk: Fixed. commit ce5fc1446e28541f76c265df2c5f123d6a20efc4

lib/runners/processor/cppcheck.rb
suetsuguk: Fixed. commit 5210f88bb69a19b514f3386dd96122a108149622

ybiquitous (approved):
LGTM! 👏

PR #1463, opened 2020-09-11
Replace ENTRYPOINT with docker-entrypoint.sh

### LINKED ISSUE #1461: Let users pass DURATION of timeout(1) at runtime
Runners limit the analysis duration with timeout(1) here.

https://github.com/sider/runners/blob/7754062b89fb3d8e286a04cb77aa6918cb424cf8/images/Dockerfile.end.erb#L11

At a glance, ` ENV.fetch('RUNNERS_TIMEOUT', '30m')` looks configurable with the environment variable `RUNNERS_TIMEOUT`, but this variable is used only in testing. I want to invoke Runners with an arbitrary duration at runtime.

I'm planning to add a script for `ENTRYPOINT` and make it parse arguments via `CMD` or read environment variables.

yykamei:
The background for the change is that I want to pass DURATION of
timeout(1) to Runners. Runners would be more flexible at runtime to
consider dynamic parameters with `docker-entrypoint.sh`.

You can pass the parameter via the command line option `--timeout` or
the environment variable `RUNNERT_TIMEOUT` like this:

```console
docker run sider/runner_shellcheck test-guid --timeout=10m
docker run --env RUNNERS_TIMEOUT=5m sider/runner_eslint test-guid
```

Close #1461


images/Dockerfile.end.erb
```diff
@@ -8,4 +8,5 @@ ENV PATH ${RUNNERS_DIR}/bin:${PATH}
 USER $RUNNER_USER
 WORKDIR $RUNNER_USER_HOME
 
-ENTRYPOINT ["/usr/bin/timeout", "--signal=SIGUSR2", "--kill-after=10s", "<%= ENV.fetch('RUNNERS_TIMEOUT', '30m') %>", "runners", "--analyzer=<%= ENV.fetch('ANALYZER') %>"]
+COPY images/docker-entrypoint.sh /
```
yykamei, line 11: > __[Goodcheck]__ How about adding `--chown` to this `COPY` instruction? ([view](https://sider.review/gh/repos/202486306/pulls/1463#issue-4573286))
>
> <details>
>
> The owner of files copied by the `COPY` instruction is the *root* user.
> In the most cases, this behavior may be unexpected.
>
> See https://docs.docker.com/engine/reference/builder/#copy
>
> | Rule |
> | ---- |
> | `sider.runners.docker.copy-require-chown` |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1463#issue-4573286) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

I think it's OK to set the owner of the file `root`.

```console
$ docker run -it --rm --entrypoint stat sider/runner_rubocop:dev /docker-entrypoint.sh
  File: /docker-entrypoint.sh
  Size: 514             Blocks: 8          IO Block: 4096   regular file
Device: 5ch/92d Inode: 840233862   Links: 1
Access: (0755/-rwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2020-09-11 07:55:59.450101802 +0000
Modify: 2020-09-11 07:55:54.326891971 +0000
Change: 2020-09-11 07:55:58.903448267 +0000
 Birth: -
```

ybiquitous (approved):
LGTM 👍

yykamei:
Thank you for merging this PR :pray:

PR #1489, opened 2020-09-17
Run update-ca-certificates(8) before analyzing

yykamei:
From this change, Runners supports the environment variable `EXTRA_CERTIFICATE`, and it adds the certificate when the variable is set. Then, the specified certificate will be registered by calling update-ca-certificates(8).


lib/runners/testing/smoke.rb
```diff
@@ -224,6 +225,26 @@ def colored_pretty_inspect(obj, multiline: false)
         obj.awesome_inspect(indent: 2, index: false, multiline: multiline)
       end
 
+      def extra_certificate
+        key = OpenSSL::PKey::RSA.new(4096)
```
ybiquitous, line 229: [suggestion] It may be safer to add `require "openssl"`.

ybiquitous (approved):
LGTM :+1:

lib/runners/testing/smoke.rb
yykamei: Thank you. It makes sense! 3e2c6e5e

PR #1490, opened 2020-09-18
Remove the execution of adding /etc/sudoers.d/runner

yykamei:
Thanks to https://github.com/sider/devon_rex/pull/272, we don't have to add `/etc/sudoers.d/runner`.

> - [ ] Add a new [changelog](https://github.com/sider/runners/blob/master/CHANGELOG.md) entry if this change is notable.

I believe this is not worth explaining in CHANGELOG because it does not affect users' configuration.


suetsuguk (approved):
LGTM. 👍

PR #1599, opened 2020-10-26
Split test from build workflow on CI

ybiquitous:

This change aims to allow us to retry only the "Test" workflow.
(The "Build" workflow is slow because of including smoke tests)

None.


yykamei (approved):
:shipit:

PR #1622, opened 2020-11-08
Refactor `Runners::Testing::Smoke.build_pattern` method

ybiquitous:

This refactoring aims to just improve the code readability a bit.

None.


lib/runners/testing/smoke.rb
```diff
@@ -298,15 +298,13 @@ def self.build_pattern(type:,
...
-              inspect: inspect,
-            }.compact),
-          },
+            issues: issues,
+            message: message,
+            analyzer: analyzer,
+            class: binding.local_variable_get(:class),
+            backtrace: backtrace,
+            inspect: inspect,
+          }.compact,
```
ybiquitous, line 307: [note] `type`, `guid`, and `timestamp` should not be `nil`.

PR #1632, opened 2020-11-10
Reduce log for git-blame(1)

ybiquitous:

When too many issues are found, git-blame(1) is called per issue.
This can cause too many log entries.

None.

> Check the following items.

### DIFF
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -4,14 +4,14 @@ module Runners
     class CheckoutFailed < SystemError; end
     class BlameFailed < SystemError; end
 
-    def range_git_blame_info(path_string, start_line, end_line)
+    def range_git_blame_info(path_string, start_line, end_line, trace: false)
       @git_blame_cache ||= {}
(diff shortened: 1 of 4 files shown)


lib/runners/workspace/git.rb
```diff
@@ -11,7 +11,7 @@ def range_git_blame_info(path_string, start_line, end_line)
       return cache_value if cache_value
 
       stdout, _ = shell.capture3!("git", "blame", "-p", "-L", "#{start_line},#{end_line}", git_source.head, "--", path_string,
-                                  trace_stdout: false, trace_stderr: true)
+                                  trace_stdout: false, trace_stderr: false, trace_command_line: false)
```
ybiquitous, line 14: [note] Since we use `capture3!`, I think it needless to enable `trace_stderr`.
If some error occurs, an exception should be raised. We can get details from the exception.

yykamei (approved):
:+1:

ybiquitous:
Thanks!

PR #1635, opened 2020-11-11
[PHPCS] Use XML reporter instead of JSON

ybiquitous:

Rarely, JSON strings that PHP_CodeSniffer outputs are broken.
I think the reason is due to the JSON reporter's implementation.
It is by string concatenation, not using a library.

So, this change uses the XML reporter instead of the JSON one.

See also:
- https://github.com/squizlabs/PHP_CodeSniffer/wiki/Reporting#printing-an-xml-report
- https://github.com/squizlabs/PHP_CodeSniffer/blob/3.5.8/src/Reports/Xml.php
- https://github.com/squizlabs/PHP_CodeSniffer/blob/3.5.8/src/Reports/Json.php

None.

> Check the following items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 
 - Optimize `Runners::Changes#include?` [#1633](https://github.com/sider/runners/pull/1633)
 - Reduce log for git-blame(1) [#1632](https://github.com/sider/runners/pull/1632)
+- **PHP_CodeSniffer** Use XML reporter instead of JSON [#1635](https://github.com/sider/runners/pull/1635)
 
 ## 0.37.1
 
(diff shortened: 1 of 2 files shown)


yykamei (approved):
Thank you. LGTM

gnakagaw (approved):
Sorry for my late reviewing. LGTM.

ybiquitous:
@gnakagaw Do you have a plan to send a patch fixing the JSON reporter to PHPCS?
Is it still safer to use the XML reporter instead of JSON?

gnakagaw:
@ybiquitous 
> Do you have a plan to send a patch fixing the JSON reporter to PHPCS?
Yes. I'll do it.

> Is it still safer to use the XML reporter instead of JSON?
It is safe. But it is not suitable. Users encounter garbled characters (mozi-bake) in Sider's issue messages even if we use XML reporter.

ybiquitous:
OK. Let's make this PR pending.

ybiquitous:
We no longer need this PR, so close it.

Issue #877, opened 2020-03-18
Improve typing coverage for new Steep


ybiquitous:
We should progress the migration of [Steep](https://github.com/soutaro/steep) from old to new gradually (see PR #866 for details), and improve the typing coverage.

Please close this issue when we will complete the migration.

## See also

https://github.com/sider/runners/blob/eb5daa542f746112f5514937158d42314b03398d/Gemfile.steep#L3

https://github.com/sider/runners/blob/eb5daa542f746112f5514937158d42314b03398d/Steepfile#L2-L3

https://github.com/sider/runners/blob/eb5daa542f746112f5514937158d42314b03398d/.github/workflows/build.yml#L75-L91

PR #1649, opened 2020-11-16
Fix RBS partially and add soutaro/gem_rbs

ybiquitous:

This adds the Git submodule: https://github.com/soutaro/gem_rbs
By this addition, Retryable's polyfill is needless.

A part of #877


yykamei (approved):
LGTM

Would you update README for type-checking? Maybe, it's helpful to tell we're using submodules.

ybiquitous:
Good catch! I'll open a new follow-up PR. 💪

Issue #1672, opened 2020-11-25
Extract the `issue` field from a `result` schema and write issues per each line


yykamei:
Recently, our service has got hit the problem of too many issues with Runners—the Runners' result was so large that our service failed to parse it and got killed because of out of memory. I think we don't have to use and show such a big result to our customers, but metadata, such as the total number of issues, is necessary.

I looked into the problem and found out that out of memory happened when parsing each item of JSON sequence. If it's OK for us to use only the total number of issues, I would say we can go on without parsing. However, we want to extract the metadata of Runners' result in this case.

~That's why I want to add a new schema for Trace to show metadata of a result.~

~For example, if the last line contains metadata such as:~

```json
{"result_metadata":{"total_issues":18}}
```

~we can skip the previous lines by consuming the IO stream, and it's sufficient to parse only the last line.~

---

EDIT: We decided to extract `issues` from a `result` schema.

The output by Runners with the new schema is like this:

```json
{"result": {"type": "success", "timestamp": "..."}}
{"issues":{"position":"begin","length":2000}}
{"issue":{...}}
{"issue":{...}}
{"issue":{...}}
{"issue":{...}}
{"issues":{"position":"end","length":2000}}
```

In order to easily get back to the previous format, I want to introduce the new parameter (maybe, an environment variable) `NEW_ISSUE_SCHEMA`.

ybiquitous:
Another solution:

What we want is to know the number of issues before evaluating an array of issues.
So, the following idea seems considerable to me:

- Add `issues_count` to `result`.
- Remove `issues` from `result`.
- Add `result_issues` next to `result`.

```
{"result": { ..., "issues_count": 100 }}
{"result_issues": [...]}
```

---

An example of the current trace:

https://sider.review/gh/repos/202486306/pulls/1619/logs/fa934ddf-e6e5-46bf-95cf-7b0e50ab0f46/raw

<img width="660" alt="image" src="https://user-images.githubusercontent.com/473530/100297971-3b094400-2fd3-11eb-8aaa-97819d020c58.png">

yykamei:
Is it acceptable to change the existing schema? If so, I'll follow your suggestion.

ybiquitous:
I think it's acceptable because we have ever added fields like #1363.

yykamei:
OK. Thanks.

This time, we are going to remove the field `issues` from the `result`. Is it still acceptable? I think users of the Runners must change the means of extracting issues.

[...]

PR #1674, opened 2020-11-26
Add the new option `new_issue_schema`

yykamei:

We considered changing the result schema here, but the change could have
a big impact on developers.

https://github.com/sider/runners/issues/1672

So, before implementing the above issue, I want to introduce the new
option `new_issue_schema` to let developers choose a schema.

This is a part of #1672 

> Check the following items.


ybiquitous (approved):
👍

yykamei:
Thank you so much.

PR #1677, opened 2020-11-27
Write issues per each line when `new_issue_schema` is `true`

### LINKED ISSUE #1672: Extract the `issue` field from a `result` schema and write issues per each line

yykamei:

This change makes Runners write issues per each line with `new_issue_schema`. This will reduce the size of the `result`'s line but make the log longer and bigger.

I tried to remain the class `Runners::Results::Base` and its descendants as possible as I could, and `#as_json` has been changed to receive the `with_issues` argument. In order to avoid updating each `Processor`, I wanted Runners to control the presence of `issues` in the `result` schema at generating JSON.

Close #1672

> Check the following items.

### DIFF
--- a/lib/runners/schema/result.rb
+++ b/lib/runners/schema/result.rb
@@ -19,7 +19,8 @@ module Runners
       let :warning, object(message: string, file: string?)
       let :analyzer, object(name: string, version: string)
 
-      let :success, object(guid: string, timestamp: string, type: literal("success"), issues: array(Issue.issue), analyzer: analyzer)
+      let :success, enum(object(guid: string, timestamp: string, type: literal("success"), issues: array(Issue.issue), analyzer: analyzer),
+                         object(guid: string, timestamp: string, type: literal("success"), analyzer: analyzer))
       let :failure, object(guid: string, timestamp: string, type: literal("failure"), message: string, analyzer: optional(analyzer))
       let :error, object(guid: string, timestamp: string, type: literal("error"), class: string, backtrace: array(string), inspect: string, analyzer: optional(analyzer))
 
--- a/test/cli_test.rb
+++ b/test/cli_test.rb
@@ -104,6 +104,27 @@ class CLITest < Minitest::Test
     assert objects.any? { |hash| hash[:ci_config] == nil }
   end
 
+  def test_run_with_new_issue_schema
+    json = options_json({ new_issue_schema: true, source: new_source })
+    cli = CLI.new(argv: ["--analyzer=rubocop", "test-guid"], stdout: stdout, stderr: stderr, options_json: json)
+    cli.instance_variable_set(:@processor_class, TestProcessor)
+    cli.run
+
+    # It write JSON objects to stdout
+
+    output = stdout.string
+    reader = JSONSEQ::Reader.new(io: StringIO.new(output), decoder: -> (string) { JSON.parse(string, symbolize_names: true) })
+    objects = reader.each_object.to_a
+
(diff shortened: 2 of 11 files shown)


lib/runners/schema/result.rb
```diff
@@ -19,7 +19,8 @@ module Schema
       let :warning, object(message: string, file: string?)
       let :analyzer, object(name: string, version: string)
 
-      let :success, object(guid: string, timestamp: string, type: literal("success"), issues: array(Issue.issue), analyzer: analyzer)
+      let :success, enum(object(guid: string, timestamp: string, type: literal("success"), issues: array(Issue.issue), analyzer: analyzer),
+                         object(guid: string, timestamp: string, type: literal("success"), analyzer: analyzer))
```
ybiquitous, line 23: [question] Cannot we simply make `issues` optional?

```suggestion
      let :success, object(guid: string, timestamp: string, type: literal("success"), issues: array?(Issue.issue), analyzer: analyzer)
```

yykamei: Of course, yes. I chose `enum` to completely remove the `issues` from the JSON.

ybiquitous: OK, let's keep it as-is. 👍

test/cli_test.rb
```diff
@@ -104,6 +104,27 @@ def test_run_when_sider_yml_is_broken
...
+    output = stdout.string
+    reader = JSONSEQ::Reader.new(io: StringIO.new(output), decoder: -> (string) { JSON.parse(string, symbolize_names: true) })
+    objects = reader.each_object.to_a
+
+    assert objects.any? { |hash| hash[:trace] == 'command_line' && hash[:command_line] == ["test", "command"] }
+    assert objects.any? { |hash| hash[:trace] == 'status' && hash[:status] == 31 }
+    assert objects.any? { |hash| hash[:trace] == 'warning' && hash[:message] == 'hogehogewarn' }
+    assert objects.any? { |hash| hash[:warnings] == [{ message: 'hogehogewarn', file: nil }] }
+    assert objects.any? { |hash| hash[:issues] == { position: 'begin', length: 0 } }
+    assert objects.any? { |hash| hash[:issues] == { position: 'end', length: 0 } }
```
ybiquitous, line 124: [suggestion] How about testing a case that there are some issues?

yykamei:
We decided to keep the schema.

PR #1687, opened 2020-11-30
[TSLint] Stop updating deps by dependabot

ybiquitous:

TSLint is deprecated. We don't need to update the dependencies.

Related to #1663


yykamei (approved):
It makes sense.

PR #1716, opened 2020-12-09
Fix RBS for `Runners::IO` classes

ybiquitous:

This change aims to improve the coverage of type-checking via RBS and Steep.

This is a part of #877.


ybiquitous:
CI passes!

PR #1722, opened 2020-12-10
Improve RBS for `Runners::Options`

ybiquitous:

This change aims to improve the RBS and type-check for the `Runners::Options` class.

A part of #877

### DIFF
--- a/lib/runners/schema/options.rb
+++ b/lib/runners/schema/options.rb
@@ -3,7 +3,7 @@ module Runners
     Options = _ = StrongJSON.new do
       # @type self: Types::Options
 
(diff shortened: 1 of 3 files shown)


lib/runners/schema/options.rb
```diff
@@ -3,7 +3,7 @@ module Schema
     Options = _ = StrongJSON.new do
       # @type self: Types::Options
 
-      let :source, object?(
+      let :source, object(
```
ybiquitous, line 6: [note] `source` is actually required because `GitSource.new(**nil)` raises a `TypeError`.

PR #1724, opened 2020-12-10
Fix RBS for `Runners::Ruby::*` modules/classes

ybiquitous:

This change aims to improve the RBS and type-check for the modules and classes under `Runners::Ruby::*`.

A part of #877

### DIFF
--- a/lib/runners/ruby/gem_installer/source.rb
+++ b/lib/runners/ruby/gem_installer/source.rb
@@ -8,27 +8,15 @@ module Runners
         Rubygems.new(DEFAULT)
       end
 
-      def self.rubygems(source)
-        Rubygems.new(source)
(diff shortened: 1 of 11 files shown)


lib/runners/ruby/gem_installer/source.rb
```diff
@@ -8,27 +8,15 @@ def self.default
         Rubygems.new(DEFAULT)
       end
 
-      def self.rubygems(source)
-        Rubygems.new(source)
-      end
-
-      def self.git(repo, **options)
-        Git.new(repo, **options)
-      end
```
ybiquitous, line 17: [note] Used only in the test.

ybiquitous:
CI passed!

PR #1729, opened 2020-12-11
Fix RBS for `Runners::Processor::Brakeman`

ybiquitous:

This change adds a more type-safe class `Runners::Command` than `Array[String]`,
It also adds `Runners::Ruby#ruby_analyzer_command` and deprecates `Runners::Ruby#ruby_analyzer_bin`.

A part of #877


ybiquitous:
CI passes!

PR #1732, opened 2020-12-11
Stop overriding `Runners::Ruby#default_gem_specs`

ybiquitous:

This change refactoring aims to make type-checking easier.
Thus, this improvement for type-checking leads to a better API and simpler code.
Overriding methods tend to be hard.

See also #877


yykamei (approved):
OK

PR #1753, opened 2020-12-15
Fix RBS for `Runners::Processor::RuboCop`

ybiquitous:

This change aims to improve the RBS and type-check for the `Runners::Processor::RuboCop` class.

A part of #877

### DIFF
--- a/lib/runners/processor/rubocop.rb
+++ b/lib/runners/processor/rubocop.rb
@@ -244,9 +244,6 @@ module Runners
       warnings.each { |msg, file| add_warning(msg, file: file) }
     end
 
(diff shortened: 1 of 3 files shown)


lib/runners/processor/rubocop.rb
```diff
@@ -176,74 +244,6 @@ def check_rubocop_yml_warning(stderr)
       warnings.each { |msg, file| add_warning(msg, file: file) }
     end
 
-    def run_analyzer
```
ybiquitous, line 179: [note] Inline `#run_analyzer`.

PR #1761, opened 2020-12-16
Simplify pull request template

ybiquitous:

There is not much to add to a new Runner.

None.


suetsuguk (approved):
👍 I agree with the background you mentioned.

PR #1773, opened 2020-12-16
Fix RBS for `Runners::Processor::Jshint`

ybiquitous:

This change aims to improve the RBS and type-check for the `Runners::Processor::Jshint` class.

A part of #877

> Check the following.

### DIFF
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -257,16 +257,20 @@ module Runners
 
     class InvalidXML < SystemError; end
 
-    def read_report_xml(file_path = report_file)
-      output = read_report_file(file_path)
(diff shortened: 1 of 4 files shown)


lib/runners/processor.rb
```diff
@@ -257,16 +257,29 @@ def read_report_file(file_path = report_file)
 
     class InvalidXML < SystemError; end
 
-    def read_report_xml(file_path = report_file)
-      output = read_report_file(file_path)
-      root = REXML::Document.new(output).root
+    def read_xml(text)
```
ybiquitous, line 260: [note] Added `Processor#read_xml` to reduce duplication.

PR #1781, opened 2020-12-17
Consider line range in `Runners::Changes#include?`

ybiquitous:

Some tools have a line range (start and end) for an issue.
But the current implementation of `Runners::Changes#include?` checks only a start line.
In the following case, an issue is never reported.

Code:

```ruby
def foo     # L1
  p "Hi"    # L2 <--- changed line, but never reported!
end         # L3
```

Issue: "Too long method with L1-3"

This change aims to fix the bug by checking a line range of an issue.

None.

> Check the following.

### DIFF
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -29,34 +29,33 @@ module Runners
     end
 
     def include?(issue)
-      if patches
-        location = issue.location
+      git_diff_patches = patches
+      if git_diff_patches
         # NOTE: #find_patch_by_file omits just renamed files.
         # @see https://github.com/packsaddle/ruby-git_diff_parser/issues/272
-        patch = find_patch_by_file(issue.path.to_path)
-        if patch && location
-          patch.changed_lines.one? { |line| location.start_line == line.number }
-        elsif patch
-          true
+        changed_lines = changed_lines_by_file(git_diff_patches, issue.path.to_path)
+        location = issue.location
+        if location
+          changed_lines.any? { |line| location.include_line?(line) }
         else
-          false
+          !changed_lines.empty?
         end
       else
         changed_paths.member?(issue.path)
       end
     end
 
-    private def find_patch_by_file(path)
-      git_diff_patches = patches or raise "#patches is required"
-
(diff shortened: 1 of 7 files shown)


lib/runners/changes.rb
```diff
@@ -29,34 +29,33 @@ def delete_unchanged(dir:, except: [], only: [])
...
-        @patches_by_file[path]
+      @changed_lines_cache ||= {}
+      hit = @changed_lines_cache[path]
+      if hit
+        hit
       else
-        found = git_diff_patches.find_patch_by_file(path)
-        @patches_by_file[path] = found
-        found
+        lines = patches.find_patch_by_file(path)&.changed_lines&.map(&:number) || []
```
ybiquitous, line 55: [note] `GitDiffParser::Patch#changed_lines` calculates each time per call and it is necessary for our project. So, this change uses cache for `changed_lines`, not `find_patch_by_file`.

https://github.com/packsaddle/ruby-git_diff_parser/blob/e4d8a975ded35e531fa945a60b18a219a60339e0/lib/git_diff_parser/patch.rb#L57-L78

lib/runners/changes.rb
```diff
@@ -29,34 +29,33 @@ def delete_unchanged(dir:, except: [], only: [])
...
     def include?(issue)
-      if patches
-        location = issue.location
+      git_diff_patches = patches
+      if git_diff_patches
         # NOTE: #find_patch_by_file omits just renamed files.
         # @see https://github.com/packsaddle/ruby-git_diff_parser/issues/272
-        patch = find_patch_by_file(issue.path.to_path)
-        if patch && location
-          patch.changed_lines.one? { |line| location.start_line == line.number }
```
ybiquitous, line 38: [note] [`one?`](https://ruby-doc.org/core-2.7.2/Enumerable.html#method-i-one-3F) checks *exactly once* so it iterates all elements. I think it's a needless calculation. Instead, this change uses [`any?`](https://ruby-doc.org/core-2.7.2/Enumerable.html#method-i-any-3F).

yykamei (approved):
OK

PR #1788, opened 2020-12-17
Fix RBS for `Runners::Processor::RemarkLint`

ybiquitous:

This change aims to improve the RBS and type-check for the `Runners::Processor::RemarkLint` class.

A part of #877

> Check the following.

### DIFF
--- a/lib/runners/processor/remark_lint.rb
+++ b/lib/runners/processor/remark_lint.rb
@@ -2,8 +2,10 @@ module Runners
   class Processor::RemarkLint < Processor
     include Nodejs
 
-    Schema = StrongJSON.new do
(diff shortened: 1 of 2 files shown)


lib/runners/processor/remark_lint.rb
```diff
@@ -46,20 +48,24 @@ def setup
...
 
     def analyze(_changes)
-      run_analyzer
-    end
+      _, stderr, _ = capture3(nodejs_analyzer_bin, *cli_options, *analysis_target)
 
-    private
+      issues, errors = parse_result(stderr)
 
-    def remark_lint_version!(global: false)
```
ybiquitous, line 54: [note] This method is no longer used.

PR #1836, opened 2020-12-22
Rename Rake task files: `.rb` -> `.rake`

ybiquitous:

In general, A Rake task file has an extension `.rake` instead of `.rb`.
See <https://ruby.github.io/rake/>.

Also, `rubocop-rake` checks `.rake` files by default.
See <https://github.com/rubocop-hq/rubocop-rake/blob/v0.5.1/config/default.yml#L5>.

None.

> Check the following.


Rakefile
```diff
@@ -2,12 +2,8 @@ require "rake/testtask"
...
 require "aufgaben/release"
 require "aufgaben/bump/ruby"
-require_relative "tasks/bump/analyzers"
-require_relative "tasks/bump/devon_rex"
-require_relative "tasks/docker/timeout_test"
-require_relative "tasks/rbs/update_gems"
-require_relative "tasks/readme/generate"
-require_relative "tasks/steep/stats"
+
+import(*FileList["tasks/**/*.rake"])
```
ybiquitous, line 6: [note] https://ruby.github.io/rake/Rake/DSL.html#method-i-import

PR #1853, opened 2020-12-23
Manage uri via Bundler

ybiquitous:

- https://rubygems.org/gems/uri
- https://github.com/ruby/uri
- https://github.com/ruby/uri/compare/v0.10.0...v0.10.1

The default version in Ruby 2.7.2 is **0.10.0**.

None.

> Check the following.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'open3'
 gem 'psych'
 gem 'json'
 gem 'fileutils'
+gem 'uri'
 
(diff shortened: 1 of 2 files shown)


ybiquitous:
> You have already activated uri 0.10.0, but your Gemfile requires uri 0.10.1. Since uri is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports uri as a default gem. (Gem::LoadError)

Issue #1865, opened 2020-12-25
Unexpectedly Yarn v2 is used


ybiquitous:
With an analysis, Yarn v2 is unexpectedly used. The following is a part of the log:

```
$ yarn -v
2.4.0
exit status: 0
Installing npm packages...
$ yarn install --ignore-engines --ignore-scripts --no-progress --non-interactive --frozen-lockfile
Unknown Syntax Error: Unsupported option name ("--ignore-scripts").

$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--skip-builds]
exit status: 1
Command failed. Retrying...
```

The current stable version of Yarn is v1, so we need to investigate why Yarn v2 is used.

https://classic.yarnpkg.com/en/

<img width="427" alt="image" src="https://user-images.githubusercontent.com/473530/103112189-07cfd880-4697-11eb-9e5b-3c488e57d2c7.png">

ybiquitous:
The cause is `yarn-path` in `.yarnrc` or `yarnPath` in `.yarnrc.yml`:

- https://classic.yarnpkg.com/en/docs/yarnrc/#toc-yarn-path
- https://yarnpkg.com/configuration/yarnrc#yarnPath

PR #1866, opened 2020-12-25
Add debug code to investigate #1865

ybiquitous:

This change aims to investigate the cause of #1865. We should revert this change before release.

Related to #1865 

> Check the following.


ybiquitous:
I will try it with the canary.

PR #1876, opened 2020-12-25
Use always Yarn v1 even if `.yarnrc` exists

### LINKED ISSUE #1865: Unexpectedly Yarn v2 is used

ybiquitous:

Yarn has a feature to change the Yarn binary at runtime.
(see <https://github.com/sider/runners/issues/1865#issuecomment-751174460> for details)

Due to this feature, our version of Yarn sometimes is not used if `.yarnrc` exists in a user's repository.

~~So, this change saves `.yarnrc` files to another place before running the `yarn` command.~~
So, this change uses the `--no-default-rc` flag. Note that this flag is undocumented but the help text describes it:

```console
$ yarn help | grep no-default-rc
    --no-default-rc                     prevent Yarn from automatically detecting yarnrc and npmrc files
```

By doing this, our version of Yarn will be always used.

Fix #1865

> Check the following.


lib/runners/nodejs.rb
```diff
@@ -166,7 +166,25 @@ def npm_install(option)
...
     end
 
-    # @see https://yarnpkg.com/en/docs/cli/install
+    def run_yarn(arg, *args)
+      # NOTE: Avoid the effects of `.yarnrc` or `.yarnrc.yml`.
+      #
+      # @see https://classic.yarnpkg.com/en/docs/yarnrc/#toc-yarn-path
+      # @see https://yarnpkg.com/configuration/yarnrc#yarnPath
+      mktmpdir do |backup_dir|
+        backup_files = [".yarnrc", ".yarnrc.yml", ".yarnrc.yaml"].map { |file| current_dir / file }
```
ybiquitous, line 175: [note] `.yarnrc.yaml` is not supported now but it may be supported in the future.

yykamei (commented):
:thinking: This PR seems to completely stop using rc files. I think the option, `--no-default-rc`, looks sufficient.

ybiquitous:
Oh, `--no-default-rc` is missing in the Yarn document:
https://classic.yarnpkg.com/en/docs/cli/install

It looks nice. I'll try it.

ybiquitous:
@yykamei `--no-default-rc` works well with c0990c16. I've updated this PR description, too.

yykamei (approved):
Thanks 👍

PR #1878, opened 2020-12-25
Support metrics code clone

### LINKED ISSUE #1827: add Code Duplication Detector runner
Add a runner to detect code duplication with `pmd_cpd`. 

Note: It seems to be identical `pmd_cpd` runner at this point. But we have a plan to replace `pmd_cpd` with our original tool.

suetsuguk:

Fix #1827

> Check the following.


README.md
```diff
@@ -29,6 +29,7 @@ All 36 analyzers are provided as a Docker image:
 | JSHint | [docker](https://hub.docker.com/r/sider/runner_jshint), [source](https://github.com/jshint/jshint), [doc](https://help.sider.review/tools/javascript/jshint) | ✅ |
 | ktlint | [docker](https://hub.docker.com/r/sider/runner_ktlint), [source](https://github.com/pinterest/ktlint), [doc](https://help.sider.review/tools/kotlin/ktlint) | ✅ |
 | LanguageTool | [docker](https://hub.docker.com/r/sider/runner_languagetool), [source](https://github.com/languagetool-org/languagetool), [doc](https://help.sider.review/tools/others/languagetool) | ✅ |
+| Metrics Code Clone | [docker](https://hub.docker.com/r/sider/runner_metrics_codeclone), [source](https://github.com/pmd/pmd), [doc](https://help.sider.review/tools/others/metrics_codeclone) | ✅ |
```
suetsuguk, line 32: > __[remark-lint]__ Link to https://help.sider.review/tools/others/metrics_codeclone is dead ([view](https://sider.review/gh/repos/202486306/pulls/1878#issue-5268446))
>
> <details>
>
> | Rule | Severity |
> | ---- | -------- |
> | `no-dead-urls` | `warn` |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1878#issue-5268446) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

The document needs to be created accordingly.

lib/runners/processor/metrics_codeclone.rb
```diff
@@ -0,0 +1,50 @@
+module Runners
+  class Processor::MetricsCodeClone < Processor
+    include PmdCpdBase
+
+    Schema = _ = StrongJSON.new do
```
ybiquitous, line 5: [question] Does it conflict with `PmdCpdBase::Scema`?

suetsuguk: No, it doesn't. Indeed, I got a runtime error when I omitted the `Schema = _ = ...` block from `metrics_codeclone.rb`.
Local variables defined in a module cannot be accessed from outside of its scope as described [here](https://docs.ruby-lang.org/ja/2.7.0/doc/spec=2fvariables.html#local). If it is defined as a class variable with prefix `@@`, the variable can be accessed from modules which include the module as described [here](https://docs.ruby-lang.org/ja/2.7.0/doc/spec=2fvariables.html#class).

ybiquitous: I could not easily understand which `PmdCpdBase::Schema` or `Processor::MetricsCodeClone::Schema` when I viewed the code `Schema.issue`.
So, it seems better to me to keep only `PmdCpd::Schema.issue` in `processor/pmd_cpd.rb`. 🤔

suetsuguk: It can be done by doing the following: the `PmdCpdBase` module generates an array of `hash` instead of the current array of `Issue` tagged by `PmdCpdBase::Schema.issue`. The `PmdCpdBase::Schema.issue` is no longer needed. In `processor/pmd_cpd.rb`, the array of hash is converted into an array of `Issue` tagged by `PmdCpd::Schema.issue` defined in `processor/pmd_cpd.rb`.

This code structure will have good symmetry for both `PmdCpd` and `MetricsCodeClone` classes. Is this acceptable for you?

gnakagaw (commented):
[suggestion] Can we independence `metrics_code_clone` runner from existing `pmd_cpd`?

I've grasped your intention avoiding making code duplication between existing `pmd_cpd` and new one. However, I advocate that we should implement new cpd runner independent from `pmd_cpd` runner implementation even if it makes large code duplication.

Reasons:
1. It makes some extra cost for maintaining `pmd_cpd` runner.
2. We should not modify existing runner code (`pmd_cpd`) for metrics runners. Metrics runner is for a just started project. There is a possibility that we will change the specification drastically. Thus, we avoid to influence the existing runner code.
3. We have a plan to replace `pmd_cpd` with our unique technology. If that happens, the code communization becomes redundant.

But this is my own thought. How do you think?

suetsuguk:
Ironically, a clone detecter is cloning itself 😄 
It is true that we will apply our new technology someday. But we have not planned the schedule yet. Meanwhile we need to handle PMD CPD.
I think more changes in PMD CPD may happen before the new technology is released. In this circumstance, I have come up with the current implementation.
1. Maintenance cost: Changes in PMD CPD will affect only `pmd_cpd_base.rb`
2. Expandability in code: The `pmd_cpd_base.rb` module generates primitive data, and the `metrics_codeclone.rb` sums up them. This structure will cover changes in specification to some extent.
3. Code changes: The only change for `pmd_cpd.rb` is to divide the file into two files. It is easy to revert the changes when the new technology.

lib/runners/processor/metrics_codeclone.rb
```diff
@@ -0,0 +1,50 @@
...
+        clones: integer,
+      )
+    end
+
+    PmdCpdBase.register_config_schema(name: :metrics_codeclone)
+
+    def analyze(changes)
+      issues = run_analyze(changes)
+
+      file_issues =
```
gnakagaw, line 18: [suggestion] Can we use `inject` to aggregate the issues? It makes the code more simple.

```ruby
file_issues = 
  issues
    .inject(Hash.new(0)) { |arr, issue|
      arr.tap { |x| x[issue.path] += 1 }
     }.map { |issue_per_file|
       Issue.new(path: issue_per_file[0],
         ...
         object: {
           clones: issue_per_file[1]
         }
         ...
```

ybiquitous:
[suggestion] I also think it not good to change the existing `Processor::PmdCpd` class much.
Instead, as the following code, how about using the delegation?

```ruby
module Runners
  class Processor::MetricsCodeClone < Processor
    def initialize(...)
      super(...)
      @pmd_cpd = Processor::PmdCpd.new(...)
    end

    def analyze(...)
      result = @pmd_cpd.analyze(...)
      return result unless result.is_a? Results::Success

      file_issues =
        result.issues
          .map { |issue| issue.path }
          .uniq
          .map { |filepath| construct_file_issue(issues, filepath) }

      Results::Success.new(guid: guid, analyzer: analyzer, issues: file_issues)
    end
  end
end
```

lib/runners/processor/metrics_codeclone.rb
ybiquitous: [IMO] Honestly, the current code is good to me (using `inject` seems more complex to me).

gnakagaw: Surely, it looks like more complex code for me after I slept on it. My suggestion may be redundant..  I'll leave it up to @suetsuguk.

suetsuguk: OK. I will keep the current code. Thank you.

suetsuguk:
Good catch! I will change the code with using delegation, and see how it works.

suetsuguk:
@ybiquitous 
We are almost there, but I am stacked.

I tried to run the following command, but failed.
`$ bundle exec rake docker:build docker:smoke ANALYZER=metrics_codeclone DEBUG=true ONLY=warnings`

As seen from the log, I notice remarkable messages as:
- `pmd_cpd` exits with a java.io.FileNotFoundException error. The stderr says "java.io.FileNotFoundException: Couldn't find directory/file 'foo'".
- The `dir` value reported by `Runners::Shell::ExecError` is `dir=#<Pathname:/tmp/d20210106-925-121ui4b>`

The `root_dir` and `files` values are specified in `sider.yml` of the test named `warnings` as;
https://github.com/sider/runners/blob/8ef46e344adf8e7bb215b3ecf8fa62bd55d90f0e/test/smokes/metrics_codeclone/warnings/sider.yml#L3-L4

It seems that the `Shell(shell.rb)` instance is not in the `/tmp/.../src` directory even though `roor_dir` is set to `src`.
I went through the code, but I could not find the root cause of the problem. Can you give me a hand?

ybiquitous:
@suetsuguk As you commented, the cause seems that `Processor#@shell` is not shared. I'll look into it more, so please wait. 🙏 

https://github.com/sider/runners/blob/47652b77b8ae018361b6b25c141ca64b4d4636b9/lib/runners/processor.rb#L44-L46

suetsuguk:
@ybiquitous @gnakagaw Thanks to #1907, we have achieved the `MetricsCodeClone` class without any changes in `pmd_cpd.rb`. Please review.

lib/runners/processor/metrics_codeclone.rb
```diff
@@ -0,0 +1,83 @@
...
+      @pmd_cpd.show_runtime_versions
+    end
+
+    def setup(&block)
+      @pmd_cpd.setup(&block)
+    end
+
+    def use_git_metadata_dir?
+      @pmd_cpd.use_git_metadata_dir?
+    end
```
ybiquitous, line 43: [suggestion] If using `Forwardable`, we can reduce the boilerplate code. 👍 

https://ruby-doc.org/stdlib-2.7.2/libdoc/forwardable/rdoc/Forwardable.html

```ruby
def_delegators @pmd_cpd, :config_linter, :analyzer_bin, ...
```

sig/runners/processor/metrics_codeclone.rbs
```diff
@@ -0,0 +1,15 @@
+module Runners
+  class Processor::MetricsCodeClone < Processor
+
+    class SchemaClass < StrongJSON
+      def issue: () -> StrongJSON::Type::Object[untyped]
+    end
+    Schema: SchemaClass
+
+    def initialize: (**untyped) -> void
```
ybiquitous, line 9: [suggestion] This line may be needless?

```suggestion
```

lib/runners/processor/metrics_codeclone.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    private
+
+    def construct_file_issue(issues, filepath)
+      issues_in_file = issues.select { |issue| issue.path == filepath }
+      clones = issues_in_file.length
+      sum_of_lines = issues_in_file.inject(0) do |sum, issue|
+        issue_obj = issue.object or raise "Required object: #{issue.inspect}"
+        sum + issue_obj[:lines]
+      end
```
ybiquitous, line 68: [suggestion] `Array#sum` with a block may be able to make the code clear.

https://ruby-doc.org/core-2.7.2/Array.html#sum-method

```suggestion
      sum_of_lines = issues_in_file.sum do |issue|
        issue_obj = issue.object or raise "Required object: #{issue.inspect}"
        issue_obj[:lines]
      end
```

lib/runners/processor/metrics_codeclone.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+
+    private
+
+    def construct_file_issue(issues, filepath)
+      issues_in_file = issues.select { |issue| issue.path == filepath }
+      clones = issues_in_file.length
+      sum_of_lines = issues_in_file.inject(0) do |sum, issue|
+        issue_obj = issue.object or raise "Required object: #{issue.inspect}"
+        sum + issue_obj[:lines]
```
ybiquitous, line 67: [suggestion] Using `fetch` may be safer:

```suggestion
        sum + issue_obj.fetch(:lines)
```

lib/runners/processor/metrics_codeclone.rb
```diff
@@ -0,0 +1,83 @@
...
+      Issue.new(
+        path: filepath,
+        location: nil,
+        id: "metrics_code-clones",
+        message: msg,
+        object: {
+          clones: clones,
+          },
+        schema: Schema.issue,
+        )
```
ybiquitous, line 80: [nitpick] Incorrect indent.

```suggestion
        },
        schema: Schema.issue,
      )
```

images/metrics_codeclone/build.gradle
```diff
@@ -0,0 +1,11 @@
...
+    id "java"
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation "net.sourceforge.pmd:pmd-dist:6.30.0"
+}
```
ybiquitous, line 11: [suggestion] This file should be the same as `images/pmd_cpd/build.gradle`. Can we remove this file?

https://github.com/sider/runners/blob/d367e7f02881b450c881f0278e988ae96a632a8b/images/pmd_cpd/build.gradle#L10

sig/runners/processor/metrics_codeclone.rbs
suetsuguk: The line is needed. The `steep` check fails when the line is omitted.

```
$ bundle exec steep check
...
lib/runners/processor/metrics_codeclone.rb:15:4: MethodArityMismatch: method=initialize (def initialize(**params))
lib/runners/processor/metrics_codeclone.rb:16:12: UnexpectedSplat: type= (**params)
```

ybiquitous: OK, I got it. 👍

gnakagaw (approved):
LGTM

BTW, delegation is a grate solution! 👍

images/metrics_codeclone/build.gradle
suetsuguk: Seen from a generated Dockerfile, the build.gradle file is copied from `images/metrics_codeclone/build.gradle` to the image. This is done by `Dockerfile.java.erb`.
https://github.com/sider/runners/blob/d367e7f02881b450c881f0278e988ae96a632a8b/images/metrics_codeclone/Dockerfile#L10

Instead, we can use a symbolic file which refers to `../pmd_cpd/build.gradle`. But that would be going too far in this case.

ybiquitous: If there are 2 `build.gradle`, we will receive updates from Dependabot twice.
I think this is noisy for us and there is no reason to have different versions of PMD CPD.

If the symbolic link resolves the double `build.gradle`, it seems nice to me.

lib/runners/processor/metrics_codeclone.rb
suetsuguk: Good catch! Fixed. commit 262815274b506793a34094325f1e810596975790

lib/runners/processor/metrics_codeclone.rb
suetsuguk: Fixed. commit e6f649e8dc57a69445f5074e76d88b8a0333c611

lib/runners/processor/metrics_codeclone.rb
suetsuguk: Fixed. commit af6cbcf73a1fd82f8821bbaa892cf095a7137073

lib/runners/processor/metrics_codeclone.rb
suetsuguk: Fixed according with commit af6cbcf73a1fd82f8821bbaa892cf095a7137073

images/metrics_codeclone/build.gradle
suetsuguk: The symbolic link works. The entry for `/images/metrics_codeclone` in `dependabot.yml` has also been deleted. commit 3f1daf3c3c4474718f89cf1c179fddbc21763f65

README.md
ybiquitous: The URL <https://help.sider.review/tools/metrics/codeclone> may be simple. 🤔

lib/runners/processor/metrics_codeclone.rb
```diff
@@ -0,0 +1,65 @@
...
+      :use_git_metadata_dir?
+
+    def initialize(**params)
+      super(**params)
+      @pmd_cpd = PmdCpd.new(**params)
+    end
+
+    def analyze(changes)
+      result = @pmd_cpd.analyze(changes)
+      @pmd_cpd.warnings.each { |warn| @warnings << warn }
```
ybiquitous, line 29: [suggestion] Can we use the delegation also for the `#warnings` method?

```ruby
def_delegators :@pmd_cpd, :warnings
``

File: lib/runners/processor/metrics_codeclone.rb [hunk 11]
```diff
@@ -0,0 +1,65 @@
...
+      sum_of_lines = issues_in_file.sum do |issue|
+        issue_obj = issue.object or raise "Required object: #{issue.inspect}"
+        issue_obj.fetch(:lines)
+      end
+      msg = "The number of code clones is #{clones} with total #{sum_of_lines} lines."
+
+      Issue.new(
+        path: filepath,
+        location: nil,
+        id: "metrics_code-clones",
```
ybiquitous, line 56: [suggestion] Can we unify the ID?

```suggestion
        id: "metrics_codeclone",
```

lib/runners/processor/metrics_codeclone.rb
suetsuguk: Fixed. commit 181ee0ad16901b2b2a8dbc4053cd242aab681527

lib/runners/processor/metrics_codeclone.rb
suetsuguk: Fixed. commit 251635b3d4016606321113c112e0001f1993896b

README.md
ybiquitous: @suetsuguk Any thoughts?

suetsuguk: Basically, I agree with it. I went through the code for sideci and found that we need to add `:metrics` language category to build the URL. But I cannot know the impact to the code base yet.

ybiquitous: Is it possible that we will add language-specific metrics like Ruby in the future?
If possible, a problem in which category `/ruby/` or `/metrics/` is suitable would occur. 🤔

suetsuguk: OK. I am now working on adding metrics analyzers to Sideci. I will consider this issue accordingly.

ybiquitous (approved):
LGTM 👍

README.md
suetsuguk: Changed the doc path to `tools/metrics/codeclone`.
commits 022eec8b00848544feb83725fbcf2ad395d3165a and 8a3a6d60680d578ba1187c0892d3c402701325f4

PR #1893, opened 2021-01-05
Improve `pip install`

ybiquitous:

- Add method: `Runners::Python#pip_install`.
- Run `pip list` to show installed packages for our debug.
- Set `--disable-pip-version-check` to the `pip` command to reduce noises.
  (see also <https://github.com/sider/devon_rex/issues/359>)

None.

> Check the following.

### DIFF
--- a/lib/runners/python.rb
+++ b/lib/runners/python.rb
@@ -5,5 +5,13 @@ module Runners
       capture3! "pip", "--version"
       capture3! "pipenv", "--version"
     end
+
+    def pip_install(*packages)
+      unless packages.empty?
+        options = %w[--disable-pip-version-check]
+        capture3! "pip", "install", *options, *packages
+        capture3! "pip", "list", *options, *packages
+      end
+    end
   end
 end
--- a/lib/runners/python.rb
+++ b/lib/runners/python.rb
@@ -9,7 +9,7 @@ module Runners
     def pip_install(*packages)
       unless packages.empty?
         options = %w[--disable-pip-version-check]
-        capture3! "pip", "install", *options, *packages
+        capture3! "pip", "install", "--quiet", *options, *packages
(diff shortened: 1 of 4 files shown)


lib/runners/python.rb
```diff
@@ -5,5 +5,13 @@ def show_runtime_versions
       capture3! "pip", "--version"
       capture3! "pipenv", "--version"
     end
+
+    def pip_install(*packages)
+      unless packages.empty?
+        options = %w[--disable-pip-version-check]
```
yykamei, line 11: How about adding `--quiet` because you added a new command-line, `pip list`?

yykamei (approved):
LGTM

lib/runners/python.rb
ybiquitous: Sounds good. Is `--quiet` needed for `pip install`?

ybiquitous: Oh, `pip list --quiet` outputs nothing, right?

gnakagaw (approved):
LGTM 👍

lib/runners/python.rb
yykamei: Yes as long as no errors happen.

ybiquitous: I add `pip list` because it will allow us to debug easily by installed packages.
So, adding `--quiet` seems meaningless.

yykamei: Sorry, `--quiet` should be added only for `pip install`, not for `pip list`.

ybiquitous: Done 739d4137 👍

yykamei: Thank you :pray:

yykamei (approved):

PR #1899, opened 2021-01-06
[Flake8] add debug code

gnakagaw:
This PR adds several debug code for `Flake8` runner to investigate an analysis error. 

* add verbose output
* add timeout (10m) for Flake8 runner.

NOTE: This PR is for a debugging task only. We have to revert this PR before next release.

PR #1900, opened 2021-01-06
[Flake8] remove debug code (#1899, #1903, #1908)

gnakagaw:
This PR revert several debug code in #1899 ,#1903 and #1908.

PR #1903, opened 2021-01-06
[Flake 8] add more debug code

gnakagaw:
add several debug code for an investigation for an issue.

We've already added debug code in #1899. But it is useless in a time-out situation because the runner don't output `result_file` in the situation. This commit force to output debug log to `stderr` trace.


ybiquitous (approved):
LGTM 👍 

Please don't forget to update also PR #1900.

PR #1908, opened 2021-01-07
[Flake8] debug: disable multi-processing

gnakagaw:
This PR is for debugging purpose. We'll revert before the next release. 
Related: #1903 #1899

This PR disables multi-processing for `Flake 8` runner.
The commit adds `-j 1` option for Flake 8.

Here is the description for the option
>-j JOBS, --jobs=JOBS  Number of subprocesses to use to run checks in parallel. This is ignored on Windows. The default, "auto", will auto-detect the number of processors available to use. (Default: auto)

https://flake8.pycqa.org/en/latest/manpage.html#options


ybiquitous (approved):

gnakagaw:
will revert in #1900

PR #1914, opened 2021-01-08
Add TraceWriter.finish

### LINKED ISSUE #1912: Add a new trace record that indicates the finish of an analysis
It would be useful for callers of sider/runners to detect the completion of the analysis from traces. Currently, sider/runners writes a trace with "Elapsed time: ..." at the end of traces. However, this message is just a `:message` record, so it could be changed in the future. I want a rigid schema for **finish**. How about creating a new trace record **finish** like this?

```ruby
let :finish, object(trace: literal(:finish), message: string, duration_in_sec: integer, recorded_at: string)
```

yykamei:
This new method is responsible for indicating the finish of an analysis.
It's basically assumed to be called from `CLI`.

Fix #1912

### DIFF
--- a/lib/runners/cli.rb
+++ b/lib/runners/cli.rb
@@ -63,14 +63,14 @@ module Runners
 
         result.tap do
           finished_at = Time.now
-          trace_writer.header "Analysis finished", recorded_at: finished_at
+          trace_writer.header "Analysis finished"
 
           if result.is_a? Results::Success
             trace_writer.message "#{result.issues.size} issue(s) found."
           end
 
           trace_writer.message "Finished at #{finished_at.utc}"
-          trace_writer.message "Elapsed time: #{format_duration(finished_at - started_at)}"
+          trace_writer.finish "Elapsed time: #{format_duration(finished_at - started_at)}", duration_in_sec: Integer(finished_at - started_at)
         end
       end
     ensure
--- a/lib/runners/cli.rb
+++ b/lib/runners/cli.rb
@@ -70,7 +70,8 @@ module Runners
           end
 
           trace_writer.message "Finished at #{finished_at.utc}"
-          trace_writer.finish "Elapsed time: #{format_duration(finished_at - started_at)}", duration_in_sec: Integer(finished_at - started_at)
+          trace_writer.message "Elapsed time: #{format_duration(finished_at - started_at)}"
+          trace_writer.finish duration_in_sec: Integer(finished_at - started_at)
         end
       end
     ensure
--- a/lib/runners/cli.rb
+++ b/lib/runners/cli.rb
@@ -71,7 +71,7 @@ module Runners
 
           trace_writer.message "Finished at #{finished_at.utc}"
           trace_writer.message "Elapsed time: #{format_duration(finished_at - started_at)}"
-          trace_writer.finish duration_in_sec: Integer(finished_at - started_at), started_at: started_at, finished_at: finished_at
+          trace_writer.finish started_at: started_at, finished_at: finished_at
         end
       end
     ensure
--- a/lib/runners/cli.rb
+++ b/lib/runners/cli.rb
@@ -71,7 +71,7 @@ module Runners
 
           trace_writer.message "Finished at #{finished_at.utc}"
           trace_writer.message "Elapsed time: #{format_duration(finished_at - started_at)}"
-          trace_writer.finish duration_in_sec: Integer(finished_at - started_at)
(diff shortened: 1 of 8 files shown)


lib/runners/cli.rb
```diff
@@ -63,14 +63,14 @@ def run
 
         result.tap do
           finished_at = Time.now
-          trace_writer.header "Analysis finished", recorded_at: finished_at
+          trace_writer.header "Analysis finished"
```
yykamei, line 66: I don't think `recorded_at` should be the same as `finished_at` here.

lib/runners/cli.rb
```diff
@@ -63,14 +63,14 @@ def run
...
 
           if result.is_a? Results::Success
             trace_writer.message "#{result.issues.size} issue(s) found."
           end
 
           trace_writer.message "Finished at #{finished_at.utc}"
-          trace_writer.message "Elapsed time: #{format_duration(finished_at - started_at)}"
+          trace_writer.finish "Elapsed time: #{format_duration(finished_at - started_at)}", duration_in_sec: Integer(finished_at - started_at)
```
ybiquitous, line 73: [suggestion] How about keeping the existing `#message` and adding `#finish` without a message?
If doing so, Runners clients don't have to change the code to show the finish message.

```suggestion
          trace_writer.message "Elapsed time: #{format_duration(finished_at - started_at)}"
          trace_writer.finish started_at: started_at, finished_at: finished_at, duration_in_sec: Integer(finished_at - started_at)
```

yykamei: When it comes to Sider, `finish` without `message` seems to require us to change our code because we just see the key, `"trace"`, to extract traces from the output. Perhaps, we should exclude traces without `message`. Is it OK?

ybiquitous: Sorry, I can't understand the following: 😓 

> Perhaps, we should exclude traces without `message`.

What code do you mean?

yykamei: 16d328ba

lib/runners/cli.rb
```diff
@@ -63,14 +63,15 @@ def run
...
-          trace_writer.header "Analysis finished", recorded_at: finished_at
+          trace_writer.header "Analysis finished"
 
           if result.is_a? Results::Success
             trace_writer.message "#{result.issues.size} issue(s) found."
           end
 
           trace_writer.message "Finished at #{finished_at.utc}"
           trace_writer.message "Elapsed time: #{format_duration(finished_at - started_at)}"
+          trace_writer.finish duration_in_sec: Integer(finished_at - started_at)
```
ybiquitous, line 74: [suggestion] Not only `duration_in_sec`, how about adding `started_at` and `finished_at`? They are unused now, but they may be useful in the future.

ybiquitous: `duration_in_sec` can be calculated automatically also like this:

```ruby
def finish(started_at:, finished_at:, recorded_at: now)
  self << { trace: :finish, started_at: started_at, finished_at: finished_at, duration_in_sec: Integer(finished_at - started_at), recorded_at: recorded_at }
end
```

ybiquitous (commented):
Just in case, please add an entry to the changelog. 🙏 
(because the response schema is changed)

lib/runners/cli.rb
yykamei: 0296887d

lib/runners/trace_writer.rb
```diff
@@ -82,6 +82,12 @@ def error(message, recorded_at: now, max_length: 4_000)
       end
     end
 
+    def finish(duration_in_sec:, started_at:, finished_at:, recorded_at: now)
+      started_at = started_at.utc.iso8601(3)
+      finished_at = finished_at.utc.iso8601(3)
+      self << { trace: :finish, duration_in_sec: duration_in_sec, started_at: started_at, finished_at: finished_at, recorded_at: recorded_at }
```
ybiquitous, line 88: To fix the type errors, we have to fix the signature like this:

https://github.com/sider/runners/blob/09ea1ae60340e34578a407790f4601a0e52c229f/sig/time.rbs#L2

```diff
-def iso8601: () -> String
+def iso8601: (?Integer) -> String
```

ybiquitous (approved):
LGTM 👍

lib/runners/cli.rb
yykamei: Thank you for approving this PR. I changed it after your approval, a893c682.

PR #1916, opened 2021-01-08
Do not use `--no-default-rc` flag for `yarn`

ybiquitous:

This change aims to fix a bug about the `--no-default-rc` flag of Yarn (introduced with #1876).

If using `--no-default-rc`, Yarn ignores any auto tokens in `.yarnrc` or `.npmrc` files.
This means that we **cannot install private packages** via auth tokens like FontAwesome.

```console
$ yarn --help | grep no-default-rc
    --no-default-rc                     prevent Yarn from automatically detecting yarnrc and npmrc files
```

Thus, this change moves yarnrc files temporarily instead of using `--no-default-rc`.

Related to #1876

> Check the following.


ybiquitous:
Here are the smoke test cases about yarnrc:

https://github.com/sider/runners/blob/27da66ffbb490d31405e5a0354bc9d4f4ae6ce17/test/smokes/eslint/expectations.rb#L491-L503

yykamei (approved):

ybiquitous:
@yykamei Thanks for the review 👍

PR #1925, opened 2021-01-12
[RuboCop] Fix issue links for rubydoc.info

ybiquitous:

[rubydoc.info](https://www.rubydoc.info) returns the 404 page for some gems.
(e.g. <https://www.rubydoc.info/gems/chefstyle>)

So, this change uses links within the output of the `gem info` command, instead.
See also <https://guides.rubygems.org/command-reference/#gem-info>.

None.

> Check the following.


lib/runners/processor/rubocop.rb
```diff
@@ -253,9 +253,13 @@ def check_rubocop_yml_warning(stderr)
     end
 
     # @see https://github.com/rubocop-hq/rubocop/blob/v0.76.0/lib/rubocop/cop/message_annotator.rb#L62-L63
-    def extract_links(original_message)
-      URI.extract(original_message, %w(http https))
```
ybiquitous, line 257: [note] `URI.extract` is obsolete.

```console
$ ruby -W -r uri -e 'URI.extract ""'
-e:1: warning: URI.extract is obsolete
```

https://ruby-doc.org/stdlib-2.7.2/libdoc/uri/rdoc/URI.html#extract-method

lib/runners/processor/rubocop.rb
```diff
@@ -253,9 +253,13 @@ def check_rubocop_yml_warning(stderr)
...
-        .map { |uri| uri.delete_suffix(",").delete_suffix(")") }
+    def extract_links(text)
+      @uri_regexp ||= URI::DEFAULT_PARSER.make_regexp(["http", "https"])
+      text.to_enum(:scan, @uri_regexp).map do
+        match = Regexp.last_match or raise
+        uri = match[0] or raise
+        uri.delete_suffix(",").delete_suffix(")")
+      end
```
yykamei, line 262: LGTM.

By the way, I got an error after running the new test on my local machine. That's because my environment has installed `rubocop-thread_safety` locally. How about using `uniq` just in case?

```
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-["https://github.com/covermymeds/rubocop-thread_safety"]
+["https://github.com/covermymeds/rubocop-thread_safety", "https://github.com/covermymeds/rubocop-thread_safety"]
```

lib/runners/processor/rubocop.rb
```diff
@@ -253,9 +253,13 @@ def check_rubocop_yml_warning(stderr)
...
     # @see https://github.com/rubocop-hq/rubocop/blob/v0.76.0/lib/rubocop/cop/message_annotator.rb#L62-L63
-    def extract_links(original_message)
-      URI.extract(original_message, %w(http https))
-        .map { |uri| uri.delete_suffix(",").delete_suffix(")") }
+    def extract_links(text)
+      @uri_regexp ||= URI::DEFAULT_PARSER.make_regexp(["http", "https"])
+      text.to_enum(:scan, @uri_regexp).map do
+        match = Regexp.last_match or raise
+        uri = match[0] or raise
+        uri.delete_suffix(",").delete_suffix(")")
```
yykamei, line 261: [question] Why is it necessary to delete these suffixes?

ybiquitous: If these `delete_suffix` would be removed, the smoke test would fail as follows:

image

See also https://github.com/sider/runners/commit/2f8eb3dd08812448d2a3bf8d38a95fcc9efd35d6

lib/runners/processor/rubocop.rb
ybiquitous: Good catch! Fixed via af45eaa.

lib/runners/processor/rubocop.rb
yykamei: Thank you.

yykamei (approved):
:+1:

PR #1930, opened 2021-01-13
Rename `duration_in_sec` to `duration_in_ms` for `finish` trace

ybiquitous:

This change aims to keep consistency in the trace data schema.

https://github.com/sider/runners/blob/ae3c69b8d627f8f198d96fc602a7b3f1aa52deae/lib/runners/schema/trace.rb#L12

https://github.com/sider/runners/blob/ae3c69b8d627f8f198d96fc602a7b3f1aa52deae/lib/runners/schema/trace.rb#L18

See also #1914

> Check the following.

### DIFF
--- a/lib/runners/trace_writer.rb
+++ b/lib/runners/trace_writer.rb
@@ -8,11 +8,11 @@ module Runners
     end
 
     def command_line(command_line, recorded_at: now)
-      self << { trace: :command_line, command_line: command_line.map { |v| filter.mask(v) }, recorded_at: recorded_at }
+      self << { trace: :command_line, command_line: command_line.map { |v| filter.mask(v) }, recorded_at: time_to_s(recorded_at) }
     end
 
     def status(status, recorded_at: now)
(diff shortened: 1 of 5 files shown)


lib/runners/trace_writer.rb
```diff
@@ -110,5 +110,10 @@ def each_slice(string, size:)
         yield slice, truncated
       end
     end
+
+    def calc_duration_in_ms(start, finish)
+      seconds = finish - start
+      (seconds * 1000).truncate
```
yykamei, line 116: How about using `to_r` or `to_f` to get a more accurate duration?

```suggestion
      seconds = finish.to_r - start.to_r
      (seconds * 1000).truncate
```

ybiquitous: Perhaps, I think `Time#-` considers such accuracy internally... 🤔 

https://ruby-doc.org/core-2.7.2/Time.html#2D-method

yykamei: OK

yykamei (approved):

Issue #1825, opened 2020-12-21
add File Information Collector runner


gnakagaw:
Add a runner to collect file information as below:
* physical line of code
* last commit datetime

Note that we have to override `Processor#use_git_metadata_dir?` because we need git metadata to get last commit datetime.

PR #1931, opened 2021-01-13
Add a new smoke test type for git metadata analysis

gnakagaw:
We're implementing a runner to analyze git metadata (.git/ directory). #1825 
This PR modifies smoke test module to preserve git metadata directory in smoke test cases.

In a preparation phase, the smoke test module creates a new git repo, copies target files and makes a commit for them. It fails the process if the test case directory contains `.git/` directory.

I added `setup_existing_git_repository` method to prevent the problem. This is called instead of `prepare_git_repository` method when the runner analyse git metadata. Git metadata analysis is a special test case for the runners. Thus, I added a new test types and adding test case methods: `add_test_with_git_metadata` and `add_offline_test_with_git_metadata`.

checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 [Full diff](https://github.com/sider/runners/compare/0.40.5...HEAD)
 
 - **ESLint** Add `eslint-plugin-cypress` to pre-installed packages [#1936](https://github.com/sider/runners/pull/1936)
-- Add a new smoke test type using git metadata [#1931](https://github.com/sider/runners/pull/1931)
+
 ## 0.40.5
 
 [Full diff](https://github.com/sider/runners/compare/0.40.4...0.40.5)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 [Full diff](https://github.com/sider/runners/compare/0.40.5...HEAD)
 
 - **ESLint** Add `eslint-plugin-cypress` to pre-installed packages [#1936](https://github.com/sider/runners/pull/1936)
-
+- Add a new smoke test type using git metadata [#1931](https://github.com/sider/runners/pull/1931)
 ## 0.40.5
 
 [Full diff](https://github.com/sider/runners/compare/0.40.4...0.40.5)
--- a/test/smokes/smoke.rb
+++ b/test/smokes/smoke.rb
@@ -19,12 +19,13 @@ module Runners
       include Minitest::Assertions
 
       class TestParams
-        attr_reader :name, :pattern, :offline
+        attr_reader :name, :pattern, :offline, :use_git_metadata
 
-        def initialize(name:, pattern:, offline:)
+        def initialize(name:, pattern:, offline:, use_git_metadata:)
           @name = name
           @pattern = pattern
           @offline = offline
+          @use_git_metadata = use_git_metadata
         end
 
         def ==(other)
@@ -109,11 +110,19 @@ module Runners
 
       def run_test(params, out)
         command_output, _ = Dir.mktmpdir do |dir|
-          repo_dir, base, head = prepare_git_repository(
-            workdir: Pathname(dir).realpath,
-            smoke_target: expectations.parent.join(params.name).realpath,
-            out: out,
-          )
+          if params.use_git_metadata
+            repo_dir, base, head = setup_existing_git_repository(
+                workdir: Pathname(dir).realpath,
+                smoke_target: expectations.parent.join(params.name).realpath,
+                out: out,
+                )
+          else
+            repo_dir, base, head = prepare_git_repository(
+                workdir: Pathname(dir).realpath,
+                smoke_target: expectations.parent.join(params.name).realpath,
+                out: out,
+                )
+          end
           cmd = command_line(params: params, repo_dir: repo_dir, base: base, head: head)
           sh!(*cmd, out: out, exception: false)
         end
@@ -204,6 +213,17 @@ module Runners
         end
       end
 
+      def setup_existing_git_repository(workdir:, smoke_target:, out:)
+        smoke_dir = workdir.join("smoke").to_path
+        FileUtils.copy_entry "#{smoke_target}", smoke_dir
+
+        Dir.chdir(smoke_dir) do
+          head_commit, _ = sh! "git", "rev-parse", "HEAD", out: out
+          # TODO: Ignored Steep error
+          _ = [smoke_dir, nil, head_commit.chomp]
+        end
+      end
+
       def debug?
         ENV["DEBUG"] == "true"
       end
@@ -270,11 +290,19 @@ module Runners
       end
 
       def self.add_test(name, **pattern)
-        add_test_helper TestParams.new(name: name, pattern: build_pattern(**pattern), offline: false)
+        add_test_helper TestParams.new(name: name, pattern: build_pattern(**pattern), offline: false, use_git_metadata: false)
       end
 
       def self.add_offline_test(name, **pattern)
-        add_test_helper TestParams.new(name: name, pattern: build_pattern(**pattern), offline: true)
+        add_test_helper TestParams.new(name: name, pattern: build_pattern(**pattern), offline: true, use_git_metadata: false)
+      end
+
+      def self.add_test_with_git_metadata(name, **pattern)
+        add_test_helper TestParams.new(name: name, pattern: build_pattern(**pattern), offline: false, use_git_metadata: true)
+      end
+
+      def self.add_offline_test_with_git_metadata(name, **pattern)
+        add_test_helper TestParams.new(name: name, pattern: build_pattern(**pattern), offline: true, use_git_metadata: true)
       end
 
       def self.add_test_helper(test)
--- a/test/smokes/smoke.rb
+++ b/test/smokes/smoke.rb
@@ -111,13 +111,13 @@ module Runners
       def run_test(params, out)
         command_output, _ = Dir.mktmpdir do |dir|
           if params.use_git_metadata
-            repo_dir, base, head = setup_existing_git_repository(
+            repo_dir, base, head = prepare_existing_git_repository(
                 workdir: Pathname(dir).realpath,
                 smoke_target: expectations.parent.join(params.name).realpath,
                 out: out,
                 )
           else
-            repo_dir, base, head = prepare_git_repository(
+            repo_dir, base, head = prepare_new_git_repository(
                 workdir: Pathname(dir).realpath,
                 smoke_target: expectations.parent.join(params.name).realpath,
                 out: out,
@@ -187,7 +187,7 @@ module Runners
         commands
       end
 
-      def prepare_git_repository(workdir:, smoke_target:, out:)
+      def prepare_new_git_repository(workdir:, smoke_target:, out:)
         # Create a bare repository
         bare_dir = workdir.join("bare").to_path
         sh! "git", "init", "--bare", bare_dir, out: out
@@ -213,7 +213,7 @@ module Runners
         end
       end
 
-      def setup_existing_git_repository(workdir:, smoke_target:, out:)
+      def prepare_existing_git_repository(workdir:, smoke_target:, out:)
         smoke_dir = workdir.join("smoke").to_path
         FileUtils.copy_entry smoke_target, smoke_dir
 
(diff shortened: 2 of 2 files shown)


yykamei:
I understood your situation. You want to add `.git` in a smoke test target like `test/smokes/analyzer/with_git/.git`, don't you? If so, is it really necessary to include`.git` in a smoke target? Can't we use the Git metadata created by `prepare_git_repository'?

test/smokes/smoke.rb
```diff
@@ -205,6 +214,17 @@ def prepare_git_repository(workdir:, smoke_target:, out:)
...
       end
 
+      def setup_existing_git_repository(workdir:, smoke_target:, out:)
+        smoke_dir = workdir.join("smoke").to_path
+        FileUtils.copy_entry "#{smoke_target}", smoke_dir
+
+        Dir.chdir(smoke_dir) do
+          head_commit, _ = sh! "git", "rev-parse", "HEAD", out: out
+          # TODO: Ignored Steep error
+          _ = [smoke_dir, nil, head_commit.chomp]
```
ybiquitous, line 224: [suggestion] Please remove the TODO comment that is no longer needed. See #1933.

```suggestion
          [smoke_dir, nil, head_commit.chomp]
```

test/smokes/smoke.rb
```diff
@@ -205,6 +214,17 @@ def prepare_git_repository(workdir:, smoke_target:, out:)
         end
       end
 
+      def setup_existing_git_repository(workdir:, smoke_target:, out:)
+        smoke_dir = workdir.join("smoke").to_path
+        FileUtils.copy_entry "#{smoke_target}", smoke_dir
```
ybiquitous, line 219: [nitpick]

```suggestion
        FileUtils.copy_entry smoke_target, smoke_dir
```

test/smokes/smoke.rb
```diff
@@ -205,6 +214,17 @@ def prepare_git_repository(workdir:, smoke_target:, out:)
         end
       end
 
+      def setup_existing_git_repository(workdir:, smoke_target:, out:)
```
ybiquitous, line 217: [suggestion] How about renaming the methods for readability?

- `prepare_git_repository` -> `prepare_new_git_repository`
- `setup_existing_git_repository` -> `prepare_existing_git_repository`

test/smokes/smoke.rb
```diff
@@ -109,11 +110,19 @@ def run
...
+                smoke_target: expectations.parent.join(params.name).realpath,
+                out: out,
+                )
+          else
+            repo_dir, base, head = prepare_git_repository(
+                workdir: Pathname(dir).realpath,
+                smoke_target: expectations.parent.join(params.name).realpath,
+                out: out,
+                )
+          end
```
ybiquitous, line 125: [suggestion] How about the following refactoring?

```suggestion
          repo_args = {
            workdir: Pathname(dir).realpath,
            smoke_target: expectations.parent.join(params.name).realpath,
            out: out,
          }
          repo_dir, base, head =
            if params.use_git_metadata
              setup_existing_git_repository(**repo_args)
            else
              prepare_git_repository(**repo_args)
            end
```

gnakagaw:
@yykamei 
> You want to add `.git` in a smoke test target like `test/smokes/analyzer/with_git/.git`, don't you?

Yes

> If so, is it really necessary to include`.git` in a smoke target? Can't we use the Git metadata created by `prepare_git_repository`?

We can't the auto generated git metadata because the commit datetime is dynamic. [The datetime is the moment the smoke test module executes `git commit` ](https://github.com/sider/runners/blob/master/test/smokes/smoke.rb#L198). 

`metrics_fileinfo` runner (#1825) use `git log` command to get the last commit datetime for each file. On the other hand, commit datetime in the smoke test cases are deterministic.(i.e., we write the exact datetime in the test case) Thus, if the datetime is dynamic,  the smoke tests always fail.

yykamei:
That's understandable.

By the way, isn't it sufficient to use `:_` like this? Do you want to exactly make your tests match the DateTime?

https://github.com/sider/runners/blob/64da70be0d8bee234f349ecc2aba47122880a734/test/smokes/shellcheck/expectations.rb#L17

gnakagaw:
@yykamei 
> By the way, isn't it sufficient to use `:_` like this? Do you want to exactly make your tests match the DateTime?

Commit datetime is one of  target information generated by `metric_fileinfo` runner (#1825). In my opinion, the smoke test should ensure that the runner generate the information correctly.

yykamei:
Isn't `Regexp` also suitable like this?

https://github.com/sider/runners/blob/01b113befc65dea4fad369da0126498075c74f99/test/smokes/cppcheck/expectations.rb#L30

gnakagaw:
@yykamei
You mean that it is enough to check string format, don't you? It is one of options. But we can't check some test case with that.

Here is an example. We assume that there is a git repository like this. I added a file and modified twice.
```bash
gnakagaw@sv2:~/tmp/git-tst$ ls
dummy.txt
gnakagaw@sv2:~/tmp/git-tst$ git log --name-status
commit 4946de022b83025fdca370a16032092f5ca8a0c1 (HEAD -> master)
Author: Gaku Nakagawa <gaku@sleeek.io>
Date:   Wed Jan 13 11:36:00 2021 +0900

    modify once more

M       dummy.txt

commit 141964f724cbab52b63c9888f2dd265beed039f0
Author: Gaku Nakagawa <gaku@sleeek.io>
Date:   Wed Jan 13 11:35:14 2021 +0900

    modify file

M       dummy.txt

commit d3c4acc020910f04b3b2e42966a3d5f765c71bd7
Author: Gaku Nakagawa <gaku@sleeek.io>
Date:   Fri Jan 1 10:00:00 2021 +0900

    add a file

A       dummy.txt
gnakagaw@sv2:~/tmp/git-tst$
```
The generated datetime by `metrics_fileinfo` must be ` Wed Jan 13 11:36:00 2021 +0900`. To ensure this, we need exact matching for datetime in a smoke test.

yykamei:
> But we can't check some test case with that.

I'm curious why we can't check them?

gnakagaw:
> I'm curious why we can't check them?
`metrics_runner` might return ` Fri Jan 1 10:00:00 2021 +0900`, the first commit datetime, if we have mistakes. We cannot catch the case only with string pattern matching, can we?

yykamei:
I understood why I'm less likely to agree with that. I can't exactly imagine the implementation of `metrics_runner` at present. At this point, I'm not convinced that `metrics_runner` really requires an exact match. However, I understood you really want to make the test match accurately. Thanks.

test/smokes/smoke.rb
```diff
@@ -205,6 +211,16 @@ def prepare_git_repository(workdir:, smoke_target:, out:)
         end
       end
 
+      def prepare_existing_git_repository(workdir:, smoke_target:, out:)
```
yykamei, line 214: I think it's not sufficient because the directory named `.git` cannot be committed in Git; it means `.git` cannot be included in https://github.com/sider/runners. Maybe, an additional argument is required to be specified the folder that contains Git metadata.

gnakagaw: > Maybe, an additional argument is required to be specified the folder that contains Git metadata.

We can do that with `.git` file (not a directory) with out any command line options for `git` command. 
```
$ pwd
[masked]/git_test
$ mkdir my_git_metadata
$ git init  --separate-git-dir=my_git_metadata
Initialized empty Git repository in [masked]/git_test/my_git_metadata/
$ touch foo
$ git add foo
$ git commit -m "add foo"
[master (root-commit) 12945cf] add foo
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo
$ ls -al
total 16
drwxr-xr-x  3 gnakagaw gnakagaw 4096 Jan 13 13:04 .
drwxrwxrwx 12 gnakagaw gnakagaw 4096 Jan 13 13:03 ..
-rw-r--r--  1 gnakagaw gnakagaw   52 Jan 13 13:04 .git
-rw-r--r--  1 gnakagaw gnakagaw    0 Jan 13 13:04 foo
drwxr-xr-x  8 gnakagaw gnakagaw 4096 Jan 13 13:04 my_git_metadata
$ cat .git
gitdir: /home/gnakagaw/tmp/git_test/my_git_metadata
$ git log --oneline
12945cf (HEAD -> master) add foo
$
```

So, we can store a smoke test set like this.
```
$ pwd
[masked]/runners/test/smokes/metrics_fileinfo
$ tree -L 3 -a
.
├── expectations.rb
└── success
    ├── .git
    ├── dot_git
    │   ├── COMMIT_EDITMSG
    │   ├── HEAD
    │   ├── branches
    │   ├── config
    │   ├── description
    │   ├── hooks
    │   ├── index
    │   ├── info
    │   ├── logs
    │   ├── objects
    │   └── refs
    └── hello.rb
```

gnakagaw:
> I understood why I'm less likely to agree with that. I can't exactly imagine the implementation of `metrics_runner` at present. At this point, I'm not convinced that `metrics_runner` really requires an exact match. However, I understood you really want to make the test match accurately. Thanks.

Thanks a lot for understanding. 🙏 

To tell the truth, I also think that it is rare that we really need exact datetime match. I'll use ` git log -1 [file name]` to get the last commit datetime for a file.  So it doesn't happen unless I made terrible mistakes. However, I think that we have to check the value exactly because it is the target value generated by `metrics_fileinfo` runner.

test/smokes/smoke.rb
yykamei: Is this config still available on GitHub Actions?

gnakagaw: It have failed on #1937. . . I found that we cannot add `.git`  even if it is file not directory. I'll make another plan..

yykamei: How about just renaming `.git` to something and renaming it to `.git` again during tests?
The workflow is like this:

1. A developer creates a smoke test with Git and puts `my_git_dir` to our VCS
2. `Runners::Testing::Smoke` prepares a working directory with the specified `my_git_dir` and rename it to `.git` in the directory
3. Start test

gnakagaw: Thanks for your suggestion. I've implement so that in 20608e62d925eef3df3cb4e3d383ac4965cc1654

gnakagaw: It seems to be ok on Github Action job. 
https://github.com/sider/runners/pull/1937/checks?check_run_id=1706336659

yykamei (approved):
OK

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 [Full diff](https://github.com/sider/runners/compare/0.40.5...HEAD)
 
 - **ESLint** Add `eslint-plugin-cypress` to pre-installed packages [#1936](https://github.com/sider/runners/pull/1936)
-
+- Add a new smoke test type using git metadata [#1931](https://github.com/sider/runners/pull/1931)
```
ybiquitous, line 10: [nitpick] An empty line seems needed before a heading.

```suggestion
- Add a new smoke test type using git metadata [#1931](https://github.com/sider/runners/pull/1931)

```

test/smokes/smoke.rb
```diff
@@ -109,11 +110,16 @@ def run
...
         command_output, _ = Dir.mktmpdir do |dir|
-          repo_dir, base, head = prepare_git_repository(
-            workdir: Pathname(dir).realpath,
-            smoke_target: expectations.parent.join(params.name).realpath,
-            out: out,
-          )
+          repo_args = {
+              workdir: Pathname(dir).realpath,
+              smoke_target: expectations.parent.join(params.name).realpath,
+              out: out,
```
ybiquitous, line 116: [nitpick] Fix indentation.

```suggestion
            workdir: Pathname(dir).realpath,
            smoke_target: expectations.parent.join(params.name).realpath,
            out: out,
```

CHANGELOG.md
ybiquitous: Oh, sorry. This PR is about only testing, so we don't need to add a changelog entry.

gnakagaw: OK. reverted in 13c4340ce7a85d5ca16f06fcba1ccaede66a42bb

test/smokes/smoke.rb
gnakagaw: fixed: 26cbef794e0cb9329b56970ab65195d47e44719d

ybiquitous (approved):
👍

gnakagaw:
Thanks for your kind reviewing. I'll merge it.

PR #1937, opened 2021-01-14
Add file information collector runner

### LINKED ISSUE #1825: add File Information Collector runner

gnakagaw:
Fix #1825

This PR adds `metrics_fileinfo` runner: a runner to collect lines of code and last commit datetime.
NOTE: This PR modify more than 200 files. But, almost of them are git metadata files (.git/). You may ignore that.

checklist:


images/metrics_fileinfo/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+# NOTE: DO *NOT* EDIT THIS FILE. IT IS GENERATED.
+# PLEASE UPDATE Dockerfile.erb INSTEAD OF THIS FILE.
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+FROM sider/devon_rex_base:2.28.0
+
+# install file command (file type guesser)
+USER root
```
gnakagaw, line 8: > __[Goodcheck]__ Try to avoid using the `root` user as much as possible. ([view](https://sider.review/gh/repos/202486306/pulls/1937#issue-5307353))
>
> <details>
>
> Each runner is executed by the non-root user for security reasons.
> Using the `USER` command leads to more layers and bigger image size.
>
> | Rule |
> | ---- |
> | `sider.runners.root_user_in_dockerfile` |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1937#issue-5307353) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

We need root operation to install `file` command.

images/metrics_fileinfo/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+USER $RUNNER_USER
+
+
+ARG RUNNERS_DIR=${RUNNER_USER_HOME}/runners
+
+# Install required gems first (due to slow download)
+COPY --chown=${RUNNER_USER}:${RUNNER_GROUP} Gemfile Gemfile.lock runners.gemspec analyzers.yml ${RUNNERS_DIR}/
+COPY --chown=${RUNNER_USER}:${RUNNER_GROUP} lib/runners/version.rb ${RUNNERS_DIR}/lib/runners/
+
+RUN cd ${RUNNERS_DIR} && \
```
gnakagaw, line 21: > __[hadolint]__ Use WORKDIR to switch to a directory ([view](https://sider.review/gh/repos/202486306/pulls/1937#issue-5307350))
>
> <details>
>
> | Rule | Severity |
> | ---- | -------- |
> | `DL3003` | `warning` |
>
> References:
> - <https://github.com/hadolint/hadolint/wiki/DL3003>
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1937#issue-5307350) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

This is implemented  in a base template file..

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,83 @@
+module Runners
+  class Processor::MetricsFileInfo < Processor
+    Schema = _ = StrongJSON.new do
+      # @type self: SchemaClass
+
+      let :runner_config, Schema::BaseConfig.base
+      let :issue, object(
+          line_of_code: integer?,
+          last_commit_datetime: string
```
suetsuguk, line 9: Can you use `Time` class rather than `string` so that we can easily find most recent file in a directory at the Sideci code?
Also, `git log` with `--format=format:%at(not aI)` tells us UNIX timestamp to be passed to the `Time.at()` method.

ybiquitous: StrongJSON does not natively support the `Time` class. Perhaps, it seems preferred to save a timestamp as the ISO8601 format like `TraceWriter`:

https://github.com/sider/runners/blob/d04de90bbed730a159a3efcdc168a09fae9f9889/lib/runners/trace_writer.rb#L101

README.md
```diff
@@ -29,6 +29,7 @@ All 36 analyzers are provided as a Docker image:
 | JSHint | [docker](https://hub.docker.com/r/sider/runner_jshint), [source](https://github.com/jshint/jshint), [doc](https://help.sider.review/tools/javascript/jshint) | ✅ |
 | ktlint | [docker](https://hub.docker.com/r/sider/runner_ktlint), [source](https://github.com/pinterest/ktlint), [doc](https://help.sider.review/tools/kotlin/ktlint) | ✅ |
 | LanguageTool | [docker](https://hub.docker.com/r/sider/runner_languagetool), [source](https://github.com/languagetool-org/languagetool), [doc](https://help.sider.review/tools/others/languagetool) | ✅ |
+| Metrics File Info | [docker](https://hub.docker.com/r/sider/runner_metrics_fileinfo) | ✅ |
```
suetsuguk, line 32: Add URLs for both `source` and `doc` to the list of analyzers as;
```suggestion
| Metrics File Info | [docker](https://hub.docker.com/r/sider/runner_metrics_fileinfo), [source](https://github.com/coreutils/coreutils), [doc](https://help.sider.review/tools/metrics/fileinfo) | ✅ |
```
This is done by issuing `$ bundle exec rake readme:generate` command after adding following lines in `analyzers.yml`.
Regarding the doc path, refer to [this comment](https://github.com/sider/runners/pull/1878#discussion_r553749388).

```
  metrics_fileinfo:
    name: Metrics File Info
    github: coreutils/coreutils
    doc: tools/metrics/fileinfo
```

images/metrics_fileinfo/Dockerfile.erb
```diff
@@ -0,0 +1,12 @@
+FROM sider/devon_rex_base:2.28.0
+
+# install file command (file type guesser)
+USER root
+RUN apt-get update -y && \
+    apt-get install -qq -y --no-install-recommends file="1:5.35-4+deb10u1" && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+USER $RUNNER_USER
```
suetsuguk, line 9: I think `file` command is so stable that we can always use the latest version. Can we make the version relaxed as follows?

```suggestion
USER root
RUN apt-get update && \
    apt-get install -y --no-install-recommends file && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
USER $RUNNER_USER
```

This is the exact copy from our `shellcheck` [Dockerfile](https://github.com/sider/runners/blob/master/images/shellcheck/Dockerfile.erb).

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,83 @@
...
+      )
+    end
```
suetsuguk, line 11: [nitpick] inconsistent indentation

```suggestion
      let :issue, object(
        line_of_code: integer?,
        last_commit_datetime: string
      )
    end
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    FILE_COMMAND_VALID_PLAIN_TEXT_FORMATS = ["ASCII text", "UTF-8 Unicode text", "ISO-8859 text"].freeze
+
+    def analyzer_version
+      @analyzer_version ||= capture3!("wc", "--version").then do |stdout, |
+        stdout.split(/\R/)[0][/\d*\.\d*/].then {|ver| ver.nil? ? "unknown" : ver}
+      end
+    end
```
suetsuguk, line 19: Can you use useful `extract_version!` method of `Processor` class as?

```suggestion
    def analyzer_version
      @analyzer_version ||= extract_version!("wc")
    end
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+        generate_issues(loc, last_commit_datetime).each { |issue| result.add_issue(issue) }
+      end
+    end
+
+    private
+
+    def generate_file_list
+      Dir.glob("**/*", File::FNM_DOTMATCH).reject { |e| File.directory? e or e.start_with? ".git/" }
```
suetsuguk, line 39: Can you use the `Changes.all_files_in()` method? See the [code](https://github.com/sider/runners/blob/d04de90bbed730a159a3efcdc168a09fae9f9889/lib/runners/changes.rb#L77-L84).

suetsuguk: Just a comment. We can get a list of files except `.git` directory by issuing `git ls-files`. With `--eol` option, we also get a file attribute (e.g. text, binary). But, from the view of performance, it seems to be not good to get tons of file-names from stdout of `git` command rather than standard `Dir.glob()` method.

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,83 @@
...
+        stdout.split(/\R/)[0][/\d*\.\d*/].then {|ver| ver.nil? ? "unknown" : ver}
+      end
+    end
+
+    # This analyser use git metadata (.git/).
+    def use_git_metadata_dir?
+      true
+    end
+
+    def analyze(_changes)
```
suetsuguk, line 26: I prefer simple **one-loop** structure rather than several Hash instances for each metric.
A pseudo code below reduces the number of loops (and the memory usage of hash).

```ruby
target_files.each |file|
  loc = analyze_line_of_code(file) # returns Integer value of LOC
  last_commit_datetime = analyze_last_commit_datetime(file) # returns String value of date-time

    result.add_issue(construct_issue(loc, last_commit_datetime))
  end
end
```

Could you please reconsider?

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    def analyze_last_commit_datetime(target_files)
+      target_files.map do |file|
+        stdout, _, _  = capture3!("git", "log", "-1", "--format=format:%aI", file)
+        [file, stdout]
+      end.to_h
+    end
+
+    def is_text_file?(target_file)
+      stdout, _, _ = capture3!("file", "-E", "-b", target_file)
```
suetsuguk, line 77: I think it is better to use mime format with `--mime` which describes in more machine-readable forms.

```console
$ file -E -b README.md
exported SGML document, UTF-8 Unicode text
$ file -E -b --mime README.md
text/plain; charset=utf-8
$ file -E -b --mime logo.jpg
image/jpeg; charset=binary
```

suetsuguk:
As for [expectations.rb](https://github.com/gnakagaw/runners/blob/7cc589c707671ba1a607b970d9354c55acd94798/test/smokes/metrics_fileinfo/expectations.rb), there are no test files in sub-directories. We need to place some files in sub-directories to test the line of `path: relative_path(fname),` in metrics_fileinfo.rb.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: @suetsuguk It seems to be difficult to use `Time` class. according to ybiquitous's comment. What is the suitable format for SideCI Code?

suetsuguk: Any format which can be converted to `Time` class is acceptable. iso8601 is fine for me. The `Time.iso8601()` method will be used.

gnakagaw: @suetsuguk OK. The current code use ISO8601 extended. ([an example](https://github.com/gnakagaw/runners/blob/1825_add_file_information_collector_runner/test/smokes/metrics_fileinfo/expectations.rb#L17)). We can parse the format with `Time.iso8601()`.

```bash
irb(main):003:0> require 'time'
=> false
irb(main):004:0> Time.iso8601("2021-01-01T10:00:00+09:00")
=> 2021-01-01 10:00:00 +0900
irb(main):005:0>
```

suetsuguk: I confirmed that the Ruby docs shows the same format as an [example](https://docs.ruby-lang.org/ja/latest/method/Time/s/iso8601.html). Keep the current code unchanged. Thank you.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: Thx. I've fixed my editor's settings.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: > Can you use the equivalent Changes.all_files_in() method? 

Thanks, it's the identical method. But we cannot call it from a runner because it is a private class method of `Changes` class.
https://github.com/sider/runners/blob/d04de90bbed730a159a3efcdc168a09fae9f9889/lib/runners/changes.rb#L84

gnakagaw: > . We can get a list of files except .git directory by issuing git ls-files. With --eol option, we also get a file attribute (e.g. text, binary).
It's useful for `is_text_file?` method. We can remove `file` command from the runner container. I'll use it in it.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: It's so reasonable. I'll rewrite them.

lib/runners/processor/metrics_fileinfo.rb
suetsuguk: Seen from the code, `all_files_in()` is defined as `self` class method, and has no side-effects. I think we can raise it to public.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: I'll rewrite the method with `git ls-files --eol [target file]`. But this is good tips. Thanks.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: memo: The method was implemented in the commit.
https://github.com/sider/runners/commit/63f894ced6c25698838a70a32c517cbeeaaa7728

lib/runners/processor/metrics_fileinfo.rb
suetsuguk: The `git ls-files --eol` only tells us if a file is text or binary. We cannot obtain further details. We may want to know file types (e.g. image, pdf, zip) and use them to show the file information on Sideci in the future. Maybe not now, but someday. I agree to use `git ls-files` this time.

README.md
```diff
@@ -29,6 +29,7 @@ All 36 analyzers are provided as a Docker image:
 | JSHint | [docker](https://hub.docker.com/r/sider/runner_jshint), [source](https://github.com/jshint/jshint), [doc](https://help.sider.review/tools/javascript/jshint) | ✅ |
 | ktlint | [docker](https://hub.docker.com/r/sider/runner_ktlint), [source](https://github.com/pinterest/ktlint), [doc](https://help.sider.review/tools/kotlin/ktlint) | ✅ |
 | LanguageTool | [docker](https://hub.docker.com/r/sider/runner_languagetool), [source](https://github.com/languagetool-org/languagetool), [doc](https://help.sider.review/tools/others/languagetool) | ✅ |
+| Metrics File Info | [docker](https://hub.docker.com/r/sider/runner_metrics_fileinfo), [source](https://github.com/coreutils/coreutils), [doc](https://help.sider.review/tools/metrics/fileinfo) | ✅ |
```
gnakagaw, line 32: > __[remark-lint]__ Link to https://help.sider.review/tools/metrics/fileinfo is dead ([view](https://sider.review/gh/repos/202486306/pulls/1937#issue-5341368))
>
> <details>
>
> | Rule | Severity |
> | ---- | -------- |
> | `no-dead-urls` | `warn` |
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1937#issue-5341368) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

We'll create it later.

gnakagaw:
> As for [expectations.rb](https://github.com/gnakagaw/runners/blob/7cc589c707671ba1a607b970d9354c55acd94798/test/smokes/metrics_fileinfo/expectations.rb), there are no test files in sub-directories. We need to place some files in sub-directories to test the line of `path: relative_path(fname),` in metrics_fileinfo.rb.

fix: 50e4886159ee3d5509cafd5865539ae99a8190bc

README.md
gnakagaw: fix: 0b1d563002b4aca9e2bf36a392c88823c241090b

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: 989b6dc31fe7409e407f0dbd148571321b82eece

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: 3ee11d1f329743ce85acba5da3234c7c49dab038

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: b21733a5b1390466dafefb787f367c8e82e40485

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: I've modified the code to use `changes` from the runner harness (d92ecb683af2970612ffb215fa6333d02e41e5d6). It's identical information from`all_files_in()` if an analyse request targeting whole a branch.

This process generates `changes`.
https://github.com/sider/runners/blob/f120729ae0c75609838da843ddfb7354df77af08/lib/runners/workspace.rb#L32

digging..
https://github.com/sider/runners/blob/f120729ae0c75609838da843ddfb7354df77af08/lib/runners/changes.rb#L63-L64

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,64 @@
...
+        location: nil,
+        id: "metrics_fileinfo",
+        message: "#{path_str(path)}: loc = #{loc}, last commit datetime = #{commit_datetime}",
+        object: {
+          line_of_code: loc,
+          last_commit_datetime: commit_datetime
+        },
+        schema: Schema.issue
+      )
+    end
```
suetsuguk, line 44: [suggestion] How about passing only `path` variable to `generate_issue()` as follows?

This allows us to reduce the number of arguments of the `generate_issue()` method. Also the `path_str()` method is no longer needed.

```suggestion
    def analyze(changes)
      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
        changes.changed_paths.each do |path|
          result.add_issue(generate_issue(path))
        end
      end
    end

    private

    def generate_issue(path)
      filepath = relative_path(path).to_s
      loc = analyze_line_of_code(filepath)
      commit_datetime = analyze_last_commit_datetime(filepath)

      Issue.new(
        path: relative_path(path),
        location: nil,
        id: "metrics_fileinfo",
        message: "#{filepath}: loc = #{loc}, last commit datetime = #{commit_datetime}",
        object: {
          line_of_code: loc,
          last_commit_datetime: commit_datetime
        },
        schema: Schema.issue
      )
    end
```

gnakagaw: It's so smart. I've fixed in fda89f5c2c1be2b7775e82bd2f9afcfc90a5d995.

suetsuguk: There is `def path_str: (Pathname) -> String` remaining in [metrics_fileinfo.rbs](https://github.com/sider/runners/blob/fda89f5c2c1be2b7775e82bd2f9afcfc90a5d995/sig/runners/processor/metrics_fileinfo.rbs#L19). Thank you.

suetsuguk (approved):
Thank you!

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: Thanks cdee994cde7d8d9e6c4424b27741209de85e4fc2

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,64 @@
+module Runners
+  class Processor::MetricsFileInfo < Processor
+    Schema = _ = StrongJSON.new do
+      # @type self: SchemaClass
+
+      let :runner_config, Schema::BaseConfig.base
+      let :issue, object(
+        line_of_code: integer?,
+        last_commit_datetime: string
```
suetsuguk, line 9: @gnakagaw As for the type of `last_commit_datetime`, can you use `integer` instead of `string`?
On Sideci we need to convert it to a `Time` instance to find most recent file.
Converting from a text value takes more cost than from a integer value.
This will occur multiple times (num of directories * num of files in a directory).
An integer value may decrease readability for debugging. But, we can show the text value in the `message` value as a reference.
What do you think?

gnakagaw: @suetsuguk I almost agree with you. But we lost timezone information. Is it OK?

suetsuguk: I'm fine. We can show a human-readable text with respect to user's timezone on frontend. We can also provide smarter indications such as "2 days ago", "1 year ago" like GitHub.

gnakagaw: OK.I'll modify so that.

gnakagaw: fix: 22c6cd0423f4288181991e79296c8da37920b301 5d872f287164014a9beadc724a120a3fd2f70ca7

suetsuguk: [suggestion] With the argument `--format=format:%aI,%at`, we can get both iso8601 and unix time at a time.
This implies there is no need to use the `Time` class in this code.

```console
$ git log -1 --format=format:%aI,%at README.md
2020-11-13T11:47:01+09:00,1605235621
```

Maybe, the code goes to something like bellow. Hash is a simple way to return multiple results.

```ruby
def analyze_last_commit_datetime(target)
capture3!("git", "log", "-1", "--format=format:%aI,%at", target).then {|stdout,|
  times = stdout.split(",")

  {
    iso8601: times[0],
    unix: Integer(times[1]),
  }
}
end
```

Thank you.

suetsuguk: There is a possibility that [ISO8601](https://ja.wikipedia.org/wiki/ISO_8601) format could contain comma `,` and period `.`
We need to choose another character such as `|`, `;`, or `=`.

suetsuguk: Confirmed. Finally, how about using `Integer()` instead of `to_i` as suggested by Querly?

gnakagaw: > Integer() instead of to_i as suggested by Querly?
I'v tried that. But I faced a error from Steep. Now investigating...

images/metrics_fileinfo/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM sider/devon_rex_base:master
+
+
+ARG RUNNERS_DIR=${RUNNER_USER_HOME}/runners
+
+# Install required gems first (due to slow download)
+COPY --chown=${RUNNER_USER}:${RUNNER_GROUP} Gemfile Gemfile.lock runners.gemspec analyzers.yml ${RUNNERS_DIR}/
+COPY --chown=${RUNNER_USER}:${RUNNER_GROUP} lib/runners/version.rb ${RUNNERS_DIR}/lib/runners/
+
+RUN cd ${RUNNERS_DIR} && \
```
gnakagaw, line 14: > __[hadolint]__ Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check ([view](https://sider.review/gh/repos/202486306/pulls/1937#issue-5350716))
>
> <details>
>
> | Rule | Severity |
> | ---- | -------- |
> | `DL4006` | `warning` |
>
> References:
> - <https://github.com/hadolint/hadolint/wiki/DL4006>
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/1937#issue-5350716) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

Close. This is generated from a template file.

lib/runners/processor/metrics_fileinfo.rb
suetsuguk: Confirmed. Thank you.

suetsuguk:
@ybiquitous Do you have the time to go through and review the code changes in this PR?  Thank you.

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,66 @@
...
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+        changes.changed_paths.each do |path|
+          result.add_issue(generate_issue(path))
+        end
+      end
+    end
+
+    private
+
+
```
ybiquitous, line 32: [nitpick] duplicate empty lines:

```suggestion

```

sig/runners/processor/metrics_fileinfo.rbs
```diff
@@ -0,0 +1,20 @@
...
+      def runner_config: () -> StrongJSON::Type::Object[untyped]
+      def issue: () -> StrongJSON::Type::Object[untyped]
+    end
+    Schema: SchemaClass
+
+    private
+
+    def generate_issue: (Pathname)-> Issue
+
+    def analyze_line_of_code: (String) -> (nil | int)
```
ybiquitous, line 13: [nitpick]

```suggestion
    def analyze_line_of_code: (String) -> Integer?
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,66 @@
...
+    def use_git_metadata_dir?
+      true
+    end
+
+    def analyze(changes)
+      Results::Success.new(guid: guid, analyzer: analyzer).tap do |result|
+        changes.changed_paths.each do |path|
+          result.add_issue(generate_issue(path))
+        end
+      end
```
ybiquitous, line 27: [suggestion] How about this refactoring?

```suggestion
      Results::Success.new(
        guid: guid,
        analyzer: analyzer,
        issues: changes.changed_paths.map { |path| generate_issue(path) },
      )
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,66 @@
...
+      last_commit_iso, last_commit_epoch = analyze_last_commit_datetime(filepath)
+
+      Issue.new(
+        path: relative_path(path),
+        location: nil,
+        id: "metrics_fileinfo",
+        message: "#{filepath}: loc = #{loc}, last commit datetime = #{last_commit_iso}",
+        object: {
+          line_of_code: loc,
+          last_commit_datetime: Integer(last_commit_epoch)
```
ybiquitous, line 45: [question] What reason to use epoch time instead of ISO format?

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,66 @@
...
+        },
+        schema: Schema.issue
+      )
+    end
+
+    def analyze_line_of_code(target)
+      is_text_file?(target) ? capture3!("wc", "-l", target).then {|stdout,| Integer(stdout.split(" ")[0])} : nil
+    end
```
ybiquitous, line 53: [suggestion] How about this refactoring?

- Call `is_text_file?` out of `analyze_line_of_code`
- Not use `then`, it's more straitforward

```suggestion
    def analyze_line_of_code(target)
      stdout, _ = capture3!("wc", "-l", target)
      Integer(stdout.split(" ")[0])
    end
```

Usage:

```ruby
loc = is_text_file?(target) ? analyze_line_of_code(target) : nil
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      let :runner_config, Schema::BaseConfig.base
+      let :issue, object(
+        line_of_code: integer?,
+        last_commit_datetime: integer
+      )
+    end
+
+    def analyzer_version
+      @analyzer_version ||= extract_version!("wc")
```
ybiquitous, line 14: [suggestion] `wc` version is inaccurate because `git` also is used. So how about using the placeholder `0.0.0`?

```suggestion
      "0.0.0"
```

Or using the Runners version seems good:

```suggestion
      Runners::VERSION
````

File: lib/runners/processor/metrics_fileinfo.rb [hunk 21]
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    private
+    def generate_issue(path)
+      filepath = relative_path(path).to_s
+      loc = analyze_line_of_code(filepath)
+      last_commit_iso, last_commit_epoch = analyze_last_commit_datetime(filepath)
+
+      Issue.new(
+        path: relative_path(path),
```
ybiquitous, line 37: [nitpick] We can avoid `relative_path` calls twice:

```suggestion
      filepath = relative_path(path)
      loc = analyze_line_of_code(filepath.to_path)
      last_commit_iso, last_commit_epoch = analyze_last_commit_datetime(filepath.to_path)

      Issue.new(
        path: filepath,
```

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: This is a requirement from Sider service-backend system. 
(CC: @suetsuguk )

https://github.com/sider/runners/pull/1937#discussion_r561438825

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,64 @@
...
+    def analyze_line_of_code(target)
+      is_text_file?(target) ? capture3!("wc", "-l", target).then {|stdout,| Integer(stdout.split(" ")[0])} : nil
+    end
+
+    def analyze_last_commit_datetime(target)
+      last_commit = capture3!("git", "log", "-1", "--format=format:%aI|%at", target).then { |stdout,| stdout.split("|")}
+      [last_commit[0], last_commit[1]]
+    end
+
+    def is_text_file?(target)
```
ybiquitous, line 58: [nitpick] In Ruby, no use of `is_` prefix is natural:

```suggestion
    def text_file?(target)
```

lib/runners/processor/metrics_fileinfo.rb
ybiquitous: Oh, I got it, but I disagree with the following comment.

> Converting from a text value takes more cost than from a integer value.

My answer is "we cannot know until measuring the performance".
Creating `Time` instances is performed everywhere in our Rails app, so I think the cost is ignorable.

Instead, epoch time is less readable to humans, so I'm concerned that our debug will be harder.

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,64 @@
...
+      capture3!("git", "ls-files", "--eol", "--error-unmatch", target).then do |stdout, _, _|
+        stdout.split(" ")[1].match?("w/-text") ? false : true
```
ybiquitous, line 60: [question] Is this check via `git ls-files` reliable?

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: 456433442dbd01eb89789b60e72f0c0f6b3a0a98

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: Generally, file type discrimination is a difficult task. There may not be a perfect method.

But, I think that Git's file type estimation is reliable to some extent because Git uses the information when they process the committed file.
https://git-scm.com/docs/git-ls-files#Documentation/git-ls-files.txt---eol

I've tested some ambiguous cases in `binary_files`, `multi_language`, and `unknown_extension` smoke test cases. We can determine file type correctly in cases as below.
* A binary file, but having `.txt` extension. (e.g. `no_text.txt`)
* A text files not encoded in `UTF-8` but `EUC-JP`, `ISO-2022-JP`, `Shift JIS`.
* A text file having a non-well-known extension. (e.g. `foo.my_original_extension` )

NOTE: [`file(1)` command](https://linux.die.net/man/1/file) is also one of  the method. But it is not good in checking not UTF-8 encoded files.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: 694b4a877ec950708b3df2f72fe765f53cb8d749

lib/runners/processor/metrics_fileinfo.rb
ybiquitous: OK, thanks for the detailed explanation. So could you please leave a comment on the code?
The comment should describe the background that we choose `git-ls-files` to detect text files.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: > epoch time is less readable to humans, so I'm concerned that our debug will be harder.
We have a datetime information in `message` field. We can debug with the field in a trouble. 
https://github.com/sider/runners/pull/1937#discussion_r561462630

To be honest, I prefer to have a single data representation for a data. (i.e., It is better to record datetime in ISO format.)

@suetsuguk What do you think about the issue? Especially, I'd like to discuss about the cost of `Time` class creation.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: 16f13165c9ec32eec54ed081ba7251db54d7e759

lib/runners/processor/metrics_fileinfo.rb
suetsuguk: OK. I agree to record only datetime in ISO string format. And see what happens.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: cdf20d61d3bee62e5700fdfc0e936661eb11d251

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def analyze(changes)
+      Results::Success.new(
+        guid: guid,
+        analyzer: analyzer,
+        issues: changes.changed_paths.map { |path| generate_issue(path) }
+      )
+    end
+
+    private
```
ybiquitous, line 30: [nitpick] Sorry, please add an empty line between `private` and `def` 🙏 

```suggestion
    private

```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,65 @@
...
+      Integer(stdout.split(" ")[0])
+    end
+
+    def analyze_last_commit_datetime(target)
+      last_commit = capture3!("git", "log", "-1", "--format=format:%aI|%at", target).then { |stdout,| stdout.split("|")}
+      [last_commit[0], last_commit[1]]
+    end
+
+    def text_file?(target)
+      capture3!("git", "ls-files", "--eol", "--error-unmatch", target).then do |stdout, _, _|
```
ybiquitous, line 60: [suggestion] How about these improvements?

- Set `trace_stdout: false` to suppress verbose traces.
- One call `git ls-files`.

```ruby
    def search_text_files(paths)
      # NOTE: Use `each_slice` to avoid arguments length limit.
      paths.each_slice(10000) do |targets|
        stdout, _ = capture3!("git", "ls-files", "--eol", "--error-unmatch", "--", *targets.map(&:to_path), trace_stdout: false)
        stdout.each_line(chomp: true) do |line|
          fields = line.split(" ")
          type = fields[1] or raise
          file = fields[3] or raise
          if type != "w/-text"
            text_files << file
          end
        end
      end
    end

    def text_files
      @text_files ||= Set[]
    end

    def text_file?(target)
      text_files.include?(target)
    end
```

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: 752c607dafb0541a91d7583fd9a1ff4bd15dd7f3

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: b7743ea9d2ddb1a1fc844aef34354ff653bf52f2

lib/runners/processor/metrics_fileinfo.rb
ybiquitous: That's great! 👍

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: Thanks a lot for your nice suggestion.
fix: 6498e360e3e71c11d5bc32fb6ee86111a5de578c

ybiquitous: One question: can we run `git ls-files` with no given files? If we can do it, the code will be much simpler.

gnakagaw: > can we run git ls-files with no given files?

Yes. We can get information about all files in the target repository regardless of `Changes` information. 

The result of `search_text_files` contains the files not changed  at the target commit. But it is no problem because we use `Changes` information to generate the final result.

I'll refactor the code to do so.

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: a7e20e8c5a5db9bb129111837ef3fc2a7500660c

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: 5561e4d23d0dd9de31a35af242dccc81e4763cac

ybiquitous: Good refactoring 👍

lib/runners/processor/metrics_fileinfo.rb
suetsuguk: I feel something wrong with this `[0]` at [this line](https://github.com/gnakagaw/runners/blob/5561e4d23d0dd9de31a35af242dccc81e4763cac/lib/runners/processor/metrics_fileinfo.rb#L56) because other two calls in this file use `stdout`. But, it's up to you.

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,115 @@
...
+      last_commit = analyze_last_commit_datetime(filepath.to_path)
+
+      Issue.new(
+        path: filepath,
+        location: nil,
+        id: "metrics_fileinfo",
+        message: "#{filepath}: loc = #{loc}, last commit datetime = #{last_commit}",
+        object: {
+          line_of_code: loc,
+          last_commit_datetime: last_commit
```
ybiquitous, line 44: [IMO] We often use naming like `*ed_at` according to Rails:

https://github.com/sider/runners/blob/0dc3eafffd38bba59d9fbb8394ebb5aa47eb4ff9/lib/runners/schema/trace.rb#L6

So it sounds readable. But this is my personal opinion, you can ignore it.

```suggestion
          last_committed_at: last_commit
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,115 @@
...
+    #
+    # We've tested some ambiguous cases in binary_files, multi_language, and unknown_extension smoke test cases.
+    # We can determine file type correctly in cases as below.
+    #  * A plain text file having various extensions (.txt, .rb, .md, etc..)
+    #  * A binary file having various extensions (.png, .data, etc...)
+    #  * A binary file, but having .txt extension. (e.g. no_text.txt)
+    #  * A text files not encoded in UTF-8 but EUC-JP, ISO-2022-JP, Shift JIS.
+    #  * A text file having a non-well-known extension. (e.g. foo.my_original_extension )
+    def search_text_files
+      Set[].tap do |result|
```
ybiquitous, line 102: [suggestion] Can we simplify it more? 👇 

```suggestion
    def text_files
      @text_files ||= Set[].tap do |result|
```

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: I should respect conventions of project. I've fixed them. 
fix: f9e857750285da53ce815047aad35c9e0f5eb781 and 75757e08bd8622181635335f88c6716066854000

@suetsuguk I've changed  `object` schema in `Issue` class as above. Please check that. 🙏

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: Uniformity of code is fundamental. fix: 4fa932e3ff166fea19a90bdbb08b7d51d1fa540f

lib/runners/processor/metrics_fileinfo.rb
suetsuguk: Confirmed. Changes at this moment doesn't affect Sideci.

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,113 @@
...
+        object: {
+          line_of_code: loc,
+          last_committed_at: last_commit
+        },
+        schema: Schema.issue
+      )
+    end
+
+    def analyze_line_of_code(target)
+      stdout, _ = capture3!("wc", "-l", target)
```
ybiquitous, line 51: [question] Does it make sense to reduce calls of `wc -l` as follows?

```ruby
file_loc = {}
target_files.each_slice(10000) do |targets|
  stdout, _ = capture3!("wc", "-l", *targets)
  lines = stdout.lines(chomp: true)
  lines.pop # remove summary
  lines.each do |line|
    loc, file = line.split(" ")
    file_loc[file] = Integer(loc)
  end
end
```

gnakagaw: I've restructured analysis flow to reduce call external commands (i.e. `git` and `wc`). 
fix: 4e69b7dcdb011e8a657a753323b8e67684351a52

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,141 @@
...
+      Runners::VERSION
+    end
+
+    # This analyser use git metadata (.git/).
+    def use_git_metadata_dir?
+      true
+    end
+
+    def analyze(changes)
+      targets = changes.changed_paths.map { |x| relative_path(x)}
```
ybiquitous, line 23: [suggestion] Maybe `relative_path` is needless here:

```suggestion
      targets = changes.changed_paths
```

`Changes` does it already:

https://github.com/sider/runners/blob/0dc3eafffd38bba59d9fbb8394ebb5aa47eb4ff9/lib/runners/changes.rb#L80

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,141 @@
...
+      @last_committed_at ||= {}
+    end
+
+    def analyze_last_committed_at(targets)
+      targets.each do |target|
+        path = target.to_path
+        stdout, _ = capture3!("git", "log", "-1", "--format=format:%aI", path)
+        last_committed_at[path] = stdout
+      end
+      nil
```
ybiquitous, line 86: [nitpick] Maybe `nil` is needless:

```suggestion
```

lib/runners/processor/metrics_fileinfo.rb
ybiquitous: Nice refactoring!

sig/runners/processor/metrics_fileinfo.rbs
```diff
@@ -0,0 +1,25 @@
...
+
+    def generate_issue: (Pathname)-> Issue
+
+    def lines_of_code: () -> Hash[String, Integer]
+
+    def analyze_lines_of_code: (Set[Pathname]) -> nil
+
+    def last_committed_at: () -> Hash[String, String]
+
+    def analyze_last_committed_at: (Set[Pathname]) -> Set[Pathname]
```
ybiquitous, line 19: [suggestion] `void` is available if we don't want to worry about a return value.

```suggestion
    def analyze_lines_of_code: (Set[Pathname]) -> void

    def last_committed_at: () -> Hash[String, String]

    def analyze_last_committed_at: (Set[Pathname]) -> void
```

gnakagaw: This is exactly what I have wanted! thx!

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,139 @@
...
+
+    def analyze(changes)
+      analyze_last_committed_at(changes.changed_paths)
+      analyze_lines_of_code(changes.changed_paths)
+
+      Results::Success.new(
+        guid: guid,
+        analyzer: analyzer,
+        issues: changes.changed_paths.map { |path| generate_issue(path) }
+      )
```
ybiquitous, line 30: [suggestion] Maybe we can reduce calling `Pathname#to_path` as follows:

```suggestion
      target_files = changes.changed_paths.map(&:to_path)
      analyze_last_committed_at(target_files)
      analyze_lines_of_code(target_files)

      Results::Success.new(
        guid: guid,
        analyzer: analyzer,
        issues: target_files.map { |path| generate_issue(path) }
      )
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,139 @@
...
+    end
+
+    private
+
+    def generate_issue(path)
+      filename = path.to_path
+      loc = lines_of_code[filename]
+      commit = last_committed_at[filename]
```
ybiquitous, line 38: [suggestion] How about using `fetch` to avoid nullable `commit`?

```suggestion
      commit = last_committed_at.fetch(filename)
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      Issue.new(
+        path: path,
+        location: nil,
+        id: "metrics_fileinfo",
+        message: "#{filename}: loc = #{loc}, last commit datetime = #{commit}",
```
ybiquitous, line 44: [question] If `loc` is `nil`, is the following message expected?

```
foo.rb: loc = , last commit datetime = 2020-01-01T12:09:02Z
```

sig/runners/processor/metrics_fileinfo.rbs
gnakagaw: fix: 81487675343fde5c5952f8a8859cd5193d4af252

ybiquitous:
Oh, CI fails. I think this is another problem. Please wait.

ybiquitous:
I've fixed the CI failure via #1975. Please merge `master` to this branch. 🙏

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: fix: e2466ef97f58e767838084d349323953307597a0

lib/runners/processor/metrics_fileinfo.rb
gnakagaw: Thx. I'll add a message when loc is nil
e.g. `image.png: loc = (no loc info), last commit datetime = 2021-01-01T13:00:00+09:00`

fix: 964f7b2a09b199c3903ef10c14456091051a17ea

ybiquitous: I prefer `loc = (no info)` that is simpler but looks good 👍

ybiquitous:
LGTM 👍

Would you like to include this PR for the next version of Runners?

ybiquitous (approved):
👍 Please update changelog.

gnakagaw:
Thank you for kind and long-term reviewing 🙏

PR #1955, opened 2021-01-20
Improve behavior about `Changes` class

ybiquitous:

This change omits the trace message when changed files are not present.
Because it may confuse users or us to output the log message despite the process is not actually performed.

None.

### DIFF
--- a/lib/runners/changes.rb
+++ b/lib/runners/changes.rb
@@ -59,7 +59,7 @@ module Runners
       end
     end
 
-    def self.calculate(working_dir:)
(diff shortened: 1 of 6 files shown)


lib/runners/changes.rb
```diff
@@ -59,7 +59,7 @@ def include?(issue)
       end
     end
 
-    def self.calculate(working_dir:)
+    def self.all(working_dir:)
```
ybiquitous, line 62: [note] I think the `calculate` name is not suitable because this method does not actually anything.

PR #1905, opened 2021-01-06
Bump steep from 0.37.0 to a227bcc

ybiquitous:
> Explain a summary, purpose, or background for this change.

- ~~https://github.com/soutaro/steep/blob/v0.39.0/CHANGELOG.md~~
- ~~https://github.com/soutaro/steep/compare/v0.37.0...v0.39.0~~
- https://github.com/soutaro/steep/blob/a227bcc62ad378fe88d2dfed8c600861988ba376/CHANGELOG.md
- https://github.com/soutaro/steep/compare/v0.37.0...a227bcc62ad378fe88d2dfed8c600861988ba376

*Note that the latest version [0.39.0](https://github.com/soutaro/steep/releases/tag/v0.39.0) causes many errors, so this bumps to the `master` version.*

This change also bumps `rbs` from 0.20.1 to 1.0.0:

- https://github.com/ruby/rbs/blob/v1.0.0/CHANGELOG.md
- https://github.com/ruby/rbs/compare/v0.20.1...v1.0.0

> Link related issues or pull requests.

None.

> Check the following.

- [ ] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.

PR #1962, opened 2021-01-21
add type signature of time class from rbs v1.0.0

gnakagaw:

Add type signature of [`class Time`](https://docs.ruby-lang.org/en/3.0.0/Time.html). It is required to use `class Time` in a runner processer with type checking by RBS.

NOTE: It is better to upgrade [`rbs` gem](https://rubygems.org/gems/rbs?locale=en) ideally. But we cannot do that now because [`steep`](https://github.com/soutaro/steep) has not supported them yet. We should remove this code when we are ready to update `rbs` gem.

#1825

### DIFF
--- a/sig/time.rbs
+++ b/sig/time.rbs
@@ -1,3 +1,329 @@
+# https://raw.githubusercontent.com/ruby/rbs/v1.0.0/stdlib/time/0/time.rbs
+
 class Time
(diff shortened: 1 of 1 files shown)


ybiquitous (approved):
👍 

Note: we wait for Steep update. See #1905

PR #2007, opened 2021-02-02
Use `Processor#analyzer_github` to reduce duplication

ybiquitous:

The `Runners::Processor#analyzer_github` method can reduce duplication in some `Processor` classes.

None.

### DIFF
--- a/lib/runners/processor/scss_lint.rb
+++ b/lib/runners/processor/scss_lint.rb
@@ -32,7 +32,7 @@ module Runners
     def setup
       add_warning_for_deprecated_linter(
         alternative: analyzers.name(:stylelint),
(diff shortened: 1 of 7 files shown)


lib/runners/processor/scss_lint.rb
```diff
@@ -32,7 +32,7 @@ def analyzer_version
     def setup
       add_warning_for_deprecated_linter(
         alternative: analyzers.name(:stylelint),
-        ref: "https://github.com/sds/scss-lint/blob/master/README.md#notice-consider-other-tools-before-adopting-scss-lint",
+        ref: "#{analyzer_github}#readme",
```
ybiquitous, line 35: [note] The deprecation notice is at the top of the readme, so `#readme` is enough.

Issue #1720, opened 2020-12-10
Support Slim-Lint


ybiquitous:
## Information

- Website: https://github.com/sds/slim-lint
- Language: [Slim](https://github.com/slim-template/slim) (Ruby template engine)

## Why do you need the tool?

Slim is a Ruby template engine that is just as popular as [HAML](https://github.com/haml/haml).
Although a bit old, this [comparison on Gist](https://gist.github.com/pboling/af8a0c9a095820ef8434) is helpful.

Since we have already supported [HAML-Lint](https://github.com/sds/haml-lint), I think it makes sense to support also Slim-nLint.

https://github.com/sider/runners/blob/a73b41486ce8aca4e99fe990246898dc5c039084/analyzers.yml#L63

## Additional context

Like HAML-Lint, Slim-Lint also supports the integration with RuboCop.

## See also

None.

PR #2009, opened 2021-02-03
[HAML-Lint] Improve issue ID and links for RuboCop

ybiquitous:

HAML-Lint uses internally RuboCop. But HAML-Lint issue's ID is always `RuboCop`.
This change converts it to `RuboCop:{cop_name}` such as `RuboCop:RuboCop:Lint/UselessAssignment`
and adds a link to the RuboCop cop documentation.

I think this change will allow users to handle HAML-Lint issues more easily.

Also, this change extracts common methods for RuboCop to a new module: `Runners::RuboCopUtils`.

This refactoring should be helpful for Slim-Lint (#1720).


test/smokes/haml_lint/expectations.rb
```diff
@@ -78,8 +81,7 @@
   type: "success",
   issues: [
     {
-      message:
-        "Lint/UselessAssignment: Useless assignment to variable - `unused_variable`. (https://rubystyle.guide#underscore-unused-vars)",
+      message: "Useless assignment to variable - `unused_variable`. (https://rubystyle.guide#underscore-unused-vars)",
       links: %w[https://github.com/sds/haml-lint/blob/v0.34.0/lib/haml_lint/linter#rubocop],
       id: "RuboCop",
```
ybiquitous, line 86: [note] When `DisplayCopNames: false` in `.rubocop.yml`, an issue ID is still `RuboCop`.

ybiquitous:
The CI failure should be fixed via #2011.

yykamei (approved):
I see. This change is great :+1: 

I think it's a breaking change because some IDs will change. Especially regarding Sider, some closed issues of HAML-Lint might be reopened while applying this new version of Runners to Sider.

ybiquitous:
Surely. Users may be temporarily inconvenienced, but there are not so many `RuboCop` issues recently when I look into them.

I think this is tolerable, right?

yykamei:
Yes, I agree. It's a breaking change, but virtually, there seem to be few problems. I believe.

PR #2010, opened 2021-02-03
Verify gem installation in Dockerfiles

ybiquitous:

This change aims to verify a gem installation in Dockerfiles.
`gem info` does **not** exit with an error even if a gem is not installed,
so this uses `grep` for the verification.

None.


ybiquitous:
CI passed 🚀

PR #2013, opened 2021-02-03
Add methods to `Runners::RuboCopUtils`

ybiquitous:

Added methods: `official_rubocop_plugins`, `third_party_rubocop_plugins`

Follow up of PR #2009

### DIFF
--- a/lib/runners/processor/haml_lint.rb
+++ b/lib/runners/processor/haml_lint.rb
@@ -1,5 +1,3 @@
-require_relative 'rubocop'
-
 module Runners
   class Processor::HamlLint < Processor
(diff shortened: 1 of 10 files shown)


lib/runners/processor/haml_lint.rb
```diff
@@ -1,5 +1,3 @@
-require_relative 'rubocop'
```
ybiquitous, line 1: [note] I think it is not good to depend on `Processor::RuboCop` directly. We now have `RuboCopUtils` instead.

ybiquitous:
CI passed!

PR #2020, opened 2021-02-04
[RuboCop] Set up default config only when no user config

### LINKED ISSUE #2018: RuboCopUtils#setup_default_rubocop_config should follow user-specified config file
`RuboCopUtils#setup_default_rubocop_config` currently checks the existence of `./rubocop.yml` in the `current_dir`, but some users may configure it as an alternative file. So, this method should support such a case. In addition, some linters depend on RuboCop and have their own configuration file (e.g., HAML-Lint takes care of `haml-lint.yml`). I think `#setup_default_rubocop_config` should also consider this case.

https://github.com/sider/runners/blob/00e15ecfcc554084607909759b4cc8f0bfd90a66/lib/runners/rubocop_utils.rb#L66-L74

ybiquitous:

When a user specifies a custom config file via `sider.yml`,

```yaml
linter:
  rubocop:
    config: my-rubocop.yml
```

the default config currently is copied to `.rubocop.yml` in the user's repository on staring analysis:
<https://github.com/sider/runners/blob/9239a21e2c38c69d3b03231fbe7201cb48c1bc14/images/rubocop/default_rubocop.yml>

But this copied config is ignored (because `rubocop --config my-rubocop.yml` is executed)
and `meowcop` is installed unnecessarily.
This change aims to fix that needless behavior.

Note that no smoke test cases are added because the following case satisfies them already:
<https://github.com/sider/runners/blob/9239a21e2c38c69d3b03231fbe7201cb48c1bc14/test/smokes/rubocop/with_safe_cops/sideci.yml#L5>

Fix #2018

### DIFF
--- a/lib/runners/processor/rubocop.rb
+++ b/lib/runners/processor/rubocop.rb
@@ -36,7 +36,7 @@ module Runners
       add_warning_if_deprecated_options
 
       default_gems = default_gem_specs(GEM_NAME)
-      if setup_default_rubocop_config
+      if !rubocop_config_file && setup_default_rubocop_config
         # NOTE: The latest MeowCop requires usually the latest RuboCop.
         #       (e.g. MeowCop 2.4.0 requires RuboCop 0.75.0+)
         #       So, MeowCop versions must be unspecified because the installation will fail when a user's RuboCop is 0.60.0.
@@ -70,7 +70,7 @@ module Runners
         "--out=#{report_file}",
 
         *rails_option,
-        *config_file,
+        *rubocop_config_file_option,
         *safe,
       )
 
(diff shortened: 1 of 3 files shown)


lib/runners/processor/rubocop.rb
```diff
@@ -144,9 +144,12 @@ def rails_option
       end
     end
 
-    def config_file
```
ybiquitous, line 147: [note] `config_file` seems ambiguous to me (`sider.yml`), so I've renamed it.

yykamei (approved):

yykamei (approved):

PR #2032, opened 2021-02-08
Save and check `steep_expectations.yml`

ybiquitous:

Steep 0.41.0 has introduced `--with-expectations` and `--save-expectations` options.
They should improve our management of Steep errors.

If we encounter new errors of Steep in the future, we should run `rake steep:check'[true]'`.

See also <https://github.com/soutaro/steep/pull/303>.

Related to #2031


yykamei (approved):
LGTM

PR #2044, opened 2021-02-10
add an instruction to install RBS collection for type checking

gnakagaw:

Developers are required to install [`gem_rbs_collection`](https://github.com/ruby/gem_rbs_collection) for type checking with `steep`. But there is no information about that in `README.md`. This PR adds an instruction to install `git submodule` .

none

- ~[ ] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.~ This is a PR for documentation. No need to update CHANGELOG.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,13 @@ Then, install Git hooks via [Lefthook](https://github.com/Arkweid/lefthook):
 $ bundle exec lefthook install
 ```
 
+Install [RBS collection for RubyGems](https://github.com/ruby/gem_rbs_collection) via [git submodule](https://git-scm.com/docs/git-submodule):
+
+```shell-session
+$ git submodule init
+$ git submodule update
+```
+
 Last, run the following command to show available commands in the project:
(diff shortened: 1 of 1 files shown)


ybiquitous (commented):
I think a solution in PR #2045 is better. What do you think?

README.md
```diff
@@ -191,6 +198,8 @@ To update the third-party gems' RBS, run:
 $ bundle exec rake rbs:update_gems
 ```
 
+If you get a `Unexpected error reported` error, ensure that you've got RBS for gems. See [Setup](#setup) section.
```
ybiquitous, line 201: This error message may be outdated in the future, so I think this description may be confusing for some developers.

gnakagaw:
> I think a solution in PR #2045 is better. What do you think?

I've checked #2045. It's a better solution than this PR. I'll close this PR.

PR #2014, opened 2021-02-03
[Slim-Lint] New support

ybiquitous:
> Explain a summary, purpose, or background for this change.

This change newly supports Slim-Lint.
See also [how-to-write-a-new-runner.md](https://github.com/sider/runners/blob/ba023352c89e56e74f37db3738d2369d50efe2ca/docs/how-to-write-a-new-runner.md).

> Link related issues or pull requests.

Close #1720

> Check the following if needed.

- [x] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.
- [x] Add a new tool to [`analyzers.yml`](../analyzers.yml).
- [x] Run `bundle exec rake readme:generate` and commit the updated [`README.md`](../README.md).
- [x] Add a new tool to the [Dependabot configuration](../.github/dependabot.yml) if needed.
- [x] Provide the tool's options that users can customize.
- [x] Write smoke test cases. E.g. all options, successful end, warnings, errors, etc.
- [x] Add a [CI setting](../.github/workflows/build.yml).
- [x] Write a [document](https://github.com/sider/sider-docs) and link the pull request: sider/sider-docs#558

PR #2046, opened 2021-02-10
[Slim-Lint] Set up recommended ruleset

ybiquitous:

Slim-Lint's default value of the rule `LineLength` is too strict:
<https://github.com/sds/slim-lint/blob/v0.20.2/config/default.yml>

So, this change aims to loosen the value via our recommended ruleset.

Follow-up of #2014


ybiquitous:
@yykamei Does this change make sense? When I test Slim-Lint in our project, too many `LineLength` issues are detected. 😓

ybiquitous:
(It seems we need to prepare such recommended ruleset also for HAML-Lint... 🤔 )

yykamei (approved):
LGTM

ybiquitous:
Thanks. What do you think about <https://github.com/sider/runners/pull/2046#issuecomment-776489055>?

ybiquitous:
OK, I'll address HAML-Lint also.

PR #2063, opened 2021-02-15
[GolangCI-Lint] Fix too short timeout value

ybiquitous:

We now can change timeout via `ENV['RUNNERS_TIMEOUT']`, but a fixed value is set in the GolangCI-Lint runner.
This fixed value is too short, so this change extends it to high enough.

See also:
- https://github.com/sider/runners/blob/a0e468fcd3bc8b23ea47cafb31482860483e14dc/images/docker-entrypoint.sh#L39
- https://github.com/sider/runners/blob/a0e468fcd3bc8b23ea47cafb31482860483e14dc/lib/runners/shell.rb#L124

None.

### DIFF
--- a/lib/runners/processor/golangci_lint.rb
+++ b/lib/runners/processor/golangci_lint.rb
@@ -96,10 +96,9 @@ module Runners
         opts << "--color=never"
         opts << "--concurrency=2"
 
-        # NOTE: The value should be less than the top-level timeout value.
-        #
-        # @see https://github.com/sider/runners/blob/1231de0ca047c7a23449ab1a7bb0751f39f16643/images/Dockerfile.end.erb#L11
-        opts << "--timeout=15m"
+        # NOTE: Set `--timeout` to prevent a user-defined timeout value via `.golangci.yml`.
+        #       This value should be less than the top-level timeout value `ENV['RUNNERS_TIMEOUT']`.
+        opts << "--timeout=10h"
 
         opts << "--tests=#{config_linter[:tests]}" unless config_linter[:tests].nil?
         opts << "--config=#{path_to_config}" if path_to_config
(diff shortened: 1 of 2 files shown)


ybiquitous:

yykamei (approved):
LGTM

lib/runners/processor/golangci_lint.rb
```diff
@@ -96,10 +96,9 @@ def analyzer_options
         opts << "--color=never"
         opts << "--concurrency=2"
 
-        # NOTE: The value should be less than the top-level timeout value.
-        #
-        # @see https://github.com/sider/runners/blob/1231de0ca047c7a23449ab1a7bb0751f39f16643/images/Dockerfile.end.erb#L11
-        opts << "--timeout=15m"
+        # NOTE: Set `--timeout` to prevent a user-defined timeout value via `.golangci.yml`.
+        #       This value should be less than the top-level timeout value `ENV['RUNNERS_TIMEOUT']`.
```
yykamei, line 100: This comment looks unnecessary (and incorrect).
```suggestion
```

yykamei (commented):
LGTM

lib/runners/processor/golangci_lint.rb
ybiquitous: I think it needed to explain why this value is `10h`. What idea do you have?

ybiquitous: Oh, `less` is incorrect as you said. So how about this?

```suggestion
        #       This value should be greater than the top-level timeout value `ENV['RUNNERS_TIMEOUT']`.
```

yykamei: I want to suggest adding "enough."

```suggestion
        #       This value should be enough greater than the top-level timeout value `ENV['RUNNERS_TIMEOUT']`.
```

ybiquitous: nice 👍

yykamei (approved):
:+1:

ybiquitous:
Thanks 😊

PR #2099, opened 2021-02-26
Remove *removed* Go tools from `sider.yml`

ybiquitous:

We have removed `golint`, `go_vet`, and `gometalinter` already, but they have been supported in `sider.yml` for backward compatibility.

I think it reasonable to remove the compatibility now.
(Maybe there remains in some `sider.yml` files but it should be few)

See also our [blog post](https://siderlabs.com/blog/siders-go-meta-linter-support-will-be-discontinued-at-the-end-of-may-2020-5985a2c5b5cb/).

None.


ybiquitous:
@yykamei Could you give me your opinion about this change, please?

yykamei (approved):
Thank you :put_litter_in_its_place:

yykamei:
I think it's time to proceed with something like this change :+1:

ybiquitous:
Thanks. 👍 
We may need to care about analysis failures due to this change in the next release.

PR #2122, opened 2021-03-01
[Metrics FileInfo] refactoring processor

### LINKED ISSUE #2118: [Metrics FileInfo] refactoring Metrics FileInfo runner
Here is a just list for refactoring..
* supress stdout tracing for `wc -l` 
  * https://github.com/sider/runners/blob/master/lib/runners/processor/metrics_fileinfo.rb#L67
* parallelize `git log -1` calling
  * Process creation is one of special cases that Ruby VM release Global Interpreter Lock (GIL). So, it may be possible to run `git log -1` in parallel.

[EDIT]
I found a bug in `Metrics FileInfo`. We cannot "**/hoge" format in `fnmatch?` function.  https://github.com/sider/runners/blob/master/lib/runners/processor/metrics_fileinfo.rb#L28
ref: https://docs.ruby-lang.org/ja/latest/method/Pathname/i/fnmatch.html

gnakagaw:
Fix #2118

### DIFF
--- a/lib/runners/processor/metrics_fileinfo.rb
+++ b/lib/runners/processor/metrics_fileinfo.rb
@@ -85,7 +85,7 @@ module Runners
 
     def analyze_last_committed_at(targets)
       trace_writer.message "Analyzing last commit time..." do
-        Parallel.each(targets, in_threads: Parallel.processor_count) do |target|
+        Parallel.each(targets, in_threads: 2) do |target|
           stdout, _ = capture3!("git", "log", "-1", "--format=format:%aI", "--", target, trace_stdout: false, trace_command_line: false)
           last_committed_at[target] = stdout
         end
--- a/lib/runners/processor/metrics_fileinfo.rb
+++ b/lib/runners/processor/metrics_fileinfo.rb
@@ -25,7 +25,7 @@ module Runners
       capture3!("git", "commit-graph", "write", "--reachable", "--changed-paths", "--no-progress")
 
       target_files = Pathname.glob("**/*", File::FNM_DOTMATCH).filter do |path|
-        path.file? && !path.fnmatch?("**/.git/*", File::FNM_DOTMATCH)
+        path.file? && !path.fnmatch?(".git/*", File::FNM_DOTMATCH) && !path.fnmatch?("*/.git/*", File::FNM_DOTMATCH)
       end
 
       analyze_last_committed_at(target_files)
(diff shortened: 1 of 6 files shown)


lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -25,7 +25,7 @@ def analyze(changes)
       capture3!("git", "commit-graph", "write", "--reachable", "--changed-paths", "--no-progress")
 
       target_files = Pathname.glob("**/*", File::FNM_DOTMATCH).filter do |path|
-        path.file? && !path.fnmatch?("**/.git/*", File::FNM_DOTMATCH)
+        path.file? && !path.fnmatch?(".git/*", File::FNM_DOTMATCH) && !path.fnmatch?("*/.git/*", File::FNM_DOTMATCH)
```
ybiquitous, line 28: [suggestion] How about this? ⬇️ 

```suggestion
        path.file? && !path.fnmatch?(".git/**", File::FNM_DOTMATCH)
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -85,7 +85,7 @@ def last_committed_at
 
     def analyze_last_committed_at(targets)
       trace_writer.message "Analyzing last commit time..." do
-        targets.each do |target|
+        Parallel.each(targets, in_threads: Parallel.processor_count) do |target|
```
ybiquitous, line 88: [suggestion] `Parallel.processor_count` seems unstable because of `Etc.nprocessors`:

https://github.com/grosser/parallel/blob/875c9f472e6da3450219a7ba2d29bf430f23adfe/lib/parallel/processor_count.rb#L7-L9

How about explicitly specifying the value like this? ⬇️ 

```suggestion
        Parallel.each(targets, in_threads: 2) do |target|
```

ybiquitous:
The change looks good. Have you tested the performance improvement for so many files?

gnakagaw:
@ybiquitous 
> The change looks good. Have you tested the performance improvement for so many files?

I've tested with https://github.com/git/git
before: 55.5 seconds
after: 40.0 seconds

We can accelerate to increase `in_threads` option. For example, it took 29.7 seconds with `in_threads: 8`.

ybiquitous:
Try `in_threads: 8`!

ybiquitous (approved):
LGTM 👍 
Please merge it when the CI passes.

PR #2130, opened 2021-03-02
[Flake8] Add `parallel` option

### LINKED ISSUE #1910: [Flake8] Add --jobs option
Add `linter.flake8.jobs` option to `sider.yml`

## What it this ?
support `--jobs` option for Flake 8
> -j JOBS, --jobs=JOBS  Number of subprocesses to use to run checks in parallel. This is ignored on Windows. The default, "auto", will auto-detect the number of processors available to use. (Default: auto)

https://flake8.pycqa.org/en/latest/manpage.html#options

## Why?
Flake 8 executes a analysis in parallel. In some case, this makes the analysis stuck. (i.e., timeout). With the option, we can disable parallel execution of Flake8 analysis.

ybiquitous:

This change adds the new `parallel` option which allows boolean values (default: `true`).
I think we should not provide the `jobs` option because users can specify an invalid value like `jobs: 10000`.

Close #1910


ybiquitous:
@gnakagaw Could you give me any feedback, please?

lib/runners/processor/flake8.rb
```diff
@@ -48,12 +50,15 @@ def setup
...
     def analyze(changes)
+      parallel = [true, nil].include?(config_linter[:parallel])
+
       capture3!(
         analyzer_bin,
         "--exit-zero",
         "--output-file", report_file,
         "--format", OUTPUT_FORMAT,
         "--append-config", IGNORED_CONFIG_PATH,
+        "--jobs", parallel ? "2" : "1",
```
gnakagaw, line 61: @ybiquitous 
```suggestion
        "--jobs", parallel ? "auto" : "1",
```
The default value for Flake8 `jobs` option is `auto`; Flake8 determine the optimal parallelism with the option. It is better that we set `auto` when `parallel` option is `true` in my opinion.
https://flake8.pycqa.org/en/latest/manpage.html#options

lib/runners/processor/flake8.rb
```diff
@@ -38,6 +39,7 @@ def self.config_example
           - flake8-builtins==1.4.1
           - flake8-docstrings>=1.4.0
           - git+https://github.com/PyCQA/flake8-import-order.git@51e16f33065512afa1a85a20b2c2d3be768f78ea
+        parallel: false
```
gnakagaw, line 42: The default value is `true` according to this PR description. Is this ok?
> This change adds the new parallel option which allows boolean values (default: true).

lib/runners/processor/flake8.rb
ybiquitous: Make sense. 👍 b4a67019

lib/runners/processor/flake8.rb
ybiquitous: Yes, this is just an example.

gnakagaw (approved):
LGTM

ybiquitous:
Thanks. I'll merge this when the CI passes. 🚀

Issue #2095, opened 2021-02-25
Remove deprecated `linter.{id}.options`


ybiquitous:
If we find that these `options` are not used, we can remove them:

https://github.com/sider/runners/blob/90a9a49e96fdee49cf74895df8784d618205fb16/lib/runners/processor.rb#L198-L216

*Note: They have been removed already from the [documentation](https://help.sider.review).*

PR #2132, opened 2021-03-03
[PHPCS] Re-consider options

### LINKED ISSUE #1953: [PHPCS] Add --parallel option
Add `linter.code_sniffer.parallel` option to `sider.yml`

## What is this ?
support [`--parallel` option for phpcs](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage#getting-help-from-the-command-line)

## Why?
The options is useful when we analyze a large repository or use a cost-consuming phpcs plugin. Note that this option should be opt-in because parallel processing sometimes causes troubles such as deadlock.

### LINKED ISSUE #2034: [PHPCS] Add `target` option
I propose to add a new option `target` to the PHP_CodeSniffer runner.
The reason is that users don't need to remember less about `sider.yml` because many other runners support the `target` option.

The new option `target` will be an alias of the `dir` option at first.

See also:

https://github.com/sider/runners/blob/5c97f2c2208895c7d3fd331f85e07268ec3e3661/lib/runners/processor/code_sniffer.rb#L11

### LINKED ISSUE #2085: [PHPCS] Support array type for `standard` option
The `--standard` flag allows a comma-separated list. See below:

> Multiple coding standards can be checked at the same time by passing a list of comma separated standards on the command line. A mix of external and installed coding standards can be passed if required.
>
> ```
> $ phpcs --standard=PEAR,Squiz,/path/to/MyStandard /path/to/code/myfile.inc
> ```

https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage#specifying-a-coding-standard

Currently, only `string` is supported:

https://github.com/sider/runners/blob/d8ba62d68f287b4a359a96f2023a914aa69f167f/lib/runners/processor/code_sniffer.rb#L12

ybiquitous:

- Add the `parallel` option (disabled by default).
- Add the `target` option (alias for `dir`).
- Make the `target` option receive an array.
- Make the `standard` option receive an array.
- Remove the deprecated option `linter.code_sniffer.options`.

Close #1953
Close #2034
Close #2085
Part of #2095


yykamei (approved):
LGTM
It makes sense :+1:

ybiquitous:
@yykamei Thanks. Removing `linter.code_sniffer.options` may affect few repositories, but I think it reasonable because the option has been deprecated for a long time.

ybiquitous:
@gnakagaw Could you please confirm this change just in case?

gnakagaw (approved):
LGTM 🚀

PR #2146, opened 2021-03-08
[Metrics Code Clone] Report the total number of duplicated lines

suetsuguk:
The total number of duplicated lines is used to measure duplication ratio for each file.

### DIFF
--- a/lib/runners/processor/metrics_codeclone.rb
+++ b/lib/runners/processor/metrics_codeclone.rb
@@ -7,6 +7,7 @@ module Runners
 
       let :issue, object(
         clones: integer,
+        'total-clone-lines': integer,
       )
     end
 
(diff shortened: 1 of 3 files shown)


lib/runners/processor/metrics_codeclone.rb
```diff
@@ -7,6 +7,7 @@ class Processor::MetricsCodeClone < Processor
 
       let :issue, object(
         clones: integer,
+        'total-clone-lines': integer,
```
gnakagaw, line 10: IMHO, We should not use  a hyphen (-) in JSON. It causes parse troubles in some cases. 
https://www.google.com/search?q=json+key+hyphen&oq=json+key+hyphen

Of course, we can avoid the problems with our careful coding even a JSON key contains hyphens. But, it is better to avoid such problematic key name as much as possible.

suetsuguk: Fixed. commit c9f9c265ff443f20aee5508dc6cccc2d5abac61e

gnakagaw (approved):
LGTM 🚀

suetsuguk:
Thank you for taking your time.

PR #2194, opened 2021-03-18
Secure `pip install`

ybiquitous:

This change aims to make `pip install` more secure via the `--only-binary` option.
(see <https://pip.pypa.io/en/stable/reference/pip_install/#install-only-binary>)

I believe any packages do not need source compilation for us.

None.


yykamei (approved):
LGTM

ybiquitous:
See also <https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-800639587>

PR #2191, opened 2021-03-17
Introduce common option `target`

ybiquitous:
> Explain a summary, purpose, or background for this change.

This change introduces a new option `linter.<id>.target`. The option is an alias for the existing option (non-breaking).

> Refer related issues or pull requests (e.g. `Close #123` or `None`).

Close #2096

> Check the following if needed.

- [x] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.

PR #2195, opened 2021-03-18
[PoC] Add `linter.<id>.target` option into base

ybiquitous:

This is just a PoC.

Follow-up of #2191

### DIFF
--- a/lib/runners/processor/coffeelint.rb
+++ b/lib/runners/processor/coffeelint.rb
@@ -7,6 +7,7 @@ module Runners
 
       # @type self: SchemaClass
       let :config, npm(
+        target: false,
         file: string?,
       )
     end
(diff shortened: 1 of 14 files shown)


lib/runners/processor/coffeelint.rb
```diff
@@ -7,6 +7,7 @@ class Processor::Coffeelint < Processor
 
       # @type self: SchemaClass
       let :config, npm(
+        target: false,
         file: string?,
       )
```
ybiquitous, line 12: @yykamei What do you think about this code? I feel not intuitive so much.

ybiquitous:
There are the following cases, so it seems hard to write code to handle well such cases... 🤔 

---

There are linters that:
- support multiple `target`
- support only single `target`
- don't support any `target`

lib/runners/processor/coffeelint.rb
yykamei: Thank you. I think so.

ybiquitous:
@yykamei Can I close this PR if you don't mind? Of course, you can try it again.

yykamei:
Of course. Thanks for taking time for this :pray:

ybiquitous:
Don't mind. 😄

PR #2197, opened 2021-03-18
Avoid unintentional string replace (Runners::ConfigGenerator)

### LINKED ISSUE #2198: Fail to call Runners::ConfigGenerator.new.generate in some developing environment
## 
I'm using `sider/runners` as a gem library in my project. I found that I cannot call 'Runners::ConfigGenerator.new.generate' with some error.

Here is an example:
```bash
~/tmp/gemtest $ cat Gemfile
gem 'runners', git: 'https://github.com/sider/runners.git', tag: '0.45.0'
~/tmp/gemtest $ bundle install
(..skip..)
Using runners 0.45.0 from https://github.com/sider/runners.git (at 0.45.0@776d818)
Bundle complete! 1 Gemfile dependency, 25 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
~/tmp/gemtest $ bundle exec ruby -e "require 'runners';Runners::ConfigGenerator.new.generate"
Traceback (most recent call last):
        3: from -e:1:in `<main>'
        2: from /home/gnakagaw/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/runners-776d8181a418/lib/runners/config_generator.rb:4:in `generate'
        1: from /home/gnakagaw/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/runners-776d8181a418/lib/runners/config_generator.rb:12:in `load_template'
/home/gnakagaw/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/runners-776d8181a418/lib/runners/config_generator.rb:12:in `read': No such file or directory @ rb_sysopen - /home/gnakagaw/.yml.erbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/runners-776d8181a418/lib/runners/config_generator.rb (Errno::ENOENT)
 ~/tmp/gemtest $
```

```
/home/gnakagaw/.yml.erbenv/..../config_generator.rb (Errno::ENOENT)
```
🤔 `.yml.erbenv`... What is this?

## Cause
I've found the cause. https://github.com/sider/runners/blob/master/lib/runners/config_generator.rb#L11
```ruby
11: filename = __FILE__.sub(".rb", ".yml.erb")
```
I guess that this code premise that `__FILE__` returns always return **source file name**. However, `__FILE__` returns a **full path of the source file** in some cases.

This example shows that `__FILE__` returns a full path to the source file when the file is loaded with `require` keyword.
```bash
$ cat mylib.rb
class MyLib
  def foo()
    puts "foo"
  end

  def getPath()
    puts __FILE__
  end
end
$ cat myprog.rb
require './mylib'

puts __FILE__
MyLib.new.getPath
$ ruby myprog.rb
myprog.rb
/home/gnakagaw/tmp/gemtest/mylib.rb
$
```

I'm using [`rbenv`](https://github.com/rbenv/rbenv) and store gem libraries under `$HOME/.rbenv`. So, the code converts from `.rbenv` to `.yml.erbenv` when I load `runners` gem from other project. This is the cause of this issue.

NOTE: I'm ready to submit a patch for the issue. I'll create a PR soon.

gnakagaw:

Fix #2198 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 - Simplify `sider.yml` schema [#2192](https://github.com/sider/runners/pull/2192)
 - Secure `pip install` [#2194](https://github.com/sider/runners/pull/2194)
 - Introduce common option `target` [#2191](https://github.com/sider/runners/pull/2191)
+- Bug fix in `Runners::ConfigGenerator` [#2198](https://github.com/sider/runners/issues/2198)
 
 ## 0.45.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@ All notable changes to this project will be documented in this file.
 - Simplify `sider.yml` schema [#2192](https://github.com/sider/runners/pull/2192)
 - Secure `pip install` [#2194](https://github.com/sider/runners/pull/2194)
 - Introduce common option `target` [#2191](https://github.com/sider/runners/pull/2191)
-- Bug fix in `Runners::ConfigGenerator` [#2198](https://github.com/sider/runners/issues/2198)
 
 ## 0.45.0
 
--- a/lib/runners/config_generator.rb
+++ b/lib/runners/config_generator.rb
@@ -8,7 +8,7 @@ module Runners
     private
 
     def load_template(tools)
-      filename = __FILE__.gsub(/.rb\Z/, ".yml.erb")
+      filename = __FILE__.gsub(/\.rb\Z/, ".yml.erb")
       erb = ERB.new(File.read(filename), trim_mode: "-")
       erb.filename = filename
       erb.result_with_hash({
--- a/lib/runners/config_generator.rb
+++ b/lib/runners/config_generator.rb
@@ -8,7 +8,7 @@ module Runners
     private
 
     def load_template(tools)
-      filename = __FILE__.gsub(/.rb$/, ".yml.erb")
+      filename = __FILE__.gsub(/.rb\Z/, ".yml.erb")
       erb = ERB.new(File.read(filename), trim_mode: "-")
       erb.filename = filename
       erb.result_with_hash({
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 - Simplify `sider.yml` schema [#2192](https://github.com/sider/runners/pull/2192)
 - Secure `pip install` [#2194](https://github.com/sider/runners/pull/2194)
 - Introduce common option `target` [#2191](https://github.com/sider/runners/pull/2191)
+- Bug fix in `Runners::ConfigGenerator` [#2198](https://github.com/sider/runners/issues/2198)
```
ybiquitous, line 16: [suggestion] This change is internal, so I don't think it should be on the changelog.

```suggestion
```

lib/runners/config_generator.rb
```diff
@@ -8,7 +8,7 @@ def generate(tools: [])
     private
 
     def load_template(tools)
-      filename = __FILE__.sub(".rb", ".yml.erb")
+      filename = __FILE__.gsub(/.rb\Z/, ".yml.erb")
```
ybiquitous, line 11: [nitpick] Escaping `.` seems more accurate, but this is enough 👍 

```suggestion
      filename = __FILE__.gsub(/\.rb\Z/, ".yml.erb")
```

yykamei, line 11: [nitpick] I believe `lib/runners/config_generator.rb` will never be renamed to `lib/runners/config_generator!rb`, but it's just in case.

```suggestion
      filename = __FILE__.gsub(/\.rb\Z/, ".yml.erb")
```

yykamei: Same https://github.com/sider/runners/pull/2197#discussion_r596634338

ybiquitous (approved):
Thanks! 👍

yykamei (approved):
LGTM

PR #2201, opened 2021-03-19
[PHPMD] Handle syntax error as issue

ybiquitous:

This change aims to reduce analysis errors due to PHP syntax errors.
A syntax error will be handled as an issue, instead of raising failures.

I think this behavior is reasonable because just a syntax error in a file will not fail an analysis.

None.


yykamei (approved):
It makes sense to me.

ybiquitous:
@yykamei Thank you for the quick review! 👍

PR #2202, opened 2021-03-19
Add `linter.<id>.dependencies` option for npm tools

ybiquitous:
> Explain a summary, purpose, or background for this change.

This change adds a new feature that users can specify any npm dependencies via `sider.yml`.
For example:

```yaml
linter:
  eslint:
    dependencies:
      - eslint # latest
      - eslint-config-standard@15.0.1
      - { name: "eslint-config-react-app", version: ">=5.0.0" }
    npm_install: true # ignored
```

- When the `dependencies` option is not empty, the `npm_install` option is ignored.
- If users have a problem with `npm install`, this new option should resolve it.

> Refer related issues or pull requests (e.g. `Close #123` or `None`).

Close #1689

> Check the following if needed.

- [x] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.

PR #2211, opened 2021-03-22
Introduce common option `linter.<id>.dependencies`

ybiquitous:

This change aims to apply the new option `linter.<id>.dependencies` (since #2202) to other runners.

- `linter.<id>.gems` will be an alias to `linter.<id>.dependencies`
- `linter.<id>.apt` will be deprecated and `linter.<id>.dependencies` will be recommended instead.
- `linter.<id>.jvm_deps` will be deprecated and `linter.<id>.dependencies` will be recommended instead.
  - `[group, name, version]` style (3-tuple) will be deprecated

See also #2202


lib/runners/cplusplus.rb
```diff
@@ -20,12 +27,21 @@ def install_apt_packages
...
       # select development packages and report others as warning for security concerns
-      packages = Array(config_linter[:apt]).select do |pkg|
-        # @type var pkg: String
-        if pkg.match?(/-dev(=.+)?$/)
-          true
+      packages = Array(config_linter[:dependencies] || config_linter[:apt]).filter_map do |pkg|
+        # @type var pkg: Hash[Symbol, String?] | String
+        case pkg
+        when Hash
+          name = pkg.fetch(:name) or raise pkg.inspect
```
yykamei, line 34: Is `or ...` required? `#fetch` without a default value will raise an error?

ybiquitous: This `or ...` is to prevent the Steep error.

```
+ lib/runners/cplusplus.rb:41:16: [error] Type `(::String | nil)` does not have method `end_with?`
+ │ Diagnostic ID: Ruby::NoMethod
+ │
+ └         if name.end_with? "-dev"
+                   ~~~~~~~~~
```

Since `pkg` is `Hash[Symbol, String?]`, `pkg.fetch(:name)` may return `nil` still.

lib/runners/java.rb
```diff
@@ -31,7 +38,21 @@ def fetch_deps_via_gradle!
...
+      @config_jvm_deps ||= Array(config_linter[:dependencies]).map do |dep|
+        # @type var dep: Hash[Symbol, String] | String
+        case dep
+        when Hash
+          dep.values.join(":")
+        else
+          dep
+        end
+      end
+    end
```
yykamei, line 52: Doesn't `#config_jvm_deps` have to check dependency declaration like this?

https://github.com/sider/runners/blob/bbb7fea7a7e1d7c3a667516f47675c3972912de0/lib/runners/java.rb#L37-L44

lib/runners/cplusplus.rb
yykamei: OK

lib/runners/java.rb
ybiquitous: Yes, I'd like to deprecate this behavior because:

- the Gradle style of `group:name:version` is sufficient
- `linter.<id>.dependencies` newly supports `{ name: "group:name", version: "..." }` style

What do you think?

yykamei: So, you will delete the feature. I understood. I have no opinion about that.

yykamei (approved):
LGTM

lib/runners/java.rb
ybiquitous: Thanks. I've updated the PR description.

ybiquitous:
Thanks for the review!

PR #2219, opened 2021-03-24
Simplify `Runners::Processor.register_config_schema` method

ybiquitous:

This refactoring aims to simplify the `Runners::Processor.register_config_schema` method.
Since the class now has the `.analyzer_id` method, the `name` argument of `.register_config_schema` is needless.

None.

### DIFF
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -3,10 +3,21 @@ module Runners
     extend Forwardable
     include Tmpdir
 
-    class CIConfigBroken < UserError; end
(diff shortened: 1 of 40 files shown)


lib/runners/processor.rb
```diff
@@ -3,10 +3,21 @@ class Processor
     extend Forwardable
     include Tmpdir
 
-    class CIConfigBroken < UserError; end
```
ybiquitous, line 6: [note] `CIConfigBroken` has no longer been used.

PR #2228, opened 2021-03-24
Fix unstable test and Ruby warning

ybiquitous:

This change aims to fix the following error and warning:

- `ConfigGeneratorTest#test_generate_with_tools [/home/runner/work/runners/runners/test/config_generator_test.rb:24]`
- `.../lib/runners/nodejs.rb:106: warning: instance variable @nodejs_force_default_version not initialized`

See <https://github.com/sider/runners/runs/2182668135>

None.

### DIFF
--- a/test/java_test.rb
+++ b/test/java_test.rb
@@ -98,17 +98,19 @@ class JavaTest < Minitest::Test
 
   private
 
-  TestProcessor = Class.new(Runners::Processor) do
-    include Runners::Java
-    def self.analyzer_id; :checkstyle; end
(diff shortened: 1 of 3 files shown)


test/java_test.rb
```diff
@@ -98,17 +98,19 @@ def test_invalid_config
 
   private
 
-  TestProcessor = Class.new(Runners::Processor) do
```
ybiquitous, line 101: [note] We should not call `Class.new` within a class context. It may have a side-effect.

PR #2250, opened 2021-03-30
[ESLint] RE-provide Sider's recommended set of rules

sleeeto:
> Explain a summary, purpose, or background for this change.

Enable Sider's recommended ruleset of 'ESLint'

> Refer related issues or pull requests (e.g. `Close #123` or `None`).

None

> Check the following if needed.

- [x] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.

PR #2266, opened 2021-04-02
[RuboCop] Provide Sider's recommended set of rules

sleeeto:

Provide Sider's recommended ruleset of 'rubocop'

None


images/rubocop/sider_recommended_rubocop.yml
```diff
@@ -0,0 +1,100 @@
+inherit_from:
+  - https://github.com/rubocop/rubocop/blob/HEAD/config/default.yml
```
yykamei, line 2: [question] Why these lines are required? I understand it indicates this configuration overrides the default configuration, but I think this configuration can override it without `inherit_from`. Moreover, this may bring incompatibility with the runtime RuboCop since RuboCop is not always run with the latest version.

sleeeto: OK, I removed it.

sleeeto:
@yykamei Ready.

images/rubocop/sider_recommended_rubocop.yml
```diff
@@ -0,0 +1,105 @@
+# This ruleset has only difference of RuboCop's default ruleset.
+# If you use this ruleset, you inherit the default ruleset.
```
yykamei, line 2: How about adding a date when we decided the ruleset like [this](https://github.com/sider/runners/pull/2250/files#diff-076c1293bc06401a64353fb2f168d83ba76a39fa1e1a5042de0752a55935b050R4)? I'm not sure the exact date and RuboCop version, though.

```suggestion
# This ruleset is based on RuboCop's default configuration.
# We generated with RuboCop x.xx on 2021-xx-xx.
```

sleeeto: OK, I updated it.

images/rubocop/sider_recommended_rubocop.yml
```diff
@@ -0,0 +1,105 @@
+# This ruleset is based on RuboCop's default configuration.
```
ybiquitous, line 1: [question] What means "based on RuboCop's default configuration"?

sleeeto: The default is 
"https://github.com/rubocop/rubocop/blob/master/config/default.yml"
Do I need to add this URL ?

ybiquitous: RuboCop inherits the default configuration by default, so I'm concerned that writing the "based on RuboCop's default configuration" may confusing.

https://docs.rubocop.org/rubocop/configuration.html#inheritance

sleeeto: OK, I removed it.

yykamei (approved):
LGTM

sleeeto:
@ybiquitous Ready.

ybiquitous (approved):

PR #2297, opened 2021-04-14
[Slim-Lint] Support Sass by default

ybiquitous:

Slim is often used with Sass (e.g., Rails), making sense to support Sass by default.
Without this change, users would need to add a Sass gem to `sider.yml` by themselves.

See also:
- https://rubygems.org/gems/sassc

None.


images/slim_lint/Gemfile
```diff
@@ -1,4 +1,5 @@
 source "https://rubygems.org"
 
 gem "slim_lint", "0.20.2"
+gem "sassc", "2.4.0"
```
ybiquitous, line 4: [note] [`sass`](https://rubygems.org/gems/sass] is deprecated, so I've added only `sassc`.

PR #2356, opened 2021-05-07
[PMD CPD, Metrics CodeClone] Suppress STDOUT logging 

gnakagaw:

`pmd_cpd` runnner (`metrics_codeclone` runner) write result data to `STDOUT`. It create huge runner log file when there are many code clones with current implementation. In some case, the output process consumes time more than 20 min. This is PR to suppress such logging.


ybiquitous (approved):
I'm concerned that it will be a bit hard to debug, but it looks like we have no choice. 👍🏼

suetsuguk (approved):
LGTM.

gnakagaw:
> I'm concerned that it will be a bit hard to debug, but it looks like we have no choice.
Surely. But, it takes more than 40 minutes 😭 in canary test...

PR #2349, opened 2021-05-06
[PMD CPD] Add a debug code

gnakagaw:
> Explain a summary, purpose, or background for this change.

We have faced several analysis errors `pmdcpd` saying `OutOfMemory` in Java VM context. (https://github.com/sider/sideci/issues/10979) We have not identify the cause yet. There are two possibilities. One is `pmdcpd`'s bug. The other is inappropriate heap size limitation. This PR adds an extra option expanding maximum heap size of Java VM. 

I'll revert the code after this debug process.

NOTE: The default maximum heap size is 1/4 of the total memory. It is 4GB at Sider production service.

PR #2366, opened 2021-05-10
Specify max heap size for Java tools

gnakagaw:
This is a workaround for https://github.com/sider/sideci/issues/10979. This PR adds `-Xmx8G` option for java VM when it execute `pmd_cpd`. This means that `pmd_cpd` can consume memory as heap up to 8GB. (Now the limit value is 4GB in Sider production service).

The new limitation value is not the best value because it is depends on the target source code. We, however, overcome the existing problem with the parameter. 

NOTE: This is a just workaround. We need to discuss to how to pass the parameter and refactor `pmd_cpd` and `pmc_java` code. I've opened #2367.

A experiment with debug code: https://github.com/sider/runners/pull/2349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@ Misc:
 - **Metrics Complexity** Add workarounds for lizard's bug [#2249](https://github.com/sider/runners/pull/2249)
 - Show the tool's default version on setup [#2313](https://github.com/sider/runners/pull/2313)
 - **PMD CPD** Suppress STDOUT logging [#2356](https://github.com/sider/runners/pull/2356)
-- **PMD CPD** Increase max heap size [#2366](https://github.com/sider/runners/pull/2366)
+- Specify max heap size for Java tools [#2366](https://github.com/sider/runners/pull/2366)
 
 ## 0.47.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Misc:
 - **Metrics Complexity** Add workarounds for lizard's bug [#2249](https://github.com/sider/runners/pull/2249)
 - Show the tool's default version on setup [#2313](https://github.com/sider/runners/pull/2313)
 - **PMD CPD** Suppress STDOUT logging [#2356](https://github.com/sider/runners/pull/2356)
+- **PMD CPD** Increase max heap size [#2366](https://github.com/sider/runners/pull/2366)
 
 ## 0.47.0
 
--- a/images/Dockerfile.java.erb
+++ b/images/Dockerfile.java.erb
@@ -12,3 +12,4 @@ RUN cd "${RUNNERS_DEPS_DIR}" && \
     gradle --no-build-cache --parallel --quiet installDeps && \
     rm build.gradle
 ENV CLASSPATH ${RUNNERS_DEPS_DIR}/*:${CLASSPATH}
+ENV JAVA_TOOL_OPTIONS "-XX:MaxRAMPercentage=50"
--- a/images/checkstyle/Dockerfile
+++ b/images/checkstyle/Dockerfile
@@ -32,6 +32,7 @@ RUN cd "${RUNNERS_DEPS_DIR}" && \
     gradle --no-build-cache --parallel --quiet installDeps && \
     rm build.gradle
 ENV CLASSPATH ${RUNNERS_DEPS_DIR}/*:${CLASSPATH}
+ENV JAVA_TOOL_OPTIONS "-XX:MaxRAMPercentage=50"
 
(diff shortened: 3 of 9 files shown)


ybiquitous:
> This means that `pmd_cpd` can consume memory as heap up to 8GB. (Now the limit value is 4GB in Sider production service).

I'm worried about the description above. If `pmd_cpd` will require memory greater than 4GB, what will happen?

suetsuguk:
My understanding is that the max heap size of pmd_cpd is set to 4GB on the current runner. The "c5.2xlarge" instance of runner has 16GB memory, and Java VM implicitly uses 4GB (one forth of the total memory) as the default heap size. That was not enough for this case, so he increased it to 8GB explicitly.

suetsuguk:
My understanding is that the max heap size of pmd_cpd is set to 4GB on the current runner. The instance of runner has 16GB memory, and Java VM implicitly uses 4GB (one forth of the total memory) as the default heap size. That was not enough for this case, so he increased it to 8GB explicitly.

suetsuguk (approved):
Thank you for handling the problem. Let's continue working together on the related issue.

ybiquitous:
How about using [`-XX:MaxRAMPercentage`](https://www.eclipse.org/openj9/docs/xxinitialrampercentage/) if possible?

I think it good to specify the heap size generally, not only for a workaround. But if considering various environments running Sider (e.g. Enterprise), it is hard to determine the heap size.
Perhaps, `-XX:MaxRAMPercentage` may be a better option than `-Xmx`.

What do you think?

gnakagaw:
Here is a simple test program: MemoryEater.java
```java
import java.util.ArrayList;
import java.util.Random;

public class MemoryEater {
    public static void main(String[] arg) {
        ArrayList<Object> list = new ArrayList<>();
        long c = 0;
        while(true){
            byte[] b = new byte[4096];
            new Random().nextBytes(b);
            list.add(b);

            System.out.printf("\r%f MB", (float)c/1024/1024);
            c = c + b.length;
        }
    }
}
```

Executed on a PC with 16GB RAM.
```ShellSession
bash-4.4# java -XX:MaxRAMPercentage=10 MemoryEater
1570.148438 MBException in thread "main" java.lang.OutOfMemoryError: Java heap space
        at MemoryEater.main(MemoryEater.java:9)
bash-4.4# java -XX:MaxRAMPercentage=50 MemoryEater
7849.832031 MBException in thread "main" java.lang.OutOfMemoryError: Java heap space
        at MemoryEater.main(MemoryEater.java:9)
bash-4.4#
```

We can use `MaxRAMPercentage`  for this issue.

gnakagaw:
I've modified the code to use `MaxRAMPercentage`.
fix: 39033979430bf298ecabdff3c0f42c3ae25e78e8

Should we apply the option to system wide? (i.e., apply to `devon_rex_java`). We can set with [`JAVA_TOOL_OPTIONS` environment variable](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/envvars002.html).

ybiquitous:
`JAVA_TOO_OPTIONS` looks good. I think it good to set it in this repository like `CLASSPATH`. See below:

https://github.com/sider/runners/blob/a0466aa67122b46bd693191bf7031856735f6ab1/images/Dockerfile.java.erb#L14

gnakagaw:
I added `MaxRAMPercentage` parameter in `Dockerfile.java.erb` 900d949e4917ea1a5d2bfec6b73de46ab511f890

Here is an experiment. It seems to be OK.
```
analyzer_runner@b2a4a6476eb7:~$ java MemoryEater
Picked up JAVA_TOOL_OPTIONS: -XX:MaxRAMPercentage=50
7850.875000 MBException in thread "main" java.lang.OutOfMemoryError: Java heap space
        at MemoryEater.main(MemoryEater.java:9)
analyzer_runner@b2a4a6476eb7:~$
```

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Misc:
 - **Metrics Complexity** Add workarounds for lizard's bug [#2249](https://github.com/sider/runners/pull/2249)
 - Show the tool's default version on setup [#2313](https://github.com/sider/runners/pull/2313)
 - **PMD CPD** Suppress STDOUT logging [#2356](https://github.com/sider/runners/pull/2356)
+- **PMD CPD** Increase max heap size [#2366](https://github.com/sider/runners/pull/2366)
```
ybiquitous, line 44: Can you update the changelog and the PR title?

```suggestion
- Specify max heap size for Java tools [#2366](https://github.com/sider/runners/pull/2366)
```

ybiquitous (approved):
Thanks 👍🏼

PR #2427, opened 2021-06-07
Change REXML to Nokogiri to read XML files

dev-yakuza:
> Explain a summary, purpose, or background for this change.

Change `REXML` to `Nokogiri` to read XML files for the performance

> Refer related issues or pull requests (e.g. `Close #123` or `None`).

None

> Check the following if needed.

- [x] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.

PR #2435, opened 2021-06-09
[PHPMD] Replace XML reporter with JSON one

ybiquitous:

Parsing XML is hard because:
- slow
- unstable
- dependent on a library (REXML vs Nokogiri)

The PHPMD's XML reporter implementation seems flaky because it uses string concatenation.
On the other hand, its JSON reporter uses the built-in function [`json_encode()`](https://www.php.net/manual/en/function.json-encode.php), which seems more reilable.

See also:
- https://github.com/phpmd/phpmd/blob/2.10.1/src/main/php/PHPMD/Renderer/JSONRenderer.php
- https://github.com/phpmd/phpmd/blob/2.10.1/src/main/php/PHPMD/Renderer/XMLRenderer.php

Related to #2427 


dev-yakuza (approved):
LGTM 👍 
Thanks for your works!

PR #2447, opened 2021-06-14
PMD CPD performance tuning

dev-yakuza:

This PR has two refactorings for performance tuning of `PMC CPD`.

- move out the logic not related loop from the loop
- not use the `search(xpath)`

None

### DIFF
--- a/lib/runners/processor/pmd_cpd.rb
+++ b/lib/runners/processor/pmd_cpd.rb
@@ -167,7 +167,25 @@ module Runners
       stdout.sub!(/<\?xml version="1\.0" encoding=".+"\?>/, '<?xml version="1.0" encoding="UTF-8"?>')
 
       read_xml(stdout).root.elements.each do |elem_dupli|
-        files = elem_dupli.search('file').map{ |f| to_fileinfo(f) }
+        lines_text = elem_dupli[:lines] or raise "required lines: #{elem_dupli.inspect}"
+        lines = Integer(lines_text)
+
+        tokens_text = elem_dupli[:tokens] or raise "required tokens: #{elem_dupli.inspect}"
+        tokens = Integer(tokens_text)
+
+        files = []
+        codefragment = nil
+        elem_dupli.elements.each do |elem|
+          case elem.name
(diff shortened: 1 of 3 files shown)


suetsuguk (approved):
The loop over `elements` seems to be faster than using `search` method. How much did it improve?
Thank you.

dev-yakuza:
@suetsuguk 
Thanks for your review.

The result is like the below.

|      | before | after |
| -- | ------- | ----- |
| issue count | 26,536 | 41,407 |

The number of issues created using the same time period(about 30min).

lib/runners/processor/pmd_cpd.rb
```diff
@@ -167,15 +167,38 @@ def construct_result(stdout)
...
+        codefragment = nil
+        elem_dupli.elements.each do |elem|
+          case elem.name
+          when "file"
+            files << to_fileinfo(elem)
+          when "codefragment"
+            codefragment = elem.content
+          end
+        end
+        codefragment or raise "required codefragment: #{elem_dupli.inspect}"
```
gnakagaw, line 186: [suggestion] Why don't you wrap the process in a method like `create_duplicated_file_list`?  It makes clear the code intent.

lib/runners/processor/pmd_cpd.rb
```diff
@@ -167,15 +167,38 @@ def construct_result(stdout)
...
+
+        fileobjs = create_file_objects(files)
```
gnakagaw, line 188: [suggenstion] How about change `fileobjs` to a more explicit name like `dupicated_files_group`? `fileobjs` is confusing because we also define `files` variable.

dev-yakuza: @gnakagaw 
Fixed it on https://github.com/sider/runners/pull/2447/commits/c2e3babb64e1e974eff13bd0364b8b22aa4d1e5a
Thanks!

gnakagaw (approved):

ybiquitous (approved):
LGTM 👍🏼

PR #2465, opened 2021-06-21
Add login to Public ECR

ybiquitous:
> Explain a summary, purpose, or background for this change.

See <https://github.com/docker/login-action#aws-public-elastic-container-registry-ecr>

> Refer related issues or pull requests (e.g. `Close #123` or `None`).

Related to #2464

> Check the following if needed.

- [x] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md) if this change is notable.

PR #2468, opened 2021-06-22
Publish Docker images to Amazon ECR Public

ybiquitous:

See https://docs.aws.amazon.com/AmazonECR/latest/public/docker-push-ecr-image.html

Follow-up of #2465

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,9 @@ jobs:
       - name: Publish Docker image
         if: |
           steps.define_vars.outputs.tag == 'master' || startsWith(github.ref, 'refs/tags/')
-        run: bundle exec rake docker:push
+        run: |
+          bundle exec rake docker:push
+          bundle exec rake docker:push DOCKER_REGISTRY=public.ecr.aws
         env:
           ANALYZER: ${{ matrix.analyzer }}
           TAG: ${{ steps.define_vars.outputs.tag }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
     branches: ["**"]
 
+env:
+  # TODO: Will change to `public.ecr.aws/sider`.
+  AWS_ECR_REGISTRY: public.ecr.aws/r8f9z5e1
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -67,13 +67,12 @@ jobs:
           echo "GITHUB_REF: ${GITHUB_REF}"
           tag="${GITHUB_REF#refs/*/}"
           tag="${tag///}"
+          tag=master
```
ybiquitous, line 70: **TODO: REMOVE THIS LINE.**

.github/workflows/build.yml
```diff
@@ -7,6 +7,10 @@ on:
   pull_request:
     branches: ["**"]
 
+env:
+  # TODO: Will change to `public.ecr.aws/sider`.
+  AWS_ECR_REGISTRY: public.ecr.aws/r8f9z5e1
```
gnakagaw, line 12: Can we change to `/sider` now? or later?

ybiquitous: Maybe later. I have requested the custom alias `sider`, but it does not seem to be approved by AWS yet.

image

gnakagaw: Got it 👍

gnakagaw (approved):
🚀

.github/workflows/build.yml
ybiquitous: If we will use the custom alias, do we need to re-push images?

ybiquitous: Ref: https://docs.aws.amazon.com/AmazonECR/latest/public/public-registry-settings.html

gnakagaw: > if approved, both the default alias and custom alias can be used to access your public repositories. 

Seems to be OK. But I have a concerns about image metadata. Here is a example of `docker inspect [image]`. It contains `r8f9z5e1` in image URL.. Sorry I'm not confident about this. Let's try it💪 
```
[
    {
        "Id": "sha256:dc5406089013f7817b0a2e58907bc132e7780e70c4aa304bec0b7ede3b8df340",
        "RepoTags": [
            "public.ecr.aws/r8f9z5e1/runner_brakeman:master"
        ],
        "RepoDigests": [
            "public.ecr.aws/r8f9z5e1/runner_brakeman@sha256:bdfd3704fe97c33e65d294aabe9578bf08908f6997e6a7401f30e116171b7839"
        ],
        "Parent": "",
        "Comment": "buildkit.dockerfile.v0",
```

ybiquitous:
Publishing works! 🎉 
https://github.com/sider/runners/actions/runs/959639543

.github/workflows/build.yml
ybiquitous: I’ve created an issue #2469 to fix the TODO.

PR #2495, opened 2021-07-01
Use git-clone(1) and git-sparse-checkout(1) for faster download

### LINKED ISSUE #2494: Use git-sparse-checkout(1) to reduce storage consumption
### Problem

Here is the current way to check out a repository's source files:

1. Check out all the files in a repository via git-checkout(1):
  https://github.com/sider/runners/blob/ebac3c2fad50fcab306d99007be2cdb90e8528ed/lib/runners/workspace/git.rb#L38
2. Delete ignored files via `sider.yml`:
  https://github.com/sider/runners/blob/ebac3c2fad50fcab306d99007be2cdb90e8528ed/lib/runners/harness.rb#L37-L42

If a repository has many image files (e.g., `.jpg`), even if the `ignored` option of `sider.yml` is specified, a Docker container storage is temporarily consumed between the step 1 and 2 above.

This may lead to needless requirements for Docker container storage size.

### Solution

If we can use git-sparse-checkout(1) properly, we may be able to check out only necessary files, excluding `ignore` files via `sider.yml`.

### See also

- https://git-scm.com/docs/git-sparse-checkout
- https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---sparse
- https://git-scm.com/docs/partial-clone

ybiquitous:

- The `Workspace::Git` class uses:
  - git-clone(1) instead of git-init(1)
  - git-sparse-checkout(1) before git-checkout(1)
- The `Ignoring` class is no longer needed.

Fix #2494


ybiquitous:
Here is a smoke test case for the `ignore` option:

https://github.com/sider/runners/blob/8a730fb5a0d54b4cb65b2a88e7a5cde8543dd241/test/smokes/shellcheck/with_ignore/sider.yml#L1-L4

lib/runners/workspace/git.rb
```diff
@@ -65,29 +72,43 @@ def remote_url
...
+      args = [
         '--quiet',
         '--no-tags',
         '--no-recurse-submodules',
         'origin',
         '+refs/heads/*:refs/remotes/origin/*',
         *git_source.refspec,
-      ].freeze
+      ]
+      shell.capture3_with_retry!("git", "fetch", *args, tries: try_count, sleep: sleep_lambda)
```
gnakagaw, line 94: [question] Is it acceptable to use `git fetch`? The command seems to download all git metadata from remote origin, i.e., it also download [blob objects](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) of ignored file. 

The size of a blob object is smaller than its original file because Git compresses the blob. But, generally, compression has little effect on image files because they are compressed originally.

A solution is partial cloning. e.g. `git clone --filter=blob:none --no-checkout [remote url] .`
https://git-scm.com/docs/partial-clone
https://github.blog/jp/2021-01-13-get-up-to-speed-with-partial-clone-and-shallow-clone/

I surveyed `git fetch` command specification. It doesn't have any option like `--filter=blob:none`, unfortunately.

lib/runners/workspace/git.rb
```diff
@@ -65,29 +72,43 @@ def remote_url
...
 
-    def git_fetch_args
-      @git_fetch_args ||= [
+    def git_setup
+      shell.capture3!("git", "init", "--initial-branch=main")
+      shell.capture3!("git", "config", "gc.auto", "0")
+      shell.capture3!("git", "config", "advice.detachedHead", "false")
+      shell.capture3!("git", "config", "core.quotePath", "false")
+      shell.capture3!("git", "config", "core.hooksPath", mktmpdir.to_path) # NOTE: Prevent evil hooks from being executed.
+      shell.capture3!("git", "remote", "add", "origin", remote_url)
```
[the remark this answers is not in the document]
ybiquitous: We don't need to call `git sparse-checkout init --cone` because:

- `git sparse-checkout set` enables the *sparse* mode
- *gitignore* patterns don't work with `--cone` (see the [doc](https://git-scm.com/docs/git-sparse-checkout#_cone_pattern_set))

We can see a test failure if the suggested change is applied:

```console
$ git diff
index b3cc3e54..a511f670 100644
@@ -79,6 +79,7 @@ module Runners
       shell.capture3!("git", "config", "core.quotePath", "false")
       shell.capture3!("git", "config", "core.hooksPath", mktmpdir.to_path) # NOTE: Prevent evil hooks from being executed.
       shell.capture3!("git", "remote", "add", "origin", remote_url)
+      shell.capture3!("git", "sparse-checkout", "init", "--cone")
     end
 
     # @see https://git-scm.com/docs/git-fetch

$ bundle exec ruby -I test test/workspace/git_test.rb -n test_prepare_head_source_with_ignoring_files
Run options: -n test_prepare_head_source_with_ignoring_files --seed 38329

# Running:

F

Finished in 1.451723s, 0.6888 runs/s, 0.6888 assertions/s.

  1) Failure:
WorkspaceGitTest#test_prepare_head_source_with_ignoring_files [test/workspace/git_test.rb:74]:
--- expected
+++ actual
@@ -1 +1 @@
-#<Set: {"sider.yml", "sideci.yml", ".git"}>
+#<Set: {"sider.yml", "sideci.yml", "てすと.txt", "README.md", ".git"}>


1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

lib/runners/workspace/git.rb
ybiquitous: Your idea is great! I have reimplemented `Workspace::Git` via git-clone(1). See 2dca2f46

lib/runners/workspace/git.rb
```diff
@@ -21,24 +23,31 @@ def range_git_blame_info(path_string, start_line, end_line, trace: false)
...
     def prepare_head_source
-      shell.capture3!("git", "init", "--initial-branch=main")
-      shell.capture3!("git", "config", "gc.auto", "0")
-      shell.capture3!("git", "config", "advice.detachedHead", "false")
-      shell.capture3!("git", "config", "core.quotePath", "false")
-      shell.capture3!("git", "config", "core.hooksPath", mktmpdir.to_path) # NOTE: Prevent evil hooks from being executed.
-      shell.capture3!("git", "remote", "add", "origin", remote_url)
+      git_clone
+      git_setup
+      git_fetch(git_source.refspec) unless git_source.refspec.empty?
```
ybiquitous, line 28: [note] Thanks to git-clone(1), we don't need to fetch `+refs/heads/*:refs/remotes/origin/*`.

lib/runners/workspace/git.rb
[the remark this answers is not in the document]
gnakagaw: Oops. My experiment procedure was wrong. Sorry.

You are right. `git` command set `core.sparsecheckout=true` implicitly.

gnakagaw (approved):
:shipit:

ybiquitous:
Thanks for the review. I will fix the CI failure (eslint) with another PR.

PR #2522, opened 2021-07-06
[Metrics File Info] Improve `git log --numstat` speed

ybiquitous:

`git clone --filter=blob:none` makes `git log --numstat` too slow because each blob object is downloaded on demand.

Related to #2519


suetsuguk (approved):
LGTM. Thank you!

PR #2526, opened 2021-07-08
[Metrics File Info] Fix parsing git-ls-files(1) output

ybiquitous:

When the `.gitattributes` file is included, the current parsing for the `git-ls-files(1)` command’s output fails.
Then, incorrect file paths are passed to the `wc(1)` command, and it raises the following error:

```
wc: 'eol=lf      '$'\t''.gitattributes': No such file or directory
```

This change aims to fix the parsing.

See also:
- https://git-scm.com/docs/git-ls-files#_output

None.


suetsuguk (approved):
LGTM. Thank you.

PR #2553, opened 2021-07-21
[WIP][Metrics] Filter out vendor/** from being analyzed by default

suetsuguk:
### Summary or Background

This aims to reduce analysis duration, meaningless issues and unexpected analysis errors caused by files coming from various third party sources. The common `vendor` directory gets ignored by default.

### Issues or Pull Requests

None.

### Checklist

### DIFF
--- a/lib/runners.rb
+++ b/lib/runners.rb
@@ -71,6 +71,7 @@ require "runners/schema/options"
 require "runners/schema/trace"
 require "runners/schema/config"
 require "runners/schema/result"
+require "runners/schema/metrics"
 
 # processors
 require "runners/processor"
--- a/lib/runners/config.rb
+++ b/lib/runners/config.rb
@@ -58,6 +58,10 @@ module Runners
       end
     end
 
+    def self.invert_patterns(patterns)
+      patterns.map { |pat| pat.start_with?('!') ? pat.delete_prefix('!') : "!#{pat}" }
+    end
+
     attr_reader :path, :raw_content
 
     def initialize(path:, raw_content:)
@@ -99,6 +103,10 @@ module Runners
 
     def ignore_patterns
       @ignore_patterns ||= Array(content[:ignore])
+    end
+
+    def metrics_ignore_patterns
+      @metrics_ignore_patterns || Array(content.dig(:linter, :metrics, :ignore))
     end
 
     def linter(id)
--- a/lib/runners/schema/metrics.rb
+++ b/lib/runners/schema/metrics.rb
@@ -1,0 +1,14 @@
+module Runners
+  module Schema
+    Metrics = _ = StrongJSON.new do
+      extend ConfigTypes
+      # @type self: MetricsClass
+
+      let :config, object?(
+        ignore: one_or_more_strings?,
+      )
+    end
+
+    Config.register(name: :metrics, schema: Metrics.config)
+  end
+end
--- a/lib/runners/harness.rb
+++ b/lib/runners/harness.rb
@@ -35,8 +35,9 @@ module Runners
         # HACK: MetricsFileInfo needs all Git blob objects to calculate code churn.
         #       This code is so ugly, but I could not find a good way.
         fast = processor_class != Processor::MetricsFileInfo
+        run_metrics = [Processor::MetricsCodeClone, Processor::MetricsComplexity, Processor::MetricsFileInfo].include?(processor_class)
 
-        workspace.open(fast: fast) do |git_ssh_path, changes|
+        workspace.open(fast: fast, run_metrics: run_metrics) do |git_ssh_path, changes|
           @config = conf = workspace.config
 
           begin
--- a/lib/runners/workspace.rb
+++ b/lib/runners/workspace.rb
@@ -15,12 +15,12 @@ module Runners
       @shell = Shell.new(current_dir: working_dir, trace_writer: trace_writer, env_hash: {})
     end
 
-    def open(fast: true)
+    def open(fast: true, run_metrics: false)
       prepare_ssh do |git_ssh_path|
         trace_writer.header "Set up source code"
 
         trace_writer.message "Preparing head commit tree..."
-        prepare_head_source(fast: fast)
+        prepare_head_source(fast: fast, run_metrics: run_metrics)
 
         changes =
           if options.source.base
@@ -39,7 +39,7 @@ module Runners
       []
     end
 
-    def prepare_head_source(fast: true)
+    def prepare_head_source(fast: true, run_metrics: false)
       raise NotImplementedError
     end
 
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -22,7 +22,7 @@ module Runners
       raise BlameFailed, "git-blame failed: #{exn.stderr_str}"
     end
 
-    def prepare_head_source(fast: true)
+    def prepare_head_source(fast: true, run_metrics: false)
       git_clone(fast: fast)
       git_setup
 
@@ -34,7 +34,8 @@ module Runners
       git_checkout
 
       # Next, fetch remaining files except for *ignored* files.
-      git_sparse_checkout_set "/**", *config.ignore_patterns.map { |pat| "!#{pat}" }
+      patterns = (run_metrics ? config.metrics_ignore_patterns : []) + config.ignore_patterns
+      git_sparse_checkout_set "/**", *Config.invert_patterns(patterns)
       git_checkout
     end
 
--- a/test/config_test.rb
+++ b/test/config_test.rb
@@ -14,6 +14,10 @@ class ConfigTest < Minitest::Test
     end
   end
 
+  def test_invert_patterns
+    assert_equal ["foo.txt", "!bar/baz/**", "qux/**"], Runners::Config.invert_patterns(["!foo.txt", "bar/baz/**", "!qux/**"])
+  end
+
   def test_content_without_sider_yml
     assert_equal({}, Runners::Config.new(path: nil, raw_content: nil).content)
   end
@@ -68,6 +72,7 @@ class ConfigTest < Minitest::Test
           jshint: nil,
           ktlint: nil,
           languagetool: nil,
+          metrics: nil,
           misspell: nil,
           phinder: nil,
           phpmd: nil,


suetsuguk:
As per suggestion from @dev-yakuza , we will ignore following directories and files by default.
- vender/**
- node_moduels/**
- \*.min.\*
- \*.map

ybiquitous:
@suetsuguk Can you add a simple usage for the new option of `sider.yml` to this PR description?

lib/runners/schema/metrics.rb
```diff
@@ -0,0 +1,14 @@
...
+      # @type self: MetricsClass
+
+      let :config, object?(
+        ignore: one_or_more_strings?,
+      )
+    end
+
+    Config.register(name: :metrics, schema: Metrics.config)
+  end
+end
```
ybiquitous, line 14: [suggestion] How about adding a `Schema::ConfigTypes#metrics` method as below? 👇🏼 
I feel it is more consistent with other processor classes.

```diff
diff --git a/lib/runners/processor/metrics_codeclone.rb b/lib/runners/processor/metrics_codeclone.rb
index b2a0b3e7..5cb328aa 100644
--- a/lib/runners/processor/metrics_codeclone.rb
+++ b/lib/runners/processor/metrics_codeclone.rb
@@ -6,12 +6,16 @@ module Runners
       extend Schema::ConfigTypes
 
       # @type self: SchemaClass
+      let :config, metrics
+
       let :issue, object(
         clones: integer,
         total_clone_lines: integer,
       )
     end
 
+    register_config_schema SCHEMA.config
+
     def_delegators :@pmd_cpd,
       :add_warning,
       :warnings,
diff --git a/lib/runners/processor/metrics_complexity.rb b/lib/runners/processor/metrics_complexity.rb
index 51a6400e..04b9827e 100644
--- a/lib/runners/processor/metrics_complexity.rb
+++ b/lib/runners/processor/metrics_complexity.rb
@@ -6,11 +6,15 @@ module Runners
       extend Schema::ConfigTypes
 
       # @type self: SchemaClass
+      let :config, metrics
+
       let :issue, object(
         CCN: integer,
       )
     end
 
+    register_config_schema SCHEMA.config
+
     # Set maximum recursion depth for python runtime
     # Python runtime has a stack depth limitation. The default value is 1000.  Lizard violates the limitation when it analyze complex source code files.
     # You can ease the limitation with this parameter.
diff --git a/lib/runners/processor/metrics_fileinfo.rb b/lib/runners/processor/metrics_fileinfo.rb
index 9d75df5f..a0da5471 100644
--- a/lib/runners/processor/metrics_fileinfo.rb
+++ b/lib/runners/processor/metrics_fileinfo.rb
@@ -4,6 +4,8 @@ module Runners
       extend Schema::ConfigTypes
 
       # @type self: SchemaClass
+      let :config, metrics
+
       let :issue, object(
         lines_of_code: integer?,
         last_committed_at: string,
@@ -14,6 +16,8 @@ module Runners
       )
     end
 
+    register_config_schema SCHEMA.config
+
     # Basically, the code churn is the number of times or the sum of added/deleted lines a file has changed within a specified period of time.
     # Like other products, we adopt the last 90 days before the latest commit date. However, if a project is not so active and has only
     # a few commits in a month, the churn values fluctuate on the scatter plot (churn v.s quality metrics) due to ambiguity of the small number of commits.
diff --git a/lib/runners/schema/config.rb b/lib/runners/schema/config.rb
index 0ead3de8..fb9efef9 100644
--- a/lib/runners/schema/config.rb
+++ b/lib/runners/schema/config.rb
@@ -73,6 +73,10 @@ module Runners
           **fields,
         )
       end
+
+      def metrics
+        object?(ignore: one_or_more_strings?)
+      end
     end
 
     Config = _ = StrongJSON.new do
```

lib/runners/schema/metrics.rb
```diff
@@ -0,0 +1,14 @@
+module Runners
+  module Schema
+    Metrics = _ = StrongJSON.new do
```
radomirbosak, line 3: What is the metrics schema supposed to represent?

lib/runners.rb
```diff
@@ -71,6 +71,7 @@
 require "runners/schema/trace"
 require "runners/schema/config"
 require "runners/schema/result"
+require "runners/schema/metrics"
```
radomirbosak, line 74: What is the difference between runners and processors?

radomirbosak (approved):
I don't understand the codebase much yet, and I haven't checked the test changes, but the changes in `lib/` look good to me. I assume that that's where the business logic happens.
```
$ git diff master...HEAD  --stat -- lib/
 lib/runners.rb                |  1 +
 lib/runners/config.rb         |  9 +++++++++
 lib/runners/harness.rb        |  2 +-
 lib/runners/processor.rb      |  4 ++++
 lib/runners/schema/metrics.rb | 14 ++++++++++++++
 lib/runners/workspace.rb      |  6 +++---
 lib/runners/workspace/git.rb  |  5 +++--
 7 files changed, 35 insertions(+), 6 deletions(-)
```

lib/runners/schema/metrics.rb
suetsuguk: This schema defines the option `linter.metrics.ignore` described in this PR [description](https://github.com/sider/runners/pull/2553#issue-694038166).

lib/runners.rb
suetsuguk: My understanding is that the processors is a glue class that executes a static analysis tool and translates its result into unified object. The runners is a set of processors.

lib/runners/schema/metrics.rb
suetsuguk: As I added a description about the `linter.metrics.ignore` option, this single option is intended to be shared with three metrics runners.
Very first time I tried to define an `ignore` option in each metric runner as you suggested above, but I got three separate options of `linter.metrics_fileinfo.ignore`, `linter.metrics_complexity.ignore` and `linter.metrics_codeclone.ignore`. It might require users to perform troublesome work of writing the same items in different options.
So I created a shared schema class and registered it with the identifier `metrics` as the representative of the three.

ybiquitous: I see. Then, how about the following change to improve the `Processor.register_config_schema` method?

```diff
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -15,8 +15,8 @@ module Runners
         .freeze
     end
 
-    def self.register_config_schema(schema)
-      Schema::Config.register(name: analyzer_id, schema: schema)
+    def self.register_config_schema(schema, name: analyzer_id)
+      Schema::Config.register(name: name, schema: schema)
     end
 
     def self.config_example
```

```diff
--- a/lib/runners/processor/metrics_codeclone.rb
+++ b/lib/runners/processor/metrics_codeclone.rb
@@ -6,12 +6,16 @@ module Runners
       extend Schema::ConfigTypes
 
       # @type self: SchemaClass
+      let :config, metrics
+
       let :issue, object(
         clones: integer,
         total_clone_lines: integer,
       )
     end
 
+    register_config_schema SCHEMA.config, name: :metrics
+
     def_delegators :@pmd_cpd,
       :add_warning,
       :warnings,
```

suetsuguk: That has a very simple structure. I took `metrics_fileinfo` as a metrics representative.
commit 6e23ed54770e4d7d442aa9f1ea95c874d9245321

ybiquitous: Nice! Will you change other processors than `metrics_fileinfo.rb`?

suetsuguk: [Here](https://github.com/sider/runners/blob/6d290e65e4253f97b324b38a72e5e3a22e7bea3b/lib/runners/schema/config.rb#L89-L93) is a check for multiple registration, so we can register only one schema named `metrics`.

ybiquitous: Ah, I see 😅  👍🏼

lib/runners/config.rb
```diff
@@ -35,6 +35,7 @@ def initialize(message, path_name:, raw_content:, attribute:)
 
     FILE_NAME = "sider.yml".freeze
     FILE_NAME_OLD = "sideci.yml".freeze
+    DEFAULT_METRICS_IGNORE_PATTERNS = ["**/vendor/**", "**/node_modules/**", "*.min.*", "*.map"].freeze
```
ybiquitous, line 38: [trivial] How about multi-line to reduce diff when new patterns will be added?

```suggestion
    DEFAULT_METRICS_IGNORE_PATTERNS = [
      "**/vendor/**",
      "**/node_modules/**",
      "*.min.*",
      "*.map",
    ].freeze
```

ybiquitous:
[suggestion] I feel maintainability would be worse because of adding arguments to the `Workspace#open` method.
For example, how about adding a `Workspace#processor_class` attribute like this? ⬇️ 

```diff
diff --git a/lib/runners/harness.rb b/lib/runners/harness.rb
index 08e44336..c87fb46d 100644
--- a/lib/runners/harness.rb
+++ b/lib/runners/harness.rb
@@ -30,13 +30,9 @@ module Runners
 
     def run
       ensure_result do
-        workspace = Workspace.prepare(options: options, working_dir: working_dir, trace_writer: trace_writer)
+        workspace = Workspace.prepare(options: options, working_dir: working_dir, trace_writer: trace_writer, processor_class: processor_class)
 
-        # HACK: MetricsFileInfo needs all Git blob objects to calculate code churn.
-        #       This code is so ugly, but I could not find a good way.
-        fast = processor_class != Processor::MetricsFileInfo
-
-        workspace.open(fast: fast, run_metrics: processor_class.metrics?) do |git_ssh_path, changes|
+        workspace.open do |git_ssh_path, changes|
           @config = conf = workspace.config
 
           begin
diff --git a/lib/runners/processor.rb b/lib/runners/processor.rb
index aeb77cdc..e58987fe 100644
--- a/lib/runners/processor.rb
+++ b/lib/runners/processor.rb
@@ -24,7 +24,7 @@ module Runners
     end
 
     def self.metrics?
-      !!self.name&.start_with?("Runners::Processor::Metrics")
+      [MetricsComplexity, MetricsCodeClone, MetricsFileInfo].include?(self)
     end
 
     attr_reader :guid, :working_dir, :config, :shell, :trace_writer, :warnings, :options
diff --git a/lib/runners/workspace.rb b/lib/runners/workspace.rb
index 99b8f860..877e8657 100644
--- a/lib/runners/workspace.rb
+++ b/lib/runners/workspace.rb
@@ -2,25 +2,26 @@ module Runners
   class Workspace
     include Tmpdir
 
-    def self.prepare(options:, working_dir:, trace_writer:)
-      Workspace::Git.new(options: options, working_dir: working_dir, trace_writer: trace_writer)
+    def self.prepare(options:, working_dir:, trace_writer:, processor_class:)
+      Workspace::Git.new(options: options, working_dir: working_dir, trace_writer: trace_writer, processor_class: processor_class)
     end
 
-    attr_reader :options, :working_dir, :trace_writer, :shell
+    attr_reader :options, :working_dir, :trace_writer, :processor_class, :shell
 
-    def initialize(options:, working_dir:, trace_writer:)
+    def initialize(options:, working_dir:, trace_writer:, processor_class:)
       @options = options
       @working_dir = working_dir
       @trace_writer = trace_writer
+      @processor_class = processor_class
       @shell = Shell.new(current_dir: working_dir, trace_writer: trace_writer, env_hash: {})
     end
 
-    def open(fast: true, run_metrics: false)
+    def open
       prepare_ssh do |git_ssh_path|
         trace_writer.header "Set up source code"
 
         trace_writer.message "Preparing head commit tree..."
-        prepare_head_source(fast: fast, run_metrics: run_metrics)
+        prepare_head_source
 
         changes =
           if options.source.base
@@ -39,7 +40,7 @@ module Runners
       []
     end
 
-    def prepare_head_source(fast: true, run_metrics: false)
+    def prepare_head_source
       raise NotImplementedError
     end
 
diff --git a/lib/runners/workspace/git.rb b/lib/runners/workspace/git.rb
index 0cb088eb..24369923 100644
--- a/lib/runners/workspace/git.rb
+++ b/lib/runners/workspace/git.rb
@@ -22,8 +22,8 @@ module Runners
       raise BlameFailed, "git-blame failed: #{exn.stderr_str}"
     end
 
-    def prepare_head_source(fast: true, run_metrics: false)
-      git_clone(fast: fast)
+    def prepare_head_source
+      git_clone
       git_setup
 
       # Fetch a pull request if specified (including a forked repository).
@@ -34,7 +34,8 @@ module Runners
       git_checkout
 
       # Next, fetch remaining files except for *ignored* files.
-      patterns = (run_metrics ? config.metrics_ignore_patterns : []) + config.ignore_patterns
+      patterns = config.ignore_patterns
+      patterns += config.metrics_ignore_patterns if processor_class.metrics?
       git_sparse_checkout_set "/**", *Config.invert_patterns(patterns)
       git_checkout
     end
@@ -87,14 +88,19 @@ module Runners
 
     # @see https://git-scm.com/docs/git-clone
     # @see https://git-scm.com/docs/partial-clone
-    def git_clone(fast:)
+    def git_clone
       options = %w[
         --no-checkout
         --no-recurse-submodules
         --no-tags
         --quiet
       ]
-      options << "--filter=blob:none" if fast
+
+      # NOTE: MetricsFileInfo needs all Git blob objects to calculate code churn.
+      if processor_class != Processor::MetricsFileInfo
+        options << "--filter=blob:none"
+      end
+
       shell.capture3_with_retry!("git", "clone", *options, "--", remote_url, ".", tries: try_count, sleep: sleep_lambda)
     rescue Shell::ExecError => exn
       raise CloneFailed, "git-clone failed: #{exn.stderr_str}"
```

lib/runners/processor/metrics_fileinfo.rb
```diff
@@ -16,6 +16,8 @@ class Processor::MetricsFileInfo < Processor
       )
     end
 
+    # The Schema::Config.register() method checks for multiple registration, so we can register only one schema named `metrics`.
+    # There is no need to define the `metrics` schema nor register it in Processor::MetricsComplexity and Processor::MetricsCodeClone.
```
ybiquitous, line 20: Nice! 👍🏼

suetsuguk:
Thank you for your suggestion.
commit 06276f69fa21704d40025b43c28001f8a5bb8479

suetsuguk:
As for the changes in 090854f0b8a16502e15da10f5ec8769b56269fc1, I never thought that this circular reference was allowed. A method of the `Processor` class has a reference of the child classes of `Processor`.

suetsuguk:
As for 1f9a8905ddf22ce34df06d4fba5eb207fe4cef4a, the patterns are overridden by the patterns that appear behind. I think it is better to place patterns in this order;
- default patterns (node_modules, etc.) for metrics
- user defined patterns for metrics via `linter.metrics.ignore`
- user defined patterns via `ignore`

What do you think, everyone?

ybiquitous:
@suetsuguk I agree with the order you suggests in https://github.com/sider/runners/pull/2553#issuecomment-905361063 👍🏼

PR #2664, opened 2021-09-07
Fix wrong version by updating manually

dev-yakuza:
### Summary or Background

`Runners` was released, but the version was not changed since `0.53.0`. 

`Runners::VERSION` and `git describe --abbrev=0 --tags` were different, so the version is not changed well.

This PR fixes the version manually to fix the version gap.

### Issues or Pull Requests

None

### Checklist


ybiquitous (approved):
Thanks. LGTM :+1:

PR #2650, opened 2021-08-31
Add actionlint analyzer

radomirbosak:
### Summary or Background

See #2601 

### Issues or Pull Requests

Fixes #2601 

### Checklist

- [x] Add a new entry to [CHANGELOG.md](https://github.com/sider/runners/blob/master/CHANGELOG.md)
- [x] Add a new tool to analyzers.yml.
- [x] Run bundle exec rake readme:generate and commit the updated README.md.
- [x] Add a new tool to the Dependabot configuration if needed.
- [x] Provide the tool's options that users can customize.
- [x] Write smoke test cases. E.g. all options, successful end, warnings, errors, etc.
- [x] Add a CI setting.
- [x] Write a document in sider-docs and link the pull request: https://github.com/sider/sider-docs/pull/686

PR #2675, opened 2021-09-13
Add goodcheck rule for one_or_more_strings? method

radomirbosak:
Add a goodcheck rule which suggests replacing
```
enum?(string, array(string))
```
with
```
one_or_more_strings?
```

Followup to https://github.com/sider/runners/pull/2650#discussion_r705168320


suetsuguk (approved):
LGTM. Thank you.

Issue #2767, opened 2021-10-29
Fix CI errors (ESLint)


gnakagaw:
`build (eslint)` CI job have been failed  for repository-wide. Fix it.

e.g.
- https://github.com/sider/runners/runs/3991440783
- https://github.com/sider/runners/runs/3991440168
- https://github.com/sider/runners/runs/3991330525
...

gnakagaw:
related? https://github.com/npm/cli/issues/3666

gnakagaw:
I'll update npm from v7 to v8. 
https://github.com/sider/devon_rex/pull/694

radomirbosak:
Interesting. I always thought the bug is on our side, but I could not figure out where. But it really might be an `npm` bug.

radomirbosak:
I tried updating npm, but it did not help. ESLint test still fails https://github.com/sider/runners/pull/2781

gnakagaw:
This issue will be resolved with next devon_rex update.

radomirbosak:
This was resolved by https://github.com/sider/devon_rex/pull/701 and published in devon_rex 2.46.0 https://github.com/sider/devon_rex/releases/tag/2.46.0

PR #2781, opened 2021-11-05
Bump npm to version 8.1.1

radomirbosak:
This is an attempt to resolve the ESLint CI errors #2767

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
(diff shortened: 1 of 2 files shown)


radomirbosak:
It does not seem to have helped..
image

gnakagaw:
@radomirbosak

FYI:
The approach of this PR was not good because the npm issue occurs NOT in GitHub Actions context but in `docker build` context. We have to modify contents in docker images to fix it.  https://github.com/sider/runners/pull/2795 is the best approach.

radomirbosak:
Thank you for the explanation. I was confused about where the version should be updated and didn't realize it was over at the devon_rex repository.

PR #2782, opened 2021-11-05
Do not enforce strict npm peer dependencies

radomirbosak:
This commit removes the `--strict-peer-dependencies` switch in order to fix failing CI.

The npm install `--strict-peer-dependencies` switch was added in the commit fa6a5e95d8b1858079548187997204c127ad419c, however, possibly due to an [npm bug](https://github.com/npm/cli/issues/3666) building the ESLint image now fails.

Note that when the underlying issue has been resolved, we should probably restore this switch.

Relevant docs:
https://docs.npmjs.com/cli/v7/commands/npm-install#strict-peer-deps

> **strict-peer-deps**
> * Default: false
> * Type: Boolean
>
> If set to true, and --legacy-peer-deps is not set, then any conflicting peerDependencies will be treated as an install failure, even if npm could reasonably guess the appropriate resolution based on non-peer dependency relationships.
> 
> By default, conflicting peerDependencies deep in the dependency graph will be resolved using the nearest non-peer dependency specification, even if doing so will result in some packages receiving a peer dependency outside the range set in their package's peerDependencies object.
> 
> When such and override is performed, a warning is printed, explaining the conflict and the packages involved. If --strict-peer-deps is set, then this warning is treated as a failure.

Fixes: #2767

### DIFF
--- a/images/Dockerfile.npm.erb
+++ b/images/Dockerfile.npm.erb
@@ -5,7 +5,7 @@ FROM sider/devon_rex_npm:2.45.11
 # Install the default version
 COPY --chown=<%= chown %> images/<%= analyzer %>/package.json ${RUNNERS_DEPS_DIR}/
 RUN cd "${RUNNERS_DEPS_DIR}" && \
-    npm install --strict-peer-deps --ignore-scripts --engine-strict --no-progress --no-save && \
+    npm install --ignore-scripts --engine-strict --no-progress --no-save && \
     rm package*.json && \
     # We can't use NODE_PATH
     # See the details here: https://github.com/sider/runners/issues/2636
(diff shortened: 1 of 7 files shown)


radomirbosak:
Since other solutions for #2767 did not work for me, I tried removing `--strict-peer-dependencies` here.

What do you think @gnakagaw @ybiquitous @dev-yakuza ? Are there risks involved when we remove this switch?
I'm not very familiar with npm ecosystem, so I'm not sure what can happen, but I guess some of the plugins might not work?

On the other hand, if the #2767 issue is caused by a bug, this won't cause much damage (if it's temporary).

ybiquitous:
The #2781 solution is not insufficient. We need to bump the npm version in devon_rex:
https://github.com/sider/devon_rex/pull/700

ybiquitous:
I can locally confirm that npm@8 successfully installs packages:


https://user-images.githubusercontent.com/473530/140463832-460cd5ca-f668-4044-804b-24530100f1c5.mov

radomirbosak:
Thank you for investigating! I merged the devon_rex npm version bump PR and seems to have resolved the issue - ESLint check now succeeds in a test pull request:
https://github.com/sider/runners/pull/2795

I'll need to create a new devon_rex tag so that runners can use it.

In any case, this PR is not needed anymore so I'll close it.

Issue #2812, opened 2021-11-17
Sider Scan runs at fixed intervals


sleeeto:

sleeeto:
Thanks for your support !

sleeeto:
I want to update the version of Sider Scan to 2.1.0.

PR #2813, opened 2021-11-17
Introduce Sider Scan runner

sleeeto:
### Summary or Background

add a sider scan for GitHub action.

### Issues or Pull Requests

https://github.com/sider/runners/issues/2812

### Checklist

No change logs.

### DIFF
--- a/.github/workflows/sider_scan.yml
+++ b/.github/workflows/sider_scan.yml
@@ -3,7 +3,8 @@ name: Sidr Scan Runner
 on:
   schedule:
     - cron: '0 5 * * 1-5'
-
+  workflow_dispatch:
+  
 jobs:
   restore-job:
     runs-on: ubuntu-20.04
--- a/.github/workflows/sider_scan.yml
+++ b/.github/workflows/sider_scan.yml
@@ -46,7 +46,7 @@ jobs:
           name: radump
           path: /tmp
       - run: |
-          echo { \
+          echo "{ \
             \"user\": { \"contactEmail\": \"dev@siderlabs.com\" }, \
             \"channels\": { \
               \"email\": { \
@@ -73,7 +73,7 @@ jobs:
               \"onlyIfDetect\": true, \
               \"enableDetailReport\": true \
             } \
-          } > .siderscan.json
+          }" > .siderscan.json
       - name: Start analysis
         working-directory: /scan
         run: |
--- a/.github/workflows/sider_scan.yml
+++ b/.github/workflows/sider_scan.yml
@@ -1,0 +1,62 @@
+name: Sidr Scan Runner
+
+on:
+  schedule:
+    - cron: '0 5 * * 1-5'
+
+
+jobs:
+  restore-job:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Download the lastest artifact
+        id: latest_artifact
+        continue-on-error: true
+        run: |
+          wget --header "Accept: application/vnd.github.v3+json" --header "Authorization: token ${{github.token}}" https://api.github.com/repos/${{github.repository}}/actions/artifacts -O - \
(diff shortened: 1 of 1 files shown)


.github/workflows/sider_scan.yml
```diff
@@ -0,0 +1,62 @@
...
+        continue-on-error: true
+        uses: actions/download-artifact@v2
+        with:
+          name: radump
+          path: /tmp
+      - run: ls -la
+      - run: ls /tmp
+      - name: Start analysis
+        working-directory: /scan
+        run: |
```
radomirbosak, line 52: > __[actionlint]__ shellcheck reported issue in this script: SC2215:warning:5:1: This flag is used as a command name. Bad line break or missing [ .. ]? ([view](https://sider.review/gh/repos/202486306/pulls/2813#issue-6278071))
>
> <details>
>
> | Rule |
> | ---- |
> | `shellcheck/SC2215` |
>
> References:
> - <https://github.com/koalaman/shellcheck/wiki/SC2215>
>
> You can [close this issue](https://sider.review/gh/repos/202486306/pulls/2813#issue-6278071) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

Missing backslash after
```
-f /tmp/output.radump
```

radomirbosak (changes requested):
actionlint reports errors

.github/workflows/sider_scan.yml
sleeeto: Thanks, I fixed it.

gnakagaw:
I've fixed all `actionlint` warnings.

.github/workflows/sider_scan.yml
```diff
@@ -0,0 +1,89 @@
+name: Sidr Scan Runner
```
radomirbosak, line 1:
```suggestion
name: Sider Scan Runner
```

.github/workflows/sider_scan.yml
```diff
@@ -0,0 +1,89 @@
...
+                \"defaults\": { \
+                  \"from\": \"scan-dev@siderlabs.com\" \
+                } \
+              } \
+            }, \
+            \"report\": { \
+              \"onlyIfDetect\": true, \
+              \"enableDetailReport\": true \
+            } \
+          }" > .siderscan.json
```
radomirbosak, line 76: I wonder if we can use something like bash heredoc here to avoid the backslashes (at the end of line and before double quotes).

e.g.
```bash
cat << EOF > file.txt
The current working directory is: $PWD
You are logged in as: $(whoami)
EOF
```

I haven't yet tried if this works in github actions workflow file too.

gnakagaw: I found an example using heredoc in `run` command.
https://qiita.com/hkusu/items/3dbf09f7fa83ba6d457c#%E3%82%B9%E3%82%AF%E3%83%AA%E3%83%97%E3%83%88

BTW, IMO, we should use [`envsubst`](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html) to inject secrets into the config file.

.github/workflows/sider_scan.yml
gnakagaw: Thanks. I'll fix that.

.github/workflows/sider_scan.yml
radomirbosak: Just in case, I looked at the log produced by the action, but all secrets have been fortunately replaces by `***`.

gnakagaw: >  all secrets have been fortunately replaces by ***
I've tested it of course 😄

gnakagaw:
I'm attacking the issue in #2815.

PR #2815, opened 2021-11-18
Fix SiderScan config

gnakagaw:
We've introduced `SiderScan` in #2813. We found several issues after the merging. This PullReq fix them.


lib/runners/processor/eslint.rb
```diff
@@ -32,7 +32,7 @@ class Processor::Eslint < Processor
     end
 
     CONSTRAINTS = {
-      "eslint" => Gem::Requirement.new(">= 5.0.0", "< 8.0.0").freeze,
+      "eslint" => Gem::Requirement.new(">= 5.0.0", "< 9.0.0").freeze,
```
radomirbosak, line 35: Is this supposed to be in the same PR?

gnakagaw: No. It's a mistake...

gnakagaw: @radomirbosak Thanks so much. I'll revert it.

Issue #2741, opened 2021-10-12
[ESLint] Update dependency (ESLint 8)


gnakagaw:
[ESLint 8 is now available](https://eslint.org/blog/2021/10/eslint-v8.0.0-released). Update dependency of `eslint` runner.
https://github.com/sider/runners/blob/5f6492937012dcb0cd58b88768371dd24668ec1c/lib/runners/processor/eslint.rb#L35

radomirbosak:
Sounds good.

> Another option is removing @vue/eslint-config-typescript from the dependency. But, I don't know whether it is possible or not.

I'm not sure, but it looks like it's a very simple package, it sets only three or four rule defaults for eslint. We may be able to do it by ourselves.
https://github.com/vuejs/eslint-config-typescript/blob/e11179c0dade55484ba631ef5e519ea640526c8e/recommended.js#L10-L22

radomirbosak:
Support for eslint 8.0.0 was added to `@vue/eslint-config-typescript`
https://github.com/vuejs/eslint-config-typescript/releases/tag/v8.0.0

radomirbosak:
Issue for **eslint-plugin-react**: https://github.com/yannickcr/eslint-plugin-react/issues/3055

radomirbosak:
And lastly, **eslint-plugin-react-hooks** published an alpha package with ESLint 8.0.0 support, however didn't release a proper versioned package with *latest* tag. See comment here: https://github.com/facebook/react/pull/22248#issuecomment-940731956

Setting
```json
    "eslint-plugin-react-hooks": "alpha",
```
solves the conflict, but is probably not the best long-term solution.

Relevant npmjs: https://www.npmjs.com/package/eslint-plugin-react-hooks

ybiquitous:
Is it possible to just change the upper limit as below?

https://github.com/sider/runners/blob/749bc4b4c842d4fb736bf282fe383e5c6c550083/lib/runners/processor/eslint.rb#L35

```diff
- "eslint" => Gem::Requirement.new(">= 5.0.0", "< 8.0.0").freeze,
+ "eslint" => Gem::Requirement.new(">= 5.0.0", "< 9.0.0").freeze,
```

gnakagaw:
@radomirbosak IMO, It's OK to update ESLint dependency not for `eslint` runner but for `sider/runners` repository. What do you think?

radomirbosak:
> Is it possible to just change the upper limit as below?

[...]

PR #2820, opened 2021-11-18
[ESLint] Allow ESLint8.x in user dependency

gnakagaw:
### Background
Now, users cannot run ESLint 8.x on `eslint` runner even if they specify in `package-lock.json` or `yarn.lock`. This is because [the default ESLint version on the runner is 7.32.0](https://github.com/sider/runners/blob/4192659ffeb1463de2297c78cef3e423bc694f7d/images/eslint/package.json#L7), and [the runner has a constraint to limit the upper version](https://github.com/sider/runners/blob/4192659ffeb1463de2297c78cef3e423bc694f7d/lib/runners/processor/eslint.rb#L35).

This patch eases the constraint to allow using ESLint`8.x`. Users will be able to use `8.x` when they configure it explicitly. Note that the patch don't affect on the default ESLint version.

This is an `eslint` runner trace after applying this patch. It describes that a user can use ESLint `8.2.0`. 
image

### Issues or Pull Requests
#2741

### Checklist


radomirbosak (approved):
LGTM :+1:

PR #2855, opened 2021-12-13
Run sider scan 2.1.0

sleeeto:
### Summary or Background

I updated the version of Sider Scan to 2.1.0.

### Issues or Pull Requests

https://github.com/sider/runners/issues/2812

### DIFF
--- a/.github/workflows/sider_scan.yml
+++ b/.github/workflows/sider_scan.yml
@@ -3,33 +3,12 @@ name: Sider Scan Runner
 on:
   schedule:
     - cron: '0 5 * * 1-5'
-  workflow_dispatch:
-  
+
 jobs:
-  restore-job:
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Download the lastest artifact
-        id: latest_artifact
-        continue-on-error: true
-        run: |
-          wget --header "Accept: application/vnd.github.v3+json" --header "Authorization: token ${{github.token}}" https://api.github.com/repos/${{github.repository}}/actions/artifacts -O - \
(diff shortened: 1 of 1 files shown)


.github/workflows/sider_scan.yml
```diff
@@ -70,18 +51,18 @@ jobs:
...
-        working-directory: /scan
         run: |
           sider run \
-          -c /builds/.siderscan.json \
-          -p /builds/ \
+          -c ./.siderscan.json \
+          -p ./ \
           -f /tmp/output.radump \
-          -t "\"Sider Scan Developer\" <scan-dev@siderlabs.com>"
+          -t "\"Sider Scan Developer\" <scan-dev@siderlabs.com>" \
```
radomirbosak, line 64: Is the result currently reported only to this email address?

.github/workflows/sider_scan.yml
```diff
@@ -70,18 +51,18 @@ jobs:
...
             \"report\": { \
-              \"onlyIfDetect\": true, \
               \"enableDetailReport\": true \
             } \
           }" > .siderscan.json
+      - run: mkdir -p scan_result
       - name: Start analysis
-        working-directory: /scan
         run: |
           sider run \
```
radomirbosak, line 60: It would be great to have this as an GitHub action.

.github/workflows/sider_scan.yml
radomirbosak: Also, is there a plan to implement sider scan as a PR check? Or is Scan not suitable for giving a passed/failed answer that the check requires?

PR #2939, opened 2022-02-15
Skip tests with inaccessible gem mirrors

radomirbosak:
The gems mirror https://rubygems.cae.me.uk stopped responding, and our tests started failing.
I wasn't able to find an alternative mirror, so we'll skip the alternative-source tests for now.


gnakagaw (approved):
It's OK as a workaround. But, we need to rewrite the code.

gnakagaw:
@radomirbosak 
I've googled about rubygems mirror. It seems that there is no official public rubygems mirror. Perhaps we should run a rubygems proxy such as https://github.com/rubygems/rubygems-mirror.

PR #2975, opened 2022-03-10
Clear cache before running npm install in ESLint

radomirbosak:
ESLint fails for some repos on npm install, but succeeds on retry.
However running npm twice effectively doubles the time spent on npm
install which is sometimes the slowest part of an analysis.

Clearing cache before npm install prevents it from failing.


radomirbosak:
CC @gnakagaw

radomirbosak:
This did not help. `npm install` still fails on first try with
```
npm ERR! npm ERR! Cannot read property 'pickAlgorithm' of null
```
I must come up with another solution. For now, I will revert this PR.

PR #2977, opened 2022-03-11
Update npm version to latest before running npm install

radomirbosak:
This change is an attempt to solve the `npm install` failing on first time
(with an error message like `Cannot read property 'pickAlgorithm' of null`).

The change will be reverted if it does not work.


radomirbosak:
I am trying to put this in front of the `npm install` command to update the npm version:
image

But 5 smoke tests are failing: 
* `typescript`
* `pre_installed_packages`
* `unmet_peer_deps` 
* `private_dependencies`
* `private_dependencies_with_option_dependencies`

radomirbosak:
Runner code cannot update the global npm installation, so I changed the code so that `npm` is updated much earlier - in the docker image building phase.
image

PR #2995, opened 2022-03-23
Retry docker:push command in case it fails

radomirbosak:
docker:push to aws ecr in build action sometimes fails with
`TLS handshake timeout` error. Retrying is the easiest workaround.


gnakagaw (approved):

radomirbosak:
I've checked the logs for the build action and this fix seems to work. Some pushes (e.g. build(misspell)) succeeded on 3rd attempt.

